PR #200, opened 2015-02-14
Edits to API docs

davidtheclark:
Here is installment two of edits -- this time up to `Root` node ~ line 840. I have more points for you think about, too:

Documentation style concerns:

- This time I tried to clarify the headings (which will also clarify their # links) by prefixing properties and methods with a reference to the thing they're attached to. Please look that over and see if you like it, or if there is another style that you would prefer. I think that *something* has to be there ---- for instance, to differentiate one `map` from another. Otherwise, links won't be reliable and you will have to scroll up to see which section you are in to understand which method you are looking at.
- Notice that on 667 (and some following places) I added `[]` brackets around an optional argument. I think that it is important to distinguish optional arguments. I also think that it would be important to document the expected types of each argument (e.g. string, array of strings, boolean) for every method. That would be fastest I'm sure if you did it, @ai, since you know the API best.

Clarification:

- The list module examples seem weird to me. Are those real use cases?
- The last `node.source` example also seems weird given the text above it. Shouldn't it illustrate cloning a node, or setting its source manually?
- I don't understand the `defaultType` stuff with `node.style()`, ~538.
- I don't understand what 550-551 mean (about containers).
- See line 768: I don't understand the `fast` option --- what was there before did not make sense to me.

If you could clear up all those points, that would be great. At some point I will try to finish the rest (it's a *long* document!)


ai:
Thanks, you did a really big and important work.

1. I think title prefixes for direct links to methods is a nice idea. It is a bit noise, but when we make a PostCSS site, we will remove prefixes (but leave a unique links).
2. What Markdown formats you suggest for method types?
3. I made `list` examples more similar to real CSS examples.
4. I make better examples for `node.source`.
5. Code style properties is too difficult right now :(. Let’s stay as we have in 4.0 and refector it in 5.0.
6. 550-551 lines means that `rule.nodes` can contains `Rule` or `AtRule` chlidren. It is unexpected behaviour, because CSS syntax spec do not allow it in most of cases.

ai:
About `fast` option. Code wwillsay better, that my English:

This code is slow, because regexp is slow operation:

```js
css.eachDecl(function (decl) {
    decl.value = decl.value.replace(/\d+rem/, function { … });
});
```

We can make it better with some pre-check. There is no reason to call slow regexp if `value` even don’t contains a `rem` string:

```js
css.eachDecl(function (decl) {
    if ( decl.value.indexOf('rem') != -1 ) {
        decl.value = decl.value.replace(/\d+rem/, function { … });
    }
});
```

Option `fast` is this pre-check.

davidtheclark:
Thanks @ai. I understand `fast` now and will incorporate that and your other comments into the next round of edits.

For documenting method parameters ---

For writing out the function, I like adding brackets around optional parameters. This is an established pattern used in large popular projects like [Express](http://expressjs.com/4x/api.html) and [Gulp](https://github.com/gulpjs/gulp/blob/master/docs/API.md) and [lodash](https://lodash.com/docs).

As for documenting the parameter types: I think it would add a lot of lines but would be as clear as possible if we documented each parameter and return value as in the lodash docs. We could use the patterns popularized by JSDoc & Closure Compilaer for types. Markdown format could be that we put *at the end of each method section* something like this:

```
**Arguments**
- `nameFilter` {String|RegExp} - description
```

ai:
Yeap. This format is looks good. I will add it later.

davidtheclark:
Ok. I will put what I can in there for the section towards the end when I edit them this morning.

PR #201, opened 2015-02-15
Edits to README

davidtheclark:
I have offered a lot of suggestions for improving the README. I also have a few points to ask about:

- At line 12 I reference "many, many" users of Autoprefixer. Do you have specific stats that would impress?
- In the "How PostCSS differs from Prepreprocessors" section, I think I included the spirit of your original points and added several others. But did I miss something important? Is there more that should be added?
- I could not understand the descriptions for `css2modernizr` and `postcss-single-charset`. Those should be improved by somebody who knows what they do (their own repos are not helpful either).
- I recommend noting whenever a plugin implements syntax from a CSS spec, because I think that is a major selling point for PostCSS. I tried to do that, but I may have missed something.
- I wonder if the little source map tutorial at lines 317-342 could be left out. Maybe it could be replaced with a link to an article about source maps. I don't think we have to go to the trouble of accurately explaining them here.
- I did not really understand the paragraph starting at line 386. That could use some clarification.

I hope you like the changes. I am happy to modify this PR based on feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,11 @@ PostCSS itself is very small and focused. It includes a CSS parser, a CSS node
 tree API, a source map generator, and a node tree stringifier — and that's it.
 All of the stylesheet transformations it powers are encapsulated in modular plugins.
 And these plugins are themselves small and focused plain JS functions, which typically
-receive a CSS node tree at one end, apply transformations to it, and then return
+receive a CSS node tree at one end, apply transformations to it, and return
 a modified tree that can be used by other plugins or written to a file.
 
 Individual plugins implement specific transformations; and by using various plugins
-together you can put together your ideal CSS workflow.
+together you can create an ideal CSS workflow for yourself and your team.
 For instance, if you like the power provided by preprocessors
 like Sass, you could combine [Autoprefixer], [cssnext], [CSS Grace], [postcss-nested],
 [postcss-mixins], and [postcss-easings] to write CSS like this:
@@ -63,7 +63,7 @@ like Sass, you could combine [Autoprefixer], [cssnext], [CSS Grace], [postcss-ne
 (Note that this is not *the* way to use PostCSS: this is just one possible feature set
 enabled by the plugins listed above.)
 
-Twitter account for articles, releases and new plugins: [@postcss].
+Twitter account for articles, releases, and new plugins: [@postcss].
 Weibo account: [postcss].
 
 <a href="https://evilmartians.com/?utm_source=postcss">
@@ -83,7 +83,7 @@ and variables.
 
 In contrast, PostCSS provides a simple API that modular plugins can use to understand,
 transform, and create CSS. PostCSS plugins, therefore, can be created, maintained,
-and used independently.
+and implemented independently.
 And they can perform many different tasks, not just compile special syntaxes to CSS.
 Each plugin does one thing well; and all PostCSS plugins do not have to fit together
 into a large "superset" language.
@@ -140,12 +140,12 @@ plugins that you use.
 ### Perfomance
 
 PostCSS is one of the fastest CSS parsers written in JS. (Only [CSSOM] is
-faster, and only because it is less accurate.) So PostCSS will parse your CSS and
-apply plugin transformations faster than any other CSS processor out there.
+faster, and only because it is less accurate.) So PostCSS will read your CSS and
+apply transformations faster than any other stylesheet processor out there.
 
 If you use Ruby Sass now, PostCSS could significantly improve your development process:
 *PostCSS processing is 40 times faster than Ruby Sass compilation.*
-And even if you use the entire [cssnext] plugin pack, PostCSS (written in JS) is still
+And even if you throw in the entire [cssnext] plugin pack, PostCSS (written in JS) is still
 *4 times faster than [libsass]* (written on C++).
 
 [libsass]: https://github.com/sass/libsass
@@ -155,7 +155,7 @@ And even if you use the entire [cssnext] plugin pack, PostCSS (written in JS) is
 
 PostCSS plugins can read and rebuild an entire CSS node tree.
 With this power, plugin authors are able to create tools that would be impossible
-to build into preprocessors (like [Autoprefixer]).
+to build into preprocessors (for example, [Autoprefixer]).
 
 PostCSS-powered tools can do much more than transform special syntax into browser-friendly
 CSS. PostCSS plugin authors have built linters (like [doiuse] and [postcss-bem-linter]),
@@ -163,8 +163,8 @@ code review tools (like [list-selectors]), and minifiers (like [csswring]).
 With [postcss-data-packer], you can create a cacheable "sprite"
 by moving all `data:uri` values to separate file.
 
-One unique example of PostCSS's power is [RTLCSS]. As you know, some languages, such
-as Arabic and Hebrew, write from right-to-left (RTL), instead of the more widespread left-to-right
+One unique example of PostCSS's power is [RTLCSS]. As you know, in some languages, such
+as Arabic and Hebrew, writing moves from right-to-left (RTL), instead of the more widespread left-to-right
 convention. Because a language's directionality affects its readers' perspective,
 an international site's layout needs to change for RTL users, not just its text.
 (Check out [Arabic Wikipedia] as an example.)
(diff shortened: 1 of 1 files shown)


ai:
I copy your Twitter question here to clean Twitter notifications and don’t miss it:

«Do you know if it would be accurate to say PostCSS parses and transforms your CSS faster than any other CSS processor out there?»

/cc @ai

jeddy3:
@davidtheclark This is *really* well written. I don't normally gush, but this is fantastic! :)

> PostCSS allows plugin authors to create whatever features they can imagine, while empowering PostCSS users to add only those features that they want to their workflow.

I especially like how you've chosen this as the first advantage. If the goal of rewriting the readme is to grow the user base (as @ai said), then emphasising that postcss can cater to a user's exisiting workflow, whether it's based on sass/less/stylus, rework/myth or a bit of both, is key.

I just read your [article](http://davidtheclark.com/excited-about-postcss/) about it too, where you articulate the benefits of "selectively including plugins for features that fit needs" so well.

> I recommend noting whenever a plugin implements syntax from a CSS spec, because I think that is a major selling point for PostCSS. I tried to do that, but I may have missed something.

Agreed. Personally, this is what attracted me to PostCSS, but on a larger scale:

* It is [popping up](http://alistapart.com/column/what-will-save-us-from-the-dark-side-of-pre-processors) on the likes of alistapart, which has a *huge* reach.
* The interest in [myth](https://github.com/segmentio/myth) (3500 stars).
* The trend towards simplifying CSS, which, I think, can be seen within the recent workflow articles that you listed [here](https://github.com/davidtheclark/scalable-css-reading-list#workflow-overviews).

One very minor thing, the cssnext documentation stresses that it *transpiles*, rather than *polyfills*, CSS4+ to CSS3. Is it worth using that same terminology in the readme? If only because it might allow postcss to piggy-back on some of the [6to5/babel](https://github.com/babel/babel) JavaScript transpiling momentum.

MoOx:
We should be clear about the fact that **no postcss plugin produce polyfills**.

It's a pre processing step (if you consider "processing" the moment where the browser interpret the CSS).

I used "transform" & "transpile" words to avoid "pre-processing" since people usually think about Sass & friends.

But we are just doing the same thing with a parser limited to CSS syntax.

ai:
@davidtheclark we can use npm downloads count as lower bound. But I think big companies names is enough and we can miss “and many, many other organizations and individuals around the world”.

ai:
@davidtheclark can we focused this phrase “The most popular PostCSS plugin, [Autoprefixer]” on PostCSS, not Autoprefixer. Like “PostCSS is used by…”.

ai:
@davidtheclark about 21-27 lines. First section is not for features. We have a special section for it above. It is very important to keep first section short.

ai:
82-84 lines should be moved to section at line 103

ai:
I think we should not mention benefits from line 98, because users can ask: does plugins has a conflicts. And it will be very difficult question for us :).

ai:
@davidtheclark Can I ask you to use more simple English :). Many people with very low lingvistic skills will read it. For example, “tailor” and “harness” words is difficult to understand.

ai:
Section “How PostCSS differs from Preprocessors” should be not about benefits, but about core difference. After reading user should understand what is preprocessor and what is PostCSS.

ai:
> I recommend noting whenever a plugin implements syntax from a CSS spec, because I think that is a major selling point for PostCSS. I tried to do that, but I may have missed something.

I mentioned before, that non-spec plugins is very important too, because we need to test some ideas in wild before try to create spec.

But I agree that it is a selling point. Let’s add them as 4th feature to Features section. That we have a plugins, that not a hack, but implement a CSS specs. And it is a user choice, what type of plugins to use. But spec plugins is better beause … (you already explain it better that me in your awesome article).

ai:
Word “accommodate” is very difficult to non-native speaker too :).

ai:
> [postcss-assets] isolates stylesheets from environmental changes, gets image sizes and inlines files.

Description doesn’t show how this isolation works. Maybe “put width/height from image file” is better?

ai:
`css2modernizr` read CSS, and return config. By this config you can made custom Modernizr build only with tests, that will be really used in your CSS.

`postcss-single-charset` will add `@charset` on top of the file and remove any other `@charset` from file (for example, because you join multiple CSS files by postcss-import ot gulp-contcat). Safari has a problems if it find several `@charset` at-rules in CSS.

ai:
Hm. I didn’t remember why I add text on 386 line :(. Feel free to remove it in your PR.

ai:
@davidtheclark good work, let’s update PR and we are ready to announce that we have good docs now :).

davidtheclark:
@jeddy3 Thanks!

@MoOx I definitely agree with staying away from "polyfill" and using "transpile" instead. I'll integrate that vocabulary.

@ai I will work on addressing all of your feedback and see if the next version fits what you want.

davidtheclark:
Re this comment: https://github.com/postcss/postcss/pull/201#issuecomment-74423253

I agree that the first section should be very short. The reason I lengthened that paragraph is because I thought that the long example needed more context or else it could be misleading to people skimming. Maybe the example should be moved to a different section so that the first section is even shorter? I just think that if you include an example like that right there are at the top, without more explanation about picking-and-choosing your plugins, people might look at the example and quickly assume "If I use PostCSS then I would use syntax like this." You know what I mean?

davidtheclark:
@ai does this comment https://github.com/postcss/postcss/pull/201#issuecomment-74423430 refer to the paragraph about PostCSS plugins supporting different syntaxes, or the one about plugins progressing at different rates?

davidtheclark:
What do you think about alphabetizing the plugin list? It's kind of a long list to be in a random order.

davidtheclark:
@ai What do you think about the source map tutorial at the beginning of the Source Maps section? I was wondering if it might be removed in favor of a link to a longer authoritative article somewhere.

davidtheclark:
@ai I am trying to rework the PostCSS vs. Preprocessors section so that is not about the benefits of PostCSS but just about the functional differences. (Moving benefits to Features.) But I am having some trouble understanding your points #1 and #2 enough to write them out.

> With preprocessors you write your CSS on special programming language. It is like a PHP, but you mix control statement with styles. As result your styles is slow, because programming language is too compilcated. With PostCSS you write styles on normal CSS, just with custom at-rules and functions.

Isn't this statement dependent on what plugin authors decide to do? For example, the ACSS plugins have custom comments — and am I right that somebody *could* write PostCSS plugins that allow
conditionals and loop and other control statements within your stylesheets?

> Preprocessors tools (like Compass) is written mainly in same preprocessors language. As result this tools is very limited. The libraries adds only a custom functions, variables or mixins. There is no way to add new syntax for CSS 4 polyfills. In PostCSS all magic is written on JS and uses big universe of npm packages. So you have better and smarter tools.

I thought that one of the reasons people liked Compass was because they could write Compass plugins that take full advantage of Ruby -- they didn't have to be restricted to Sass's language. (Am I wrong about that?) But I do get how other popular tools like Bourbon, Susy, and LessHat are just collections of functions, variables, and mixins. Should we maybe not mention Compass here?

ai:
@davidtheclark maybe you are right, that current example can scare some users and we should tell, that you can select any other plugins. Let’s just make this explanation shorter.

ai:
@davidtheclark I think about alphabetizing plugins, but we have too many plugins right now and big list will scary users. So I order plugins by GitHub stars.

But maybe we can find different solution. Maybe we can add some subsections with plugin categories?

ai:
@davidtheclark I still try to find good source map article. Maybe we should stay description. Anyway it is simple and we have one place to explain everything?

ai:
@davidtheclark good question about loops in ACSS :). Yes, people may write a loops on PostCSS. But it will not be really Turing machine programming language, like Sass. Maybe somebody will write full programmming language support, but it is very difficult and it will not be spirit of PostCSS.

Maybe we should think, what is PostCSS before rewrite this section. I think, PostCSS is when you write something like CSS and takes JS libraries, that transforms CSS nodes tree. In Sass you write CSS tamplates on Sass programming language and takes libraries on Sass (with some Ruby/C++/JS extennsions) which provide only mixins and functions, but can not change syntax.

Maybe we should just write some short description like that above? Because it is very important to explain what is a difference (it is like a branding) and good explanations is very short explanations. I think that current explanations in master’s `README.md` is too long.

davidtheclark:
Categories certainly would help with the plugin list, if we can come up with an adequate taxonomy. I didn't know about the Github star ordering :) I will just leave the order as is for now but open a ticket to discuss categorization.

ai:
Yeap, lets keep GitHub stars order and add categories on next week after we decide what categories we need #208

MoOx:
I've a bunch of modifications to offer too that I would submit when this PR gets merged.

MoOx:
Maybe instead of making huge PR, we should try to work on severals small PR, don't you think ?

ai:
@MoOx I agree, but right now we should finish it quickly, so we can do it in one PR.

ai:
Ouh, hope I didn’t broke anyway with this difficult rebase

ai:
Hm. Maybe current example shows only one side of PostCSS? We can add second example with only cssnext features. @MoOx @davidtheclark what do you think? Anyway many code examples is better, that more text :).

Of course, we can just remove “Sass like” example, but I think it is very important for ex-Sass users and two examples will shows the modularity power of PostCSS (we can even remove variables examples from Modulratiy section in Features) and will be good for both types of PostCSS users.

davidtheclark:
@ai I like the idea of adding at least one more example.

ai:
Nice. We can remove all cssnext plugins from “Sass like” examples, to make it shorter.

@MoOx can you write some small, but awesome example?

MoOx:
Can you be happy with what I already put in the cssnext playground ? http://cssnext.github.io/cssnext-playground/

ai:
OK, I take some parts of it.

ai:
Done. I added 2 examples to README.

@davidtheclark can I ask you to make shorter version of PostCSS description in 16—21 lines? I think we can remove 2—3 lines there.

davidtheclark:
Over at #212 I had this going:

```
PostCSS powers a growing ecosystem of modular JS plugins that you can use to enhance your
CSS workflow. These plugins can add vendor prefixes, transpile CSS4+ syntax, support variables
and mixins, inline images, and more.

PostCSS itself parses CSS into an AST (tree of nodes) and offers a simple API for
reading and transforming that AST.
```

What do you think about integrating that text?

ai:
I afraid that first section (about plugins) is longer, that we have now.

But the second section (about AST) is a much cleaner, that we have right now.

PR #205, opened 2015-02-15
Edits from Root node to end

davidtheclark:
I made it all the way through. Some edits are above Root node because I tried to clean up the links a little bit, to make it easier to just throw in links to different parts of the documentation,  and address some of the clarifications you made.

### DIFF
--- a/API.md
+++ b/API.md
@@ -35,8 +35,6 @@ You can also set plugins with the [`PostCSS#use`] method.
 
 See [`PostCSS#use`] below for details about plugin formats.
 
-[`PostCSS#use`]: https://github.com/postcss/postcss#useplugin
-
 ### `postcss.parse(css, opts)`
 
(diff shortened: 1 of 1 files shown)


ai:
Nice. Common part with links is very useful to maintain crosslinks.

davidtheclark:
I forgot to bring up two points that I was thinking about:

- There is no section for `root.toResult()`. Could you add one?
- I was wondering why the `semicolon` properties on `atrule` and `rule` return `true`/`undefined` instead of `true`/`false`. Would you mind explaining the intention?

ai:
Parser just not add `semicolon` property, when we don’t need it. Reason is to make tests and inspect cleaner.

ai:
Oops, sorry, I forgot about `toResult`. Done 2f841fa

PR #206, opened 2015-02-15
Move `defaultStyle` object from prototype to constructor

vanesyan:
It's antipatterns to add to prototype anything else except function

### DIFF
--- a/API.md
+++ b/API.md
@@ -566,7 +566,7 @@ root.nodes[0].nodes[1].style('before') //=> ' '
 ```
 
 If PostCSS can’t find any nodes to copy the code style property from,
-it will use `defaultType` value from `node.defaultStyle` object.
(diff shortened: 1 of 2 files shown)


API.md
```diff
@@ -566,7 +566,7 @@ root.nodes[0].nodes[1].style('before') //=> ' '
 ```
 
 If PostCSS can’t find any nodes to copy the code style property from,
-it will use `defaultType` value from `node.defaultStyle` object.
+it will use default from [`const defaultStyle`](https://github.com/vanesyan/postcss/blob/patch-3/lib/node.js#L4-16)
```
lydell: That link is odd. Should be `./lib/node.js#L4-16` I think.

ai: I removed `defaultStyle` from docs. Anyway it is too complicated and plugin developers should not use it.

vanesyan: @lydell This text block was already deleted

PR #207, opened 2015-02-15
Simplify `if` statements

vanesyan:


ai:
`if ( foo )` and `if ( typeof(foo) == 'undefined' )` is not the same. Remeber, that empty string `"" == false` in JS. Even `0 == false` in JS.

So, this change broke the code.

vanesyan:
@ai sorry my fault)

PR #212, opened 2015-02-16
Update top of the README to make a more clear and accessible introduction

MoOx:
I tried to reorder ideas a little bit, so people understand quickly what is postcss core, and what can offer its ecosystem.

I've deliberately remove some later part, placed with shorter explanation at the top. Tell me what you think guys.

I have lot of other changes in mind, but I think this is a good a quick start ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,26 +2,26 @@
 
 <img align="right" width="95" height="95" src="http://postcss.github.io/postcss/logo.svg" title="Philosopher’s stone, logo of PostCSS">
 
-PostCSS is a tool for transforming CSS with JS plugins. These plugins can add vendor
-prefixes, support variables and mixins, transpile CSS4 syntax, inline images, and more.
+PostCSS is a tool to manipulate and transform CSS through JavaScript. PostCSS parse latest CSS syntax (even some W3C drafts if they seems stable enough) into an JavaScript AST (tree of nodes) and offers you some accessibles APIs to play with it.
 
-PostCSS can do the same work as "preprocessors" like Sass, Less, and Stylus. But
-PostCSS is modular, 4-40x faster, and much more powerful.
+In addition to the available APIs, PostCSS allows you to chain manipulations via a system of plugins.
 
-PostCSS is used by Google, Twitter, Alibaba, and Shopify. Its most popular plugin,
-[Autoprefixer], is one of the most universally praised CSS processors available.
+_You can easily use PostCSS only via its ecosystem of plugins._  
+Plugins can add vendor prefixes, transform CSS 4+ syntax to more compatible CSS, inline images, add pre-processor like variables or mixins support, etc.
 
-PostCSS itself is very small and focused. It includes a CSS parser, a CSS node
-tree API, a source map generator, and a node tree stringifier — and that's it.
-All of the stylesheet transformations it powers are encapsulated in modular plugins.
-And these plugins are themselves small and focused plain JS functions, which typically
-receive a CSS node tree at one end, apply transformations to it, and return
-a modified tree that can be used by other plugins or written to a file.
+With its ecosystem, PostCSS with a bunch of plugins might do the same job as Sass, LESS or Stylus.
+The interesting part is that PostCSS is modular, and instead of relying on a monolitic tool,
+you can tweak your workflow very easily for yourself and your team.
+The icing on the cake is that you can write your own plugins to make your own transformations.
 
-Individual plugins implement specific transformations; and by using various plugins
-together you can create an ideal CSS workflow for yourself and your team.
-For instance, if you like the power provided by preprocessors
-like Sass, you could combine [Autoprefixer], [cssnext], [CSS Grace], [postcss-nested],
+Note that PostCSS with some plugins can be 4x to 40x times faster, that popular pre-processors.
+
+_PostCSS is used by Google, Twitter, Alibaba, and Shopify. Its most popular plugin,
+[Autoprefixer], is one of the most universally praised CSS processors available._
+
+---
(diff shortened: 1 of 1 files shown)


ai:
Ouh, I commit many typography changes to master (sorry, it is my profession deformation after working in Wikipedia :) ).

ai:
I agree, that shorter main section is important.

First line description is incredible important. We should describe there, why PostCSS is useful and what PostCSS key feature. So I think that we should mention “JS plugins” and key tasks right in the first paragraph.

MoOx:
What about

> offers you some accessibles APIs to play with it **using plugins** ?

ai:
Current structure is:

1. Short description of all README.md.
2. Is PostCSS is important, does big companies use it.
3. “Yes, we a like preprocessors, but better”, because it is main question.
4. Short description how PostCSS works.
5. Code example.
6. Links.

We shoould not mix this section.

README.md
```diff
@@ -2,26 +2,26 @@
...
-together you can create an ideal CSS workflow for yourself and your team.
-For instance, if you like the power provided by preprocessors
-like Sass, you could combine [Autoprefixer], [cssnext], [CSS Grace], [postcss-nested],
+PostCSS is a tool to manipulate and transform CSS through JavaScript. PostCSS parse latest CSS syntax (even some W3C drafts if they seems stable enough) into an JavaScript AST (tree of nodes) and offers you some accessibles APIs to play with it.
+
+In addition to the available APIs, PostCSS allows you to chain manipulations via a system of plugins.
+
+_You can easily use PostCSS only via its ecosystem of plugins._  
+Plugins can add vendor prefixes, transform CSS 4+ syntax to more compatible CSS, inline images, add pre-processor like variables or mixins support, etc.
+
```
davidtheclark: Sounds like the idea is to describe, as briefly as possible, two different things: PostCSS core (parser/api) and PostCSS plugin ecosystem. How about this (trying to be as concise as I can be)?

> PostCSS parses CSS into a JavaScript AST (tree of nodes) and offers a simple API for reading and transforming that AST. PostCSS powers a growing ecosystem of plugins that you can use to enhance your CSS workflow. Plugins can add vendor prefixes, transpile CSS4+ syntax, support variables and mixins, inline images, and more.

README.md
```diff
@@ -2,26 +2,26 @@
...
+With its ecosystem, PostCSS with a bunch of plugins might do the same job as Sass, LESS or Stylus.
+The interesting part is that PostCSS is modular, and instead of relying on a monolitic tool,
+you can tweak your workflow very easily for yourself and your team.
+The icing on the cake is that you can write your own plugins to make your own transformations.
+
+Note that PostCSS with some plugins can be 4x to 40x times faster, that popular pre-processors.
```
davidtheclark: I'm not that we need to make this section *longer*. This new text elaborates on the "modular" feature ... maybe instead we should just link to the Feature section?

> PostCSS can do the same work as preprocessors like Sass, Less, and Stylus. But PostCSS is modular, 4-40x faster, and much more powerful. Learn more about the advantageous [Features](#features) of PostCSS below.

README.md
MoOx: I think this 2 thinks should have 2 paragraphs.

MoOx: But maybe a really short version is a good idea.

davidtheclark: Sure. Maybe we should put the plugin stuff first because most readers are probably interested in that?

```
PostCSS powers a growing ecosystem of modular plugins that you can use to enhance your
CSS workflow. These plugins can add vendor prefixes, transpile CSS4+ syntax, support variables
and mixins, inline images, and more.

PostCSS itself parses CSS into a JavaScript AST (tree of nodes) and offers a simple API for
reading and transforming that AST. 
```

davidtheclark:
I wonder also, given the talk about adding at least one more example, if the current example should be cut from the top and moved to an Examples section along with at least on more. That would help make the top short.

MoOx:
The current README do not even mention an "ecosystem" in the first lines, cause that's the important part. Most users will just consume plugins.

ai:
I'm. I missed why mention of ecosystems is important?

MoOx:
Because new users you want to get are not likely to start using postcss with only custom plugins. Most of them are going to use the ecosystem (already existing plugins).

ai:
Good point. But second sentence mentions existed plugins.

davidtheclark:
I think I agree with @MoOx that the ecosystem of plugins really is probably the most important part for almost all visitors, so it might be a good idea to mention it first (maybe use the sentences above? "PostCSS powers a growing ecosystem of modular JS plugins that you can use to enhance your CSS workflow. These plugins can add vendor prefixes, transpile CSS4+ syntax, support variables and mixins, inline images, and more." --- since I also think it helps to mention that it's "growing" and to use the "enhance your CSS workflow" phrase to make sure that we're not limiting plugins to transformations.)

ai:
Yeap, we could mention it. But right now this 3 sentence firm is too long and too difficult. Main section is very important, we must fight to remove any word. I think we should have 2 sentence, just add a ecosystem weird to second sentence.

davidtheclark:
How about replacing first paragraph with:

> PostCSS is a tool for transforming CSS with JS plugins. The growing ecosystem of PostCSS plugins can add vendor prefixes, support variables and mixins, transpile CSS4 syntax, inline images, and more.

(only addition is "The growing ecosystem of PostCSS")

Do you both think that might work?

ai:
Awesome. Please send a PR.

ai:
@davidtheclark I make a commit already by your name

PR #220, opened 2015-02-19
There is no such thing as CSS4

scottkellum:
http://www.xanthir.com/b4Ko0


MoOx:
We already discuss about that somewhere.

It's not better since stuff like custom props are level 1.

We should found another formula to express this idea

Here is some ideas

- "Future-proof CSS"
- "tomorrow CSS syntax"
- "new CSS syntax"

scottkellum:
`Future CSS` is one [I have seen around](https://twitter.com/vinspee/status/568212877259227136).

PR #228, opened 2015-02-24
Sets prev to false when it is 'falsy'

mAiNiNfEcTiOn:
When using gulp-autoprefixer, I somehow managed to get a corrupt sourcemap (or something), when switching between branches which, when reaching to the line 61 (current master state)

```javascript
if ( prev === false ) return;
```

prev was _undefined_, which then would try to _JSON.parse_ it and would throw a SyntaxError saying:

```
SyntaxError: unexpected token /
```

... Problem was solved by removing the node_modules package and re-do ```npm install```... which suggests, at least, a cache problem.

### DIFF
--- a/lib/previous-map.js
+++ b/lib/previous-map.js
@@ -58,6 +58,8 @@ export default class PreviousMap {
 
     // Load previous map
     loadMap(file, prev) {
+
+        prev = prev || false;
         if ( prev === false ) return;
 
         if ( prev ) {


ai:
Sorry, but API requires `false`. `undefined` means try to autodetect previous source map (you can see this by broken tests).

mAiNiNfEcTiOn:
Ok then :) Then you should consider that, if a previous sourcemap is corrupted and you try to JSON.parse it and it fails, you should assume it is not available and not use it. Throwing an error saying "Unexpected token /" is not the solution, IMHO.

Thanks for the time wasted. Keep up the good work.

ai:
Yeap, `try { } catch { }` will be nice solution.

ai:
But how user will know about broken map?

ai:
We can print warning, but warning is easy to miss in current build tools with big output logs

mAiNiNfEcTiOn:
Well, point taken... But then, you should be more explicit in what the problem is. I had to dig in autoprefixer's code to understand it was a sourcemap problem and still, I didn't know where he was getting it from.

So ideally a "Your sourcemap appears to be broken. Check it at <path>"

ai:
I made a separated taks #229

PR #230, opened 2015-02-25
Use semver to match postcss version

w0rm:
In this PR I suggest to use [`semver.satisfies`](https://github.com/npm/node-semver) to match postcss version.

### DIFF
--- a/lib/postcss.js
+++ b/lib/postcss.js
@@ -38,9 +38,8 @@ postcss.plugin = function (name, initializer) {
             try {
                 return transformer.apply(this, arguments);
             } catch (e) {
-                var a = result.postcssVersion.split('.');
-                var b = processor.version.split('.');
-                if ( a[0] != b[0] || parseInt(a[1]) > parseInt(b[1]) ) {
+                import semver from 'semver';
+                if (!semver.satisfies(processor.version, result.postcssVersion)) {
                     if ( typeof(console) != 'undefined' && console.warn ) {
                         console.warn(result.postcssPlugin + ' is based on ' +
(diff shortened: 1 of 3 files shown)


ai:
Nice!

But why you stub `semver` in tests?

w0rm:
I fake semver results because I'm not testing semver in this test.

ai:
But you can missunderstand semver API ;).

w0rm:
I assume that `plugin.postcssVersion` to be a valid semver range, e.g. `>=2.5.0`.

w0rm:
I didn't change version examples in tests, but now I think I had to. Here are some examples of  `semver.satisfies`:

```
semver.satisfies('2.5.0', '2.5.1') === false
semver.satisfies('2.5.1', '2.5.0') === false
semver.satisfies('2.5.5', '^2.5.0') === true
semver.satisfies('2.5.5', '>=2.5.0') === true
semver.satisfies('2.5.5', '2.5.*') === true
```

ai:
`plugin.postcssVersion` is not a range. It is a PostCSS version, which is used to build plugin.

w0rm:
Oh, then I didn't get this right. I thought each plugin would have to provide version range of postcss.
Sorry, please roll back my change.

ai:
I fix it, no problem. I just waiting for `npm install` :).

ai:
Unfortunately, I revert commit. Because some of users uses PostCSS in client-side and extra library without any benefits is a bad solution.

w0rm:
Yeah, I agree to this! Plus I didn't understand the right behavior, but now I got it.

PR #248, opened 2015-03-14
Add postcss-normalize-url.

ben-eb:
https://github.com/ben-eb/postcss-normalize-url


ben-eb:
Awesome, thanks! :smiley:

PR #255, opened 2015-03-22
Parse spaces & comments inside important. Fixes #253.

### LINKED ISSUE #253: Doesn't parse excessive whitespace around !important as an important value
This is considered valid CSS in browser:

```css
body {
    color:#000    !     important;
}
```

However when examining the `declaration` it does not have the `important` key.

```
{ type: 'decl',
  parent: 
   { type: 'rule',
     nodes: [ [Circular] ],
     parent: 
      { type: 'root',
        nodes: [Object],
        source: [Object],
        semicolon: false,
        after: '\n',
        lastEach: 18,
        indexes: [Object] },
     source: { start: [Object], input: [Object], end: [Object] },
     before: '',
     between: '',
     selector: 'body',
     semicolon: true,
     after: '\n',
     lastEach: 18,
     indexes: { '18': 0 } },
  source: 
   { start: { line: 2, column: 5 },
     input: 
      { css: 'body {\n    color:#000    !     important;\n}\n',
        safe: false,
        id: '<input css 1>',
        from: '<input css 1>' },
     end: { line: 2, column: 34 } },
  before: '\n    ',
  prop: 'color',
  between: ':',
  value: '#000    !     important' }
```

ben-eb:
So now we can parse a declaration such as this correctly:

```css
h1 {
    color: black      !    /* test comment */   important;
}
```

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -229,7 +229,17 @@ export default class Parser {
                 string = this.spacesFromEnd(tokens) + string;
                 if ( string !== ' !important' ) node._important = string;
                 break;
-            } else if ( token[0] !== 'space' && token[0] !== 'comment' ) {
+            } else if (token[1] === 'important') {
+                let tokensCache = tokens.slice(0);
+                let important = this.stringTo(tokensCache, i, '!');
+                if (important.trim().indexOf('!') === 0) {
+                    node.important = true;
+                    node._important = important;
+                    tokens = tokensCache.slice(0);
+                }
+            }
+
+            if ( token[0] !== 'space' && token[0] !== 'comment' ) {
                 break;
             }
         }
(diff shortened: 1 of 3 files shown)


ai:
Will not work with :)

```js
a {
  value: important
}
```

But I am very happy, that somebody understand parser :).

ben-eb:
Ah, good catch; what do you think of a `hasToken` method to ensure that the declaration had both `!` and `important`?

Also, if you could help me understand how to unit test it I could cover these cases. :smiley:

ai:
Here is a tests for parser: https://github.com/postcss/postcss/blob/master/test/parse.js#L48

Test compares AST from parsing `test/cases/X.css` with JSON from `test/cases/X.json`.

I suggest you to add new case to [important.css](https://github.com/postcss/postcss/blob/master/test/cases/important.css) and then run `gulp cases`. This Gulp task will all JSON files by CSS AST (like take a snapshot from current parse output AST).

ben-eb:
@ai I've added test cases for this. Any other changes I should make? :smile:

ai:
@ben-eb awesome! But could you also add test for `prop: important`?

ben-eb:
Done. :+1:

ai:
Awesome. It will be OK if I will accept your PR after tuesday? I am very busy on workshop right now :'(

ben-eb:
Of course! Will just squash the history down. :smile:

PR #263, opened 2015-03-25
Helpful error message for when adding to containers

philip-peterson:
Hello!

I think it might be helpful to provide the developer with a nice error message when a declaration is attempted to be added to a container like so:

```
rule.append({
   prop: 'transform',
   vlaue: 'translate(-50%, -50%)'
});
```

In this case, the developer has incorrectly specified the value of the rule under something other than 'value' key. Currently the error message is:

```Cannot call method 'indexOf' of undefined```

with no discernible source or traceback (at least through Grunt). 

This patch adds some nicer diagnostics, including what I hope to be a more helpful error message:

```Could not construct declaration; 'value' field unspecified.```, 

as well as an additional check for the `prop` key. 

Corresponding tests for the test suite are included.

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -411,6 +411,9 @@ export default class Container extends Node {
             } else if ( nodes.type ) {
                 nodes = [nodes];
             } else if ( nodes.prop ) {
+                if (nodes.value === undefined) {
+                    throw this.error("Could not construct declaration; 'value' field unspecified.");
+                }
                 nodes = [new Declaration(nodes)];
             } else if ( nodes.selector ) {
                 import Rule from './rule';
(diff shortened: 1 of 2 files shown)


philip-peterson:
My apologies for the failing Travis CI build... I cannot seem to reproduce the lint failure on my machine for some reason (`grunt`, `grunt test`, and `grunt lint` all succeed locally). Any suggestions for how to troubleshoot that issue would be much appreciated. 

Nevertheless, it appears to be a purely cosmetic error (single quotes preferred over double quotes). Once I have it figured out I'll submit another patch to fix the quotes.

ai:
Thanks for PR.

No problem, I fixed linter warnings b0cfc8f

philip-peterson:
No problem!

PR #274, opened 2015-04-02
Add postcss-colormin.

ben-eb:
https://github.com/ben-eb/postcss-colormin


MoOx:
Nice module !
I would call this module postcss-reduce-color to be clear about what it does :)

ben-eb:
@MoOx I thought about perhaps changing the name of it, however for plugins like this that are essentially wrapping a module I think it's clearer to use the parent module name. :smiley:

PR #285, opened 2015-04-05
Move es6-promise to prod dependencies

antyakushev:
I guess this needs to be a production dependency, at least my node 0.10 build fails otherwise.


ai:
Oops

PR #288, opened 2015-04-08
Add postcss-log-warnings to plugin list

davidtheclark:


ai:
Awesome!

Only note that I order plugin links by name length and respect the Holy 80 Columns Line :).

davidtheclark:
Ah. I was wondering how those links were ordered.

PR #295, opened 2015-04-12
Update runner guidelines.

ben-eb:
I've rewritten the guidelines to use a clear, more consistent tone. :smile:

### DIFF
--- a/Runner Guidelines.md
+++ b/Runner Guidelines.md
@@ -1,11 +1,11 @@
 # PostCSS Runner Guidelines
 
-PostCSS runner is a tool, that takes plugins list from user and process
-user’s CSS through this plugins. For example, [`postcss-cli`]
-or [`gulp-postcss`]. This rules are mandatory for any PostCSS runners.
+A PostCSS runner is a tool that processes CSS through a user-defined list of
+plugins; for example, [`postcss-cli`] or [`gulp-postcss`]. These rules are
+mandatory for any such runners.
 
-This rules are not mandatory, but highly recommended for one-plugin tools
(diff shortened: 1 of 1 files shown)


Runner Guidelines.md
```diff
@@ -18,8 +18,8 @@ See also ClojureWerkz’s [open source project recommendations].
 
 ### 1.1. Accept functions in plugin parameters
 
-Some plugins accept function in parameters. If your runner has config file,
-it should be a Node.js module, because JSON format doesn’t support functions.
+PostCSS runners which use config files must be written in JavaScript, so that
+they can support plugins which accept a function, such as [`postcss-assets`]:
```
lydell: It is no longer clear that the _config_ files should be written in JS. It sounds more as if the runners should be.

Runner Guidelines.md
```diff
@@ -31,24 +31,26 @@ module.exports = [
...
 ```
 
-### 2.2. Use only asynchronous API
+### 2.2. Use only the asynchronous API
 
-PostCSS has synchronous API only for debug. Synchronous API is slower
-and can’t work with asynchronous plugins.
+PostCSS runners must use only the asynchronous API. The synchronous API is
+provided only for debugging, is slower, and can't work with asynchronous
+plugins:
```
lydell: I think it was better without the colon here. Now it looks like it is going to be demonstrated why the sync API is discouraged.

lydell:
Nice work!

ben-eb:
Good points, thanks. Will update. :+1:

PR #313, opened 2015-04-27
Update README.md

cvializ:
fs.fileWriteSync is not part of the Node fs API, so change it to the correct function name.

I had a weird copy-paste error from this!


TrySound:
@cvializ Well done!

PR #323, opened 2015-05-04
Add postcss-css-variables plugin

MadLittleMods:
[`postcss-css-variables`](https://github.com/MadLittleMods/postcss-css-variables) transforms [CSS Custom Property syntax(CSS variables)](http://dev.w3.org/csswg/css-variables/) into static representation that can be used today.

[Try it out on the code playground](http://localhost/random/postcss-test1/node_modules/postcss-css-variables/playground/) or checkout the [usage section in the readme](https://github.com/MadLittleMods/postcss-css-variables#usage).

It works out the proper variable scope for descendant or nested rules/selectors to assign the associated value.

It also works with `@rules` such as media queries and updates any property that references the variable when the media query is active.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ on different parts of CSS.
 * [`postcss-custom-media`] supports custom aliases for media queries.
 * [`postcss-custom-properties`] supports variables, using syntax from
   the W3C Custom Properties.
+* [`postcss-css-variables`] supports variables for descendant/nested rules and at-rules
 * [`postcss-custom-selectors`] adds custom aliases for selectors.
 * [`postcss-font-variant`] transpiles human-readable `font-variant` to more
   widely supported CSS.
@@ -305,6 +306,7 @@ on different parts of CSS.
 [`postcss-discard-empty`]:       https://github.com/ben-eb/postcss-discard-empty
 [`postcss-reduce-idents`]:       https://github.com/ben-eb/postcss-reduce-idents
 [`postcss-simple-extend`]:       https://github.com/davidtheclark/postcss-simple-extend
+[`postcss-css-variables`]:       https://github.com/MadLittleMods/postcss-css-variables
 [`postcss-selector-not`]:        https://github.com/postcss/postcss-selector-not
 [`postcss-default-unit`]:        https://github.com/antyakushev/postcss-default-unit
 [`postcss-media-minmax`]:        https://github.com/postcss/postcss-media-minmax


MoOx:
I am again this merged in postcss readme (**for now**): the README of this module mention mutliple times that it can do thing it cannot. It doesn't warn at all about dangerous things it create.
I already found some issues and I am going to take the time to open issues on this repo when I get time.
Also, the safe feature this plugin try to support will be integrated in postcss-custom-properties and will properly warn users when they are at the limit of the specs by using postcss messages api.

PS: it also use horribly slow regex https://github.com/MadLittleMods/postcss-css-variables/blob/master/index.js#L26 and postcss community try to avoid that at all cost.
PS2: It also violate the MIT license (that should mention Nicolas Gallagher, TJ Holowaychuk and myself for fixtures)

MadLittleMods:
#### Dangerous

I may be missing something but I am not sure of the danger that you mentioned in regards to how this plugin transforms CSS.

There is currently a [bug open](https://github.com/MadLittleMods/postcss-css-variables/issues/2) which I will be fixing promptly.

#### Slow regex

The regex is long to properly split out any CSS selector into its descendant pieces. As mentioned in the comments of that piece of regex. This almost works `/\b\s(?![><+~][\s]+?)/` but the word boundary `\b` doesn't work with attribute selectors.

I could instead use a separate/proper CSS parsing engine library to split it into pieces and although I haven't explored this area, I am not sure on the performance benefits. This definitely an area to look into the future.

Just as a ballpark-type figure with nothing to compare to. Here is a jsPerf *with only one test case* and I am getting ~35-40k ops/second in Chrome. Firefox does about ~20k ops/second. I could also be doing something terribly wrong in this test case jsPerf with caching, engine optimizing the calculation out, etc: http://jsperf.com/css-selector-split-at-descendant-combinator



#### License:

As mentioned in the [readme](https://github.com/MadLittleMods/postcss-css-variables#why):

> I do use some of the same/similar unit tests because that just makes sense to get proper code coverage.

We talked about this on gitter and you said you weren't really worried about it. I have no issue adding some form of attribution to the license but what do you suggest to cover the similar fixtures?

ai:
@MoOx many plugins has limitation and we anyway mentions them in README.md.

ai:
@MadLittleMods can you change commit:

* `at-rules` instead of `@rules`
* links is sorted by plugin name length

ai:
@MoOx I am not happy with forks, but right now I saw no rules to dismiss it.

MadLittleMods:
@ai Update commit:

 - Changed wording: `@rules` to `at-rules`
 - Moved link in accordance to plugin name length

MoOx:
I just take half an hour to take a deeper look and open some issues.

https://github.com/MadLittleMods/postcss-css-variables/issues

@MadLittleMods What you are trying to promote for potentials users looks totally wrong/buggy :/
For license I don't care for real, but maybe @necolas & @tj don't (see https://github.com/postcss/postcss-custom-media/issues/3 ). It's the deal when you resuse someone work. Maybe you completely rewrite the code, but you completely keep all the ideas, even undocumented options. So it's just no fair for all people that have worked on this. 
Related: http://stackoverflow.com/questions/26033524/licensing-under-mit-a-library-ive-rewritten-and-renamed-thats-already-under

@ai limitation and bugs are totally not clear in the README, that's my point. @MadLittleMods write in the README things that are unsupported or completely buggy. See opened issues.

Adding this in the "Future CSS Syntax" section seems wrong to me according to the bugs that will be generated by this module, over and over (only a client-side polyfill might handle custom props/vars with cascade correctly).

PR #326, opened 2015-05-05
Add postcss-conditionals plugin

andyjansson:
https://github.com/andyjansson/postcss-conditionals


tunnckoCore:
very cool! :+1: before few weeks search for something like that :)

andyjansson:
Thanks :)

PR #327, opened 2015-05-06
Add postcss-cssstats plugin to the readme

johnotander:


tunnckoCore:
omg.. why you would need to use it as postcss plugin, when you can just in 99% of the time

```js
var result = postcss().process(input).css
cssstats(result)
```

lol!

ai:
@tunnckoCore in your example you will parse CSS two times

tunnckoCore:
and correct me if im wrong, but cssstats also use postcss, which is kind-a-weird.

tunnckoCore:
and the plugin also would parse twice. https://github.com/cssstats/css-statistics/blob/master/index.js#L16

tunnckoCore:
in plugin approach

```js
var postcss = require('postcss');
var cssstats = require('postcss-cssstats');

var output = postcss([cssstats(function(stats) {
    console.log(stats)
  })])
  .process(css);
```

1. postcss would parse the css and pass it to the given plugins
2. secondly, css plugin will get that css and parse it again

its awful idea to have plugin, at all.

johnotander:
@tunnckoCore `postcss-cssstats` is currently only a `0.0.1` because `cssstats` is being reworked to either take a CSS string _or_ a PostCSS AST. So, once that functionality is added (in the next day or so), it will only parse the CSS once when an AST is passed.

Thus, it will serve appropriately as a plugin, providing a stats object in each build where you can `console.log` any output you'd like. Ideally this will be used to warn on overly specific selectors, and print number of IDs, classes, universal selectors, etc.

PR #332, opened 2015-05-10
Add postcss-merge-idents.

ben-eb:
https://github.com/ben-eb/postcss-merge-idents


ai:
BTW. I think to remove plugins from packs from this list. It became too big and I think there is no sense to use cssnano/cssnext plugins one by one.

ai:
Or maybe move them to sublists in Pack section

PR #354, opened 2015-06-01
add postcss-class-prefix

thompsongl:
Add postcss-class-prefix plugin to the readme


ben-eb:
Looks good to me. Thanks for your contribution. :smile:

PR #359, opened 2015-06-04
Add postcss-vertical-rhythm plugin

markgoodyear:


ai:
Interesting idea. I will accept it, when I will be at home.

BTW, why not just using `rem`?

markgoodyear:
Awesome. I guess it's a more simplified version of Compass's `rhythm()` function. I use my own custom function to do it in Sass (without Compass), but having a custom unit is much more CSS like. Easier to write too!

As for `rem` output, I figured when paired with [postcss-pxtorem](https://github.com/cuth/postcss-pxtorem), it wasn't necessary to have a conversion, however open to ideas :).

PR #360, opened 2015-06-06
Allow double underscores in prop names

geelen:
In writing the spec for [CSSI](https://github.com/css-modules/css-modules/blob/master/INTERCHANGE_FORMAT.md), we wanted a convention to identify unlinked variables within a file. The convention we came up with was:

```css
:import("./dependency.css") {
  __localTemporaryName: exportedNameFromDependency;
}
```

The use of double-underscores is not essential to the spec by any means, but desirable from a readability standpoint. When attempting to parse this file with PostCSS, a property starting with a `_` is treated as an IE6 hack, and the `decl.prop` variable no longer matches what was written.

This PR simply only applies the IE6 hack logic if the *first* character is an underscore **but the second one isn't**. Tests are included to confirm the check is applying correctly.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -209,7 +209,7 @@ export default class Parser {
             }
         }
 
-        if ( node.prop[0] === '_' || node.prop[0] === '*' ) {
+        if ( (node.prop[0] === '_' && node.prop[1] !== '_') || node.prop[0] === '*' ) {
             node.before += node.prop[0];
             node.prop    = node.prop.slice(1);
         }
--- a/test/cases/prop-hacks.css
+++ b/test/cases/prop-hacks.css
@@ -2,4 +2,5 @@ a {
     *color : black;
     _background: white;
     font-size/**/: big;
+    __custom: var;
 }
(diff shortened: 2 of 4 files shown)


ben-eb:
I am not really a fan of this proposed change, because it is not a part of the CSS grammar. Typically such leading characters are used by some browser developers as property hacks.

Custom parsers will be arriving in 4.2 however to support this use case.

What do you think, @ai?

ai:
Maybe we should add `__` to a property name?

ben-eb:
I don't think I ever liked the API moving non-space characters out of the property key and into before. So I'd be happier if the full property name included hacks and all non whitespace tokens. But this should be looked at in v5 as it is a breaking change.

ai:
@ben-eb current PostCSS public API is to parse any CSS and save errors inside. “Parse do not lint”, because we have separated linter tools. There is parse can parse why nested rules.

@geelen can you wait a few days while we will think about best API?

geelen:
@ai no worries! I realised this might be a contentious fix, and for now we can simply use a different prefix (`i__` instead of `__`), but it was the first time I'd seen a declaration property *change* when parsed, so I wanted to get in there and see if I could fix it.

PR #372, opened 2015-06-17
Add adapters section + PostStylus link to README

seaneking:
https://github.com/seaneking/poststylus

Feel free to reject this one if you feel it doesn't fit in the current PostCSS ecosystem, but I've made a little Stylus -> PostCSS adapter that I think would be handy for a lot of PostCSS converts. It loads an array of PostCSS plugins into Stylus' generated css, then passes the post-processed css back to Stylus for final compiling to file, similar to the way that [autoprefixer-stylus](https://github.com/jenius/autoprefixer-stylus) works.

Wasn't sure where this would fit in the readme, so added a new sub-section under plugins and put it there. Let me know if you want this changed (since I guess it's not technically a postcss plugin).

Use-case is people like me, who often still prefer to author stylesheets with a preprocessor like Stylus but want to move all the heavy-lifting/grunt-work utilities to PostCSS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,10 @@ See also plugins in modular minifier [`cssnano`].
 * [`postcss-canadian-stylesheets`] Canadian Style Sheets.
 * [`postcss-spiffing`] lets you use British English in your CSS.
 
+### Adapters
+
+* [`poststylus`] PostCSS adapter for Stylus output
+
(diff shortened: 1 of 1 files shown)


ai:
Adapters is not a plugin in PostCSS terms. It is a runner. So can you send me other PR with your awesome tool in Usage section? :)

seaneking:
done ;-)

PR #374, opened 2015-06-17
fix native gitter badge

iamstarkov:
Every time I see PostCSS readme, I wanna cry because of "strange" native gitter badge. let’s use custom badge!

Text, color and style are configurable. Probably, we can use modern _flat-square_ style. Take a look at [shields.io page](http://shields.io/)


ai:
Yep, very strange that Gitter native badge s not compatible with Travis CI.

PR #375, opened 2015-06-17
Add postcss-hexrgba to README

seaneking:
https://github.com/seaneking/postcss-hexrgba

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,7 @@ See also plugins in modular minifier [`cssnano`].
 * [`postcss-easings`] replaces easing names from easings.net
   with `cubic-bezier()` functions.
 * [`postcss-focus`] adds `:focus` selector to every `:hover`.
+* [`postcss-hexrgba`] adds shorthad hex methods to rbga() values.
 * [`postcss-generate-preset`] allows quick generation of rules.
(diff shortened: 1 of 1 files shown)


ai:
I thought that we have this feature in cssnext. Do we?

MoOx:
This is not valid CSS.
You can do that in cssnext by using `color(#fff a(.2))` (via postcss-color-function)

ai:
OK, I will accept PR when I will be in hotel. I think we need to fix spellchecker dictionary to fix a Travis

seaneking:
yep not aiming to be to-spec css, just a shortcut for those already using rgba(). But just saw https://github.com/avanes/postcss-color-alpha which is similar, just a different syntax. Up to you whether you want to include this variation in the readme or not, as it is quite similar (I personally prefer adding to the existing rgba() format, rather than adding a new decl/value)

ben-eb:
You have two typos in there; `rbga` and `shorthad`.

seaneking:
derp, sorry I'll fix it.

seaneking:
erroring on rgba (suggests rgb), so I guess the spellchecker would need to be updated.

PR #377, opened 2015-06-19
Optimize and idiomantify tokenizer

vanesyan:
Some style correction, and handwrite char codes

### DIFF
--- a/lib/tokenize.js
+++ b/lib/tokenize.js
@@ -169,7 +169,8 @@ export default function tokenize(input) {
                 break;
 
             default:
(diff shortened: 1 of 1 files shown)


ai:
Yeap, seems like good thing :)

ai:
I will check it in weekend and maybe take a benchmarks result (does browsers works with const in different way?).

vanesyan:
In theory it should works faster, than with `let` keyword, because VM knows that the const variable will not change and size of it never change.

ai:
Babel compiled `const` to `var` so we can’t see changes right now :)

PR #379, opened 2015-06-20
Add postcss-clearfix to README

seaneking:
https://github.com/seaneking/postcss-clearfix

Similar functionality already covered in https://github.com/cssdream/cssgrace but I wanted this as a standalone plugin.


ben-eb:
:+1:

PR #384, opened 2015-06-23
Support multiple nodes on DOM4 API methods

jonathantneal:
[DOM4](https://dom.spec.whatwg.org/) mutation methods accept multiple elements. Feature parity would be helpful between DOM4 and postcss.

- Supports multiple nodes on [append](https://dom.spec.whatwg.org/#dom-parentnode-appendnodes), [prepend](https://dom.spec.whatwg.org/#dom-parentnode-prependnodes), and [replaceWith](https://dom.spec.whatwg.org/#dom-childnode-replacewithnodes).
- Adds tests.
- Updates API documentation.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1036,19 +1036,15 @@ if ( decl.prop.match(/^-webkit-/) ) {
 }
 ```
 
-### `node.replaceWith(otherNode)`
+### `node.replaceWith(otherNodes...)`
 
-Inserts another node before the current node, and removes the current node.
+Inserts nodes before the current node, and removes the current node.
 
 ```js
 if ( atrule.name == 'mixin' ) {
     atrule.replaceWith(mixinRules[atrule.params]);
 }
 ```
-
-Arguments:
-
-* `otherNode: (Node)`: other node to replace current one.
 
 ### `node.clone(props)`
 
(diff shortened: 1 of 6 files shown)


ai:
I can’t accept it right now. Only as part of 5.0 API changes.

ai:
Or maybe as part of 4.2, because it will not broke a API.

ai:
Should we remove in 5.0 also arrays? `append(array)` to use only this method?

docs/api.md
```diff
@@ -1036,9 +1036,9 @@ if ( decl.prop.match(/^-webkit-/) ) {
 }
 ```
 
-### `node.replaceWith(otherNode)`
+### `node.replaceWith(otherNodes...)`
```
lydell: The dots should be on the other side. This is JavaScript, not CoffeeScript.

lib/at-rule.js
```diff
@@ -27,14 +27,14 @@ export default class AtRule extends Container {
         }
     }
 
-    append(child) {
+    append() {
```
lydell: Why not use `...children`? Performance?

lib/at-rule.js
```diff
@@ -27,14 +27,14 @@ export default class AtRule extends Container {
...
         if ( !this.nodes ) this.nodes = [];
-        return super.append(child);
+        return super.append.apply(this, arguments);
```
lydell: Same thing here. Why not `super.append(...children)`?

lib/container.js
```diff
@@ -150,18 +150,35 @@ export default class Container extends Node {
         });
     }
 
-    append(child) {
-        let nodes = this.normalize(child, this.last);
-        for ( let node of nodes ) this.nodes.push(node);
+    append() {
+        for (let index in arguments) {
```
lydell: A `for-in` to loop over an array-like object? Really?

jonathantneal:
@lydell, thank you for showing me all of these wonderful ES6 goodies. If you cannot yet tell, I have spent a long time in ES5.

If you have time, hop in the [postcss gitter room](https://gitter.im/postcss/postcss). We can work together to improve this pull for 4.2.

lib/at-rule.js
MoOx: Maybe because @jonathantneal is not aware of this es feature (babel use used and add support for es spread operator)

MoOx:
Looks great !
+1 for this PR in the next minor release

ai:
@lydell ES6 syntax is too new for us. I just know some stuff just today from @gaearon.

@jonathantneal wait for PR updates with more ES6 :). Welcome aboard :).

MoOx:
That might helps you http://babeljs.io/docs/learn-es2015/

lydell:
I shouldn’t review code when I’m too tired … that’s not the way I intended to sound! :/

ai:
@jonathantneal I am ready to merge it with 4.2 branch. Just wait for ES6 fixes :).

ai:
Merged

Issue #348, opened 2015-05-27
Add options for granular positioning to warn and error functions


davidtheclark:
Re: https://github.com/stylelint/stylelint/issues/133

Right now the positions reported by errors or warnings rely on the `node` option; therefore, the position is approximate but not exact, since a node might contain any number of lines, and the warning/error might not relate to the *first* line.

To increase accuracy, a few things could be done:

- Add an `index` option to the functions (alongside `node`), which would allow us to register an exact index for the item *relative to the specified node*. PostCSS would then calculate a line/column position based on the node's raw string value and this reported index.
- Add `line` and `column` options, so that whoever is registering the warning/error can dictate exactly what position gets reported.

Any other ideas?

davidtheclark:
I think it would also be useful to expose an API function like `node.lineOf(index)` that would return the line number of that index in the node.  Plugins might want to use these accurate line numbers for purposes other than errors/warnings.

ai:
Maybe manual `Warning` construction is a better way?

```js
var Warning = require('postcss/lib/warning');
var warn = new Warning('my own text');
result.messages.push(warn);
```

davidtheclark:
Yeah, we could certainly create our own warnings. But I thought other plugins using warnings might want the same precision (I know that any linter could use it, e.g. `postcss-bem-linter`.

ai:
@davidtheclark I afraid right now that current API suggestions is too complicated to use and to describe in docs.

Maybe the better solution is to store origin tokens from tokenizer step and then we will able to set `token` option. But what if some plugin changed value before our linter?

davidtheclark:
I don't know much at all about how the tokenizer could be used, so if you think that's a better way I defer to you :)

ai:
What about this API:

```js
node.error('Unknown variable $one', { word: '$one' })
```

davidtheclark:
I don't understand ... could you explain more?

Would `node.error()` be more appropriate than `node.warn()` for some reason?

Are you suggesting that we specify the word within the original CSS string? That would indeed be nice, but what if we have

```css
background: one, two , three , four, five;
```

And we want to make two complaints, about the spaces before the second and third commas -- how would passing `{ word: ' ,' }` distinguish?

ai:
@davidtheclark `{ word: ' ,' }` will highlight only first word.

davidtheclark:
[...]

PR #393, opened 2015-06-29
PoC for node.positionAt()

davidtheclark:
This PR is for discussion. It's just a proof-of-concept for a `positionAt()` method on the `Node` class, addressing https://github.com/postcss/postcss/issues/348, to help that discussion along. 

There's a good chance somebody with better knowledge of the codebase could find a better way to write this, using other existing stuff. But there's also a good chance that others have ideas about how this method should work (if it should exist). So I just thought I'd throw something out to see what people have to say.

This method accepts an index *relative to the beginning of `node.toString()`* -- which means, relative to the beginning of the `before` string, since that `before` string is included in `node.toString()`. I thought this might be appropriate because `node.toString()` is probably what a user would use to calculate the index of the character she's interested in. In the first two test-cases, for example, the user is giving the index of `:` relative to the start of that declaration's `toString()` result. In the third test-case, the user is giving the index of  the `p` in `3px`.

I'm thinking about this with a pretty specific application in mind: stylelint would want to provide exact positions if, for example, the colon had an unwanted space before it, or the `px` violated some linting rule.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -341,4 +341,19 @@ export default class {
             return value;
         }
     }
+
+    positionAt(indexRelativeToBefore) {
+        let index = indexRelativeToBefore - this.before.length;
+        let start = this.source.start;
+        let indexLine = start.line;
+        let startToIndex = this.toString().substr(this.before.length, index);
+        for ( let i = 0, l = startToIndex.length; i < l; i++ ) {
+            if (startToIndex[i] === '\n') indexLine += 1;
+        }
+        let startColumn = start.line === indexLine ? start.column : 1;
+        let lastNewlineToIndex = startToIndex
+          .slice(startToIndex.lastIndexOf('\n') + 1);
+        let indexColumn = startColumn + lastNewlineToIndex.length;
+        return { line: indexLine, column: indexColumn };
+    }
 }
--- a/test/node.js
+++ b/test/node.js
@@ -441,4 +441,27 @@ describe('Node', () => {
 
     });
 
+    describe.only('positionAt()', () => {
+        it('returns correct position when node starts mid-line', () => {
+            let css = parse('a { one: 1; }');
+            let one = css.first.first;
+            expect(one.positionAt(4).line).to.equal(1);
+            expect(one.positionAt(4).column).to.equal(8);
+        });
+
+        it('returns correct position when node.before contains newline', () => {
+            let css = parse('a {\n  one: 1;\n}');
+            let one = css.first.first;
+            expect(one.positionAt(6).line).to.equal(2);
+            expect(one.positionAt(6).column).to.equal(6);
+        });
+
+        it('returns correct position when node contains newlines', () => {
+            let css = parse('a {\n\tmargin: 1px,\n\t\t2px,\n\t\t3px;\n}');
+            let one = css.first.first;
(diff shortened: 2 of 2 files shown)


ai:
Aweome work. We can merge it tto 4.2 branch.

But let’s recheck our API again.

How did you see a this API user? How user will calculate `indexRelativeToBefore`?

davidtheclark:
I was thinking the user could get the `indexRelativeToBefore` by find the index of whatever character within `node.toString()`. So if `node.toString()` looked like this (for a rule): `\n  a {\n color: pink;\n}`. And the user wanted the position of the opening brace, they'd have to get the index of that opening brace. 

In this case it's easy (`node.toString().indexOf('{')`), but it could be trickier -- e.g. in stylelint we might wan to complain about the third comma in the selector list, so a simple `indexOf()` won't work. In stylelint we're using a search utility that will most of the time give us the index we need: https://github.com/stylelint/stylelint/blob/master/src/utils/styleSearch.js. It cycles through the string checking for matches and calling a callback for every match, ignoring CSS strings and comments.

davidtheclark:
Ultimately, I'd also want `node.warn` to use this via new option: I'd like to be able to pass in a node *and an index* to make the position more accurate.

ai:
This API seems too complicated and out of user cases for me.

I afraid that we will over-engineering API if we will go in this direction.

davidtheclark:
Which direction? 

I wasn't suggesting integrating that `styleSearch` utility into postcss, if that's what you were thinking. I was just pointing out how I was getting the index. User could get the index that they care about whatever way the user sees fit.

ai:
Direction of setting `index` to got warning position.

davidtheclark:
Hm ... Well I guess I could just make a separate module that does this. Open to suggestions about a simpler way to get accurate positions, I just can't think of anything right now.

ben-eb:
I think this use case is a little too specific to be included directly; but you can take what you have here and make it a separate module as it doesn't really depend on anything that doesn't exist already on the `Node` class.

ai:
Yeap, I am like idea of separated module too :).

davidtheclark:
How about `postcss-positionat`? Or want to leave this out of `postcss/` and I'll just make it a `davidtheclark/` module?

ai:
What do you think about `word` option in warnings? Maybe it will fix this issue too?

ai:
OK, now we have a API. And anyway we need some internal API for this API like `node.positionAt()`.

Let me just return to my city to have a clear mind. Anyway we can’t release this helper until 4.2 :).

davidtheclark:
Following the discussion in https://github.com/postcss/postcss/issues/348, I could modify this PR so that `node.warn()` accepts two new options: `word` and `index`. User can pass one or the other to make the warning's position more accurate. Both will internally use `node.positionAt()`. But where would I attach the accurate position -- somewhere that postcss-reporter, for example, could always look (cf. https://github.com/postcss/postcss-reporter/issues/3)?

ai:
No problem, I will add it after Custom Syntax finish.

I want to avoid calculating `before` in `index` and try to find better option word to promote word based position, not complicated integer.

ai:
Merged to `valac`

PR #394, opened 2015-06-29
Add postcss-short

jonathantneal:


ai:
@jonathantneal is postcss-short already a plugins pack?

jonathantneal:
No, postcss-short focuses on shorthand properties only now. A plugin pack is unnecessary.

If someone wishes to [add the :enter pseudo-class](https://github.com/jonathantneal/postcss-pseudo-class-enter) or to [extend font-weights](https://github.com/jonathantneal/postcss-font-weights), then those plugins should be added separately. Plus, users might prefer to [add :focus selector to every :hover](https://github.com/postcss/postcss-focus) instead. I do not believe a plugin pack is thematically relevant.

ai:
@jonathantneal but why you do not use plugin for every property shortcut? Like cssnano or cssnext with modularity?

PR #406, opened 2015-07-04
add postcss-alias to README

seaneking:
https://github.com/seaneking/postcss-alias


ai:
Awesome plugin :). It has few relations with [postcss-define-property](https://github.com/daleeidd/postcss-define-property), but I think they are completly for different cases.

/cc @daleeidd

PR #407, opened 2015-07-04
add postcss-normalize to README

seaneking:
Just a little one to auto-import the latest normalize.css

https://github.com/seaneking/postcss-normalize

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,7 @@ See also plugins in modular minifier [`cssnano`].
   in the console.
 * [`postcss-messages`] displays warning messages from other plugins
   right in your browser.
+* [`postcss-normalize`] automatically applies the latest normalize.css.
 * [`postcss-pxtorem`] converts pixel units to `rem`.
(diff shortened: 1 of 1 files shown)


ben-eb:
I don't think this is a good pattern to endorse. If it is, then other popular CSS files should be wrapped too; but why stop there? Why not `postcss-animate-foundation-normalize` which bundles three libs together? IMO it bloats the ecosystem unnecessarily and isn't any better/simpler for our users.

[postcss-import](https://github.com/postcss/postcss-import) already solves this use case. Our users can pick the CSS frameworks that they would like to load, which will of course differ from project to project.

seaneking:
Yup sure thing no worries, I know personally I thing of normalizing as almost akin to autoprefixing (ie: fixing browser inconsistencies vs using some particular framework). But totally understand the issue of feature-creep. Happy for this one to be rejected.

ben-eb:
Thanks for your understanding. I'll go ahead and close.

PR #411, opened 2015-07-07
Support runtime plugin injection.

ben-eb:
Do we need more tests than this one? It is passing all current tests.

### DIFF
--- a/lib/lazy-result.js
+++ b/lib/lazy-result.js
@@ -31,6 +31,8 @@ export default class LazyResult {
             }
         }
 
+        this.currentPlugin = -1;
+
         this.result = new Result(processor, root, opts);
     }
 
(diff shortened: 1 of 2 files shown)


ai:
I will accept it after lunch :)

ai:
What if somebody will add plugins before itself?

```
(css, result) => {
    result.processor.plugins = [newOne].concat(result.processor.plugins);
}
```

Because we use just a array index in `lazyResult.currentPlugin` we wil have infinity loop.

Or we should ignore it as rare case?

ben-eb:
I think we should ignore it for now. This functionality is more for supporting CodePen's use case; postcss-use only appends plugins to the current list. However if there is some valid use case that were to prepend plugins like you describe we can look at it then.

ai:
Released in 4.1.15.

PR #416, opened 2015-07-09
Add postcss-composes-shorthand plugin

rtsao:


ai:
I can’s still understand what is user case for this plugin? Does it CSS Modules specific?

rtsao:
A use case would be if you have a module that contains a base set of classes, which many components would compose. Then you could use a shorthand for composing classes from this module.

The plugin is currently designed to work with the CSS Modules composes syntax, which is implemented by (among other implementations) https://github.com/webpack/css-loader.

ai:
Maybe you should put link to your module in CSS Modules docs? I afraid that most of users will be confused of you plugin because that do not understand any context of this plugin usage.

PR #429, opened 2015-07-15
Add plugin#process shortcut. Fixes #337.

### LINKED ISSUE #337: Plugin#process shortcut
Mayb we should add `process()` shorcut to all plugins for debug purpose?

```js
var plugin = postcss.plugin('my-plugin', function () { … });

plugin.process(css).then(function (result) {
});
```

/cc @MoOx @simonsmith @ben-eb

ben-eb:


MoOx:
Nice !

PR #433, opened 2015-07-21
Add postcss-inline-image plugin

f:
Hi!

I developed an extension which embeds image files to the CSS file automatically.

Thanks! :)


ai:
Good start, but this plugin duplicates other without any benefits or different API.

I wait next plugin from you :-).

f:
You're right! I will push another one! Thanks! :)

PR #437, opened 2015-07-22
Add `postcss-svg-fallback` to list of plugins

justim:


ai:
Note that plugins had a abc order and links had a length order ;).

PR #441, opened 2015-07-23
Added postcss-color-hexa to readme

nicksheffield:


hudochenkov:
What is the point to have this plugin while plugin with the same functionality already exist and have better (in my opinion) syntax: `rgba(#bada55, .8)`? https://github.com/seaneking/postcss-hexrgba Maybe your plugin handle color transformation better or something else do better. So it would be good for PostCSS ecosystem to merge together this two plugins.

nicksheffield:
Yeah the syntax difference is why I thought mine was worthwhile. I think it makes more sense to leave the native `rgba()` the way it is, and to have a separate syntax for mixing hex and opacity.

MoOx:
I have to admit that hexa(#hex, alpha) make also more sense to me.
But anyway, we are all going to end up using color() function ;) https://github.com/postcss/postcss-color-function

nicksheffield:
Ah I hadn't seen this, and yeah it does what my plugin does and far more

PR #442, opened 2015-07-23
Add link to postcss-assets-rebase

cakeinpanic:


ben-eb:
There's no link to the plugin?

ai:
@ben-eb and order is wrong :). I wil fix it.

MoOx:
Also, it's seems to be a duplicate. postcss-url already does that (and many more).

ai:
@MoOx I am OK with high specific tools for one purpose. Sometimes it is difficult to understand how to use postcss-url in this way. And anyway you will have a some unshared code in postcss-url callbacks.

MoOx:
postcss-url is a complete solution to handle assets. It can inline and fallback to rebase + copy for example.

ai:
@MoOx but postcss-url can’t do postcss-assets-rebase behaviour without the line of code? I mean, that anyway user will be forced to write few lines for callback. And of course he/she want to share this code. This is why some specific solution is OK.

MoOx:
No need for a callback. Where did you see that? Callback is just another way to handle custom adjustment. https://github.com/postcss/postcss-url#usage

ai:
@MoOx hm, my mistake. So `postcss-assets-rebase` can be full reproduced by `postcss-url` just with configs?

@cakeinpanic what do you think?

cakeinpanic:
@moOx I've tried yo use `postcss-url` for our project, and also have done [PR](https://github.com/postcss/postcss-url/pull/32).
At that moment the only solution to rebase assets correctly was using callback. Its size was about 90% of `postcss-url`'s size. I thought your copy method worked correctly for some cases(but not for mine), so I decided to write my own plugin for my purposes.
As I can see, copy method was refactored, maybe it works properly now, but i've already developed my own at that point.

cakeinpanic:
@MoOx if you want, we can try to merge our plugins, no problem.

MoOx:
Fact: I only use url: rebase method (for imports), since my assets are handled by webpack.
That said, this plugin is bundled in cssnext, so I think some others people can be interested by this feature. Did you check this PR? https://github.com/postcss/postcss-url/pull/38
I think it's better to have one plugin to handle the same kind of stuff (in this case, rebase+copy), but if you plugin handle different uses cases (from what I see, not really), I don't have any problem with 2 plugins.
I just think one is better to handle similar things.

ai:
Yeap, I am thinking right not to remove this plugin.

tinchoz49:
hi @cakeinpanic as you said I did a rewrite of the copy method, here you can see the repo where i'm testing it: https://github.com/tinchoz49/test-postcss-copy

cakeinpanic:
I've checked your test repo, thank you, @tinchoz49 
Ok, so by now there are some differences between our plugins. 
In `postcss-assets-rebase`
* duplicated assets renaming
* assets folder structure saving. in `postcss-url` asset is copied with its folder structure. so if it has ten folders from .css, all ten folders would be created
* works ok when `from === to` (honestly, for our project it's the most common case)

In `postcss-url`
* hash
* base64 
* custom transform function
* basePath

For now I can say that our plugin fits better for postprocessing files generated by stylus(or LESS, or another preprocessor)

@ai what do you think, is it still count as duplicate?

ai:
Yeap, the problem with dublicate is not that it is equal, but in splitting forces. It is always better to put forces to one project, than support two.

But I am not against adding it to list.

PR #446, opened 2015-07-24
Introduce TypeScript definitions

jedmao:
I'm not sure how well-received this will be, but I wrote some TypeScript definitions for this project that allow you to get type information and documentation from at least [Visual Studio Editors](https://www.visualstudio.com/en-us/visual-studio-homepage-vs.aspx), which are now [free for open source developers](https://www.visualstudio.com/products/visual-studio-community-vs), even on [Linux and OS X](https://www.visualstudio.com/products/code-vs).

As an added bonus, this [restricts plugin developers to the public API](https://github.com/postcss/postcss/blob/master/docs/guidelines/plugin.md#24-use-only-the-public-postcss-api).

Here's an attached screenshot that illustrates methods available via inheritance from the `Node` class:

postcss-typescript

Here's another example that illustrates the knowledge that we're dealing with a `Declaration` here:

postcss-typescript2

### DIFF
--- a/.d.ts
+++ b/.d.ts
@@ -1,0 +1,4 @@
+declare module 'postcss' {
+	import postcss from 'd.ts/postcss';
+	export = postcss;
+}
--- a/d.ts/at-rule.d.ts
+++ b/d.ts/at-rule.d.ts
@@ -1,0 +1,67 @@
+import Container from './container';
+declare class AtRule extends Container implements AtRule.NewProps {
+    /**
+     * Represents an at-rule.
+     */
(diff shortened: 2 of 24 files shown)


ai:
@jedmao it is useful PR, but:

1. I need some info to update them on API changed without you.
2. You need to make PR for `valac` branch, because current master is old `4.1` release. `valac` is branch for 4.2 and it has API changes (also I plan to make some changes in this weekend).

jedmao:
That explains the `valac` branch. Thanks!

PR #451, opened 2015-07-27
Added postcss-crip

johnie:
Just a simple postcss plugin to create shorthand properties for Crips that are too lazy to write.


johnie:
[Thanks!](http://i.giphy.com/hGoDj6XY50Gw8.gif)

johnie:
Thanks!

PR #456, opened 2015-07-28
Add immutable-css tool to analysis section

johnotander:
Figure this will be useful here now that it's `1.0`, and closes johnotander/immutable-css/issues/13 created by @ai. Currently, this tool doesn't exactly function as a PostCSS plugin. However, it does use PostCSS, and incorporating the PostCSS plugin API is on the [roadmap](https://github.com/johnotander/immutable-css/issues/20).

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,7 @@ See also plugins in modular minifier [`cssnano`].
 * [`css2modernizr`] creates a Modernizr config file
   that requires only the tests that your CSS uses.
 * [`doiuse`] lints CSS for browser support, using data from Can I Use.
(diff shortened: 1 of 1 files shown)


ben-eb:
@johnotander Nice plugin. You could consider using the messages API and https://github.com/postcss/postcss-reporter for your logging, so that it is consistent with other plugins that lint CSS.

ai:
@johnotander yeap, using Message API will gives your power of many already written reporters. For example, to display messages in browser.

johnotander:
Thanks for the info! I will be sure to take this approach with the next upcoming version.

PR #457, opened 2015-07-28
Fix tests on Windows

### LINKED ISSUE #445: Tests fail on Windows
I have 28 tests failing on Windows, I think mostly due to file path issues. Unfortunately, it's rather hard to fix, because it's reading a large JSON file and checking the equality in some tests (e.g., [this line](https://github.com/postcss/postcss/blob/ac8c257ed5ae77ea688f2e5a20e88d743427250a/test/parse.js#L52)).

Mocha gives the following error in that case:
```
          "source": {
              "input": {
                  "css": "a {\n\tcolor: black\n}\n",
                  "safe": false,
     -            "file": "z:\\tab.css",
     -            "from": "z:\\tab.css"
     +            "file": "/tab.css",
     +            "from": "/tab.css"
              }
          },
```

Any suggestions on a strategy to fix this? I'm willing to help.

jedmao:
Fixes #445 

@ai I'm almost done here, but I'm getting a `parent` property scattered throughout each parsed CSS file that's throwing errors on the deep comparisons. Any ideas what's up here?

windows-errors

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -138,7 +138,8 @@ gulp.task('cases', () => {
         let css  = fs.readFileSync(path.join(cases, name));
         let root = postcss.parse(css, { from: '/' + name });
         let json = JSON.stringify(root, null, 4);
-        let file = path.join(cases, name.replace(/\.css$/, '.json'));
+        let windows = /^win/.test(process.platform) ? '.windows' : '';
+        let file = path.join(cases, name.replace(/\.css$/, `${windows}.json`));
         fs.writeFileSync(file, json + '\n');
     });
 });
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,0 +1,5 @@
+{
+	"compilerOptions": {
+		"target": "es6"
+	}
+}
--- a/test/cases/atrule-decls.windows.json
+++ b/test/cases/atrule-decls.windows.json
@@ -1,0 +1,270 @@
+{
+    "nodes": [
+        {
+            "type": "atrule",
+            "name": "page",
+            "source": {
+                "start": {
+                    "line": 1,
+                    "column": 1
+                },
+                "input": {
+                    "css": "@page :left {\n    background: black\n}\n\n@media(min-width: 0) {\n    background: white\n}\n\n@font-face {\n    family-name: \"A;' /**/\";\n}\n\n@viewport {\n    width: 110px;\n}\n\n@-ms-viewport {\n    width: 100px;\n}\n",
+                    "safe": false,
+                    "file": "Z:\\atrule-decls.css",
+                    "from": "Z:\\atrule-decls.css"
+                },
+                "end": {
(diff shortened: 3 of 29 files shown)


ai:
Maybe it is not a good time, because you already finished, but I have different idea :D.

Main problem is that your forget also to fix `gulp cases` task. It is very important to generate this tests automatically. When you change parser, you definatly didn’t want to change every test :).

So my idea is to write some `toJsonForTest(node)` function, which will recursively change object to some state-less state. You can see at `Node#toJSON` as example.

With this function we can fix paths for Windows. And also generate JSON back to files from gulp task, when we will change something in tests.

What do you think about it? (I afraid that I explain idea not in best way :) )

jedmao:
I see what you mean. I think I can do that.

jedmao:
@ai now I understand what `gulp cases` is doing, but I have to think that maybe these test cases shouldn't be pushed to the repo. Instead, it can be a `pre-test` script that generates them, which would pass on any OS w/o modification to the `gulp cases` task. What do you think?

ai:
@jedmao hmmm, nice idea. But for we can store pre-tested data on Travis CI?

jedmao:
We're about to find out (build pending).

ai:
What do you mean?

jedmao:
@ai see the latest commit. Travis is passing. I just deleted all the JSON files and added a `pretest` script in `package.json`. Works like a charm!

jedmao:
Also works on Windows, but you should do what @MoOx suggested and setup AppVeyor for the Windows build.

ai:
@jedmao but how Travis CI will know that it is correct JSON? `pretest` will be runned on same code, that should be tested.

jedmao:
I guess I don't understand when you would build the gulp cases then. Sorry. Not sure what to do here, because I'm not sure why the cases were designed this way.

ai:
@jedmao `cases` is a “reference testing” way. In some point we sure about parser and we make a reference copy of how it must work. Then we just test, that parser still reproduce same output as before.

jedmao:
ok got it.

jedmao:
@ai your strategy is working, but it's super slow! See?

super-slow

ai:
Hm. Strange. Current tests use same way to gerenate object for JSON, convert it to JSON and compare.

I will check code for some performance leaks tomorrow.

jedmao:
I got it to go much faster by including `<case>.windows.json` files for each case.

jedmao:
I'm looking for a better solution though.

jedmao:
There is another option. We could use [path.posix](https://nodejs.org/api/path.html#path_path_posix) for all path operations to keep things Unix-like. Is that a horrible idea?

ai:
Because, I found shorter solution, I took same ideas from this commit on put it under your name to sae your for history: https://github.com/postcss/postcss/commit/b3b49811703eea98b5a5caf8a5e374d6436578aa

PR #473, opened 2015-08-04
PostCSS plugin for Imperial and US customary units

cbas:
My first PostCSS plugin. Apologies for any mistakes. Just thought it would be fun to support Imperial units in CSS.


cbas:
Hmm ... Why is AppVeyor failing from a small README change? /cc @ai Do you have any idea?

ai:
@cbas do not worry for AppVeyor, it is for future 5.0 release :). Thanks for plugin.

PR #483, opened 2015-08-11
Add node#warn. Fixes #289.

### LINKED ISSUE #289: Add warn() method on node directly
Is there any reason you didn't offer a warn() method on node like for error() ?

ben-eb:


ai:
1. Add spaces in `{ node: this }`.
2. Change all examples to `node.warn` in `api.md` to promote right way.

ai:
Maybe we should put `result` as first argument. It is more mondatory even than message. And it shows in some way, that `warn` modify `result`

ben-eb:
Sure. :+1:

PR #488, opened 2015-08-15
Introduce TypeScript definitions

jedmao:
Based on https://github.com/postcss/postcss/pull/489

This PR introduces [TypeScript](http://www.typescriptlang.org/) definitions, of particular interest to PostCSS plugin developers using [Visual Studio](https://www.visualstudio.com/), now [free for open source developers](https://www.visualstudio.com/products/visual-studio-community-vs) and [available across platforms](https://www.visualstudio.com/products/code-vs). Using these definitions, you gain in-editor insight into the PostCSS API via [Intellisense](https://en.wikipedia.org/wiki/Intelligent_code_completion#IntelliSense) w/o the need to pull up documentation in a separate browser window.

postcss-intellisense

The definitions for this project expose only public API members. Everything else is hidden from you, so you can relax, knowing you are following the golden path.

Even though you can't, for example, touch the `Rule` class directly, you do have access to the `postcss.Rule` interface, which should give you everything you need. For example, you can create a callback function that is called with a rule instance like so:

```ts
function callback(rule: postcss.Rule) {
    rule.eachDecl(decl => {
        // do stuff...
    });
}
```

### DIFF
--- a/.yaspellerrc
+++ b/.yaspellerrc
@@ -84,6 +84,7 @@
     "transpile",
     "transpiles",
     "transpiling",
+    "tokenize",
     "tokenizer",
     "linters",
     "rebases",
--- a/d.ts/at-rule.d.ts
+++ b/d.ts/at-rule.d.ts
@@ -1,0 +1,41 @@
+import Container from './container';
+import postcss from './postcss';
+export default class AtRule extends Container implements postcss.AtRule {
+    /**
+     * Returns a string representing the node's type. Possible values are
+     * root, atrule, rule, decl or comment.
+     */
+    type: string;
+    /**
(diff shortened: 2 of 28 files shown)


jedmao:
For anyone interested, the definition files were generated from converted TypeScript code on my [valac-typescript](https://github.com/jedmao/postcss/tree/valac-typescript) branch. Tests were also converted into TypeScript and they all succeeded.

ai:
@jedmao I think we should also fix `build` script. This files should be copy to `npm` build?

ai:
@jedmao build script already support `d.ts`.

ai:
Thanks for great work :)

PR #490, opened 2015-08-15
Rename recursive `each` methods to `walk` methods

### LINKED ISSUE #487: Rename recursive `each` methods to `walk` methods.
I have no issue with the `each` method, as it's non-recursive, just like [`jQuery.each`](http://api.jquery.com/each/). My issue, instead, is with the similarly-named methods:
- `eachInside`
- `eachDecl`
- `eachAtRule`
- `eachRule`
- `eachComment`

Based on the name alone, I would expect the above methods to also be non-recursive, akin to the `each` method. Unfortunately, this is not the case and I find it to be highly confusing.

Instead, I propose naming these recursive methods to replace `each` with `walk`. This would be similar to [walking folder structures in Python](https://docs.python.org/2/library/os.html#os.walk) or [traversing a tree](https://msdn.microsoft.com/en-us/library/Aa227569(v=VS.60).aspx). I propose the following method names:

| Before | After |
|-------------|--------------|
| `each` | `each` |
| `eachInside` | `walk` |
| `eachDecl` | `walkDecls` |
| `eachAtRule` | `walkAtRules` |
| `eachRule` | `walkRules` |
| `eachComment` | `walkComments` |

This would separate the confusion of the `each` method and bring absolute clarity to the intention behind each `walk` method, knowing that `walk` **_always_** means recursion.

jedmao:
Fixes #487 

No rush, just trying to get a head start on this one.

Deprecates `each` methods in favor of similarly-named `walk` methods.

### DIFF
--- a/d.ts/container.d.ts
+++ b/d.ts/container.d.ts
@@ -25,7 +25,7 @@ export default class Container extends Node implements postcss.Container {
     /**
      * Iterates through the container's immediate children, calling the
      * callback function for each child. If you need to recursively iterate
-     * through all the container's nodes, use container.eachInside(). Unlike
+     * through all the container's nodes, use container.walk(). Unlike
      * the for {} -cycle or Array#forEach() this iterator is safe if you are
      * mutating the array of child nodes during iteration.
      * @param callback Iterator. Returning false will break iteration. Safe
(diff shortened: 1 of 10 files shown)


TrySound:
@jedmao As said @ai this feature will be realized in 6.0. It's too early for PR.

ai:
Of course, I can accept this PR to some branch. But I afraid, that merge work will be difficult.

Let's make PR later.

lydell:
Hmm, this PR seems to include two seemingly unrelated commits ...

ai:
@lydell yeap, old methods should be depreacted, but not deleted.

This twwo commits is commits from other alreaddy accepted PRs.

lydell:
As far as I remember, github (and git) only shows the commits that differ between the two branches. So if those other PRs already were accepted the two commits shouldn’t appear here, which indicates that something’s funky …

ai:
@jedmao can you rebase this PR in name of perfect git history?

jedmao:
Not sure where GitHub got confused, but a simple rebase got rid of those two commits. I've never seen that happen before.

Anyway, all is well now.

ai:
@jedmao why you left `eachDecl`, etc in `d.ts`? Maybe we should remove depreacted methods from descriptors?

jedmao:
`eachDecl` and such exists in the Container class because the d.ts files are generated files and that's coming from the methods that print the deprecation warnings.

That said, they [do not exist in the Container interface](https://github.com/postcss/postcss/blob/048d38ddf1a92e68831fcf2b313e73293c98bfce/d.ts/postcss.d.ts#L842-L889), which is why plugin developers will never see it in their editors with Intellisense. Only the interfaces are exposed to plugin developers. They only have access to public API members this way and will stay out of your kitchen so to speak.

PR #491, opened 2015-08-15
add postcss-input-style to README

seaneking:
https://github.com/seaneking/postcss-input-style

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,7 @@ See also plugins in modular minifier [`cssnano`].
 * [`postcss-focus`] adds `:focus` selector to every `:hover`.
 * [`postcss-generate-preset`] allows quick generation of rules.
   Useful for creating repetitive utilities.
+* [`postcss-input-style`] adds new pseudo-elements for cross-browser styling of inputs.
 * [`postcss-position`] adds shorthand declarations for position attributes.
 * [`postcss-property-lookup`] allows referencing property values without
   a variable.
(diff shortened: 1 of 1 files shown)


ai:
Right now Autoprefixer do this job, but without range and important. Maybe we should thing how to extend it?

ai:
At least, there is a difference between standard `::placeholder` and `::thumb` and `::track`.

seaneking:
Oh really? I had no idea autoprefixer handled this, since there's no non-prefixed standard (that I know of) for the various psuedo-selectors. Is the syntax the same for autoprefixer? 

And re: the !importants and such, not sure if that would fit in autoprefixer's purview, since they're not technically part of prefixing, just fixing cross-browser quirks.

ai:
Only `::placeholder` since it become a standard: https://github.com/postcss/autoprefixer-core/blob/master/test/cases/placeholder.out.css

We can handle `!impoortant` as flexbox props in Autoprefixer.

I think non-standard `::thumb` and `::track` can be in this plugin.

seaneking:
Alrighty let me just strip out `::placeholder` then.

seaneking:
okay that's done, should I change the name of the plugin as well? Seems a bit ambiguous now, if it just handles range inputs. Unless you know of any other inputs with non-standard pseudo selectors this could polyfill?

ai:
Select? Time and date inputs?

seaneking:
Alright, I'll add support for date/datetime. AFAIK <select> has no non-standard pseudo elements, but I'll look into it.

ai:
Cool plugin, BTW :)

seaneking:
Ha thanks. After a bit of digging it seems the only thing accessible on datetime is the text input field itself (as opposed to the actual datepicker), and that's only on webkit. So it seems a bit pointless to add that to the plugin. Might have to just stay a range input polyfill for now :-/

ai:
I think plugin name is good even if it works only with ranges. At least, it is extendable :-D

seaneking:
Yep agreed. I'll keep on the lookout for more to add ;-)

PR #493, opened 2015-08-17
Add clarity, fix English grammar in API doc

jedmao:


docs/api.md
```diff
@@ -126,7 +127,7 @@ cleaner.process(css, options);
 postcss([ cleaner(options) ]).process(css);
 ```
 
-Asynchronous plugin should return `Promise` instance.
+Asynchronous plugin should return a `Promise` instance.
```
ben-eb: > Asynchronous plugin**s**

docs/api.md
```diff
@@ -156,7 +157,7 @@ postcss.plugin('postcss-caniuse-test', function () {
 });
 ```
 
-Send data to next plugins using the [`Result#messages`] array.
+Send data to the next plugins using the [`Result#messages`] array.
```
ben-eb: Rather than "the next" can we put "other" here?

jedmao: Yep!

docs/api.md
jedmao: Done.

ben-eb:
Thanks. :+1:

PR #500, opened 2015-08-20
Fix postcss.d.ts export

jedmao:
This fixes an error when trying to `import postcss from 'postcss';`:

```
19:06 $tsc
lib/plugin.ts(2,8): error TS1192: Module ''postcss'' has no default export.
```

The previous export was the old way, this is the new (ES6) way.

### DIFF
--- a/postcss.d.ts
+++ b/postcss.d.ts
@@ -1,4 +1,4 @@
 declare module 'postcss' {
 	import postcss from 'd.ts/postcss';
-	export = postcss;
+	export default postcss;
 }


ai:
Do we need a 5.0.1 release?

jedmao:
That sounds right to me, unfortunately.

ai:
No problem, I plan a silence day between release and promotion, specially for fixes like this.

I just way until tomorrow evening for other fixes.

PR #504, opened 2015-08-21
Introduce code coverage with Isparta/Istanbul

### LINKED ISSUE #501: Code Coverage
How would you feel if I (or anyone) added code coverage to this project? This can be achieved with [isparta](https://github.com/douglasduteil/isparta) for ES6/babel.

See the [postcss-triangle package.json](https://github.com/jedmao/postcss-triangle/blob/6b428fd0f89011373858db02813bff6f554293a8/package.json#L9-L11) for how the tasks might look.

We don't need a badge on the README, per se, but we can definitely prevent code from being introduced that doesn't have tests to go along with it.

jedmao:
Fixes #501 

Note the `--statements 89.12 --branches 81.85 --functions 87.29 --lines 91.71` are all locked at the current coverage values. We can increase these numbers as more tests are written.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,12 +32,13 @@
     "babel-eslint":           "4.0.10",
     "gulp-eslint":            "1.0.0",
     "gulp-babel":             "5.2.1",
-    "gulp-mocha":             "2.1.3",
     "strip-ansi":             "3.0.0",
     "yaspeller":              "2.5.0",
     "gulp-util":              "3.0.6",
     "gulp-run":               "1.6.10",
     "fs-extra":               "0.23.1",
+    "istanbul":               "0.3.18",
+    "isparta":                "3.0.3",
     "eslint":                 "1.2.1",
     "sinon":                  "1.16.1",
     "mocha":                  "2.2.5",
@@ -46,6 +47,8 @@
     "del":                    "1.2.1"
   },
   "scripts": {
-    "test": "gulp"
+    "check-coverage": "istanbul check-coverage --statements 89.12 --branches 81.85 --functions 87.29 --lines 91.71",
+    "cover":          "isparta cover node_modules/mocha/bin/_mocha -i lib/*.es6",
+    "test":           "gulp && npm run cover && npm run check-coverage"
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ npm-debug.log
 
 build/
 /postcss.js
+
+coverage/
--- a/.npmignore
+++ b/.npmignore
@@ -12,3 +12,5 @@ appveyor.yml
 
 gulpfile.babel.js
 logo.svg
+
+coverage/
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -71,12 +71,6 @@ gulp.task('integration', ['build:lib', 'build:package'], (done) => {
     });
 });
 
-gulp.task('test', () => {
-    require('./');
-    let mocha = require('gulp-mocha');
-    return gulp.src('test/*.es6', { read: false }).pipe(mocha());
-});
-
 // Common
 
-gulp.task('default', ['lint', 'spellcheck', 'test', 'integration']);
+gulp.task('default', ['lint', 'spellcheck', 'integration']);


jedmao:
Also, if you want to see where the coverage is lacking, run `npm run cover` and then open `coverage/lcov-report/index.html` to see what's not covered.

andyjansson:
+1

package.json
```diff
@@ -46,6 +47,8 @@
     "del":                    "1.2.1"
   },
   "scripts": {
-    "test": "gulp"
+    "check-coverage": "istanbul check-coverage --statements 89.12 --branches 81.85 --functions 87.29 --lines 91.71",
+    "cover":          "isparta cover node_modules/mocha/bin/_mocha -i lib/*.es6",
+    "test":           "gulp && npm run cover && npm run check-coverage"
```
iamstarkov: please use [npm-run-all](https://www.npmjs.com/package/npm-run-all) for better developer experience on windows, take a look at this [`postpublish` script example](https://github.com/iamstarkov/generator-tiny-es-nm/blob/master/app/templates/_package.json#L16)

iamstarkov: also, if we want coverage sources and not transpiled versions, then its better to use isparta, which is already added to deps

iamstarkov: what is a different between `cover` and `check-coverage`

ai: `cover` generates coverage report, `check-coverage` raise a error is it is less, than previous

iamstarkov: if postcss will use coveralls service then it will be for free for all pull-requests

package.json
```diff
@@ -46,6 +47,8 @@
     "del":                    "1.2.1"
   },
   "scripts": {
-    "test": "gulp"
+    "check-coverage": "istanbul check-coverage --statements 89.12 --branches 81.85 --functions 87.29 --lines 91.71",
+    "cover":          "isparta cover node_modules/mocha/bin/_mocha -i lib/*.es6",
```
iamstarkov: you dont need to define path to binary in the node_modules, npm are doing it for you

jedmao: Actually, I do, because I tried your syntax first, but it doesn't work on Windows.

package.json
jedmao: @ai will have to setup coveralls. Is npm-run-all really necessary just for this one command?

ai: @iamstarkov autocomments looks not so good. Broken CI on low coverage is better :).

ai: @jedmao @iamstarkov I think extra tool is not cool without a strong reason. What will be a problem on Windows without `npm-run-all`?

ai:
@jedmao awesome work. I will accept this PR on weekend (PostCSS 5.0 release took all my free time :( )

package.json
jedmao: It's not a problem on Windows if you're using Bash, like I do, but `&&` doesn't work on the native command prompt. I think that's what he's taking about.

package.json
jedmao: This is because `_mocha` points to `node_modules/.bin/_mocha.cmd`, which doesn't work in this case.

ai: Why not `node_modules/.bin/mocha`?

package.json
jedmao: If that's a problem I can introduce npm-run-all. I thought he was more concerned about the sequential task part.

iamstarkov: > but && doesn't work on the native command prompt. I think that's what he's talking about.

exactly

package.json
iamstarkov: > Actually, I do, because I tried your syntax first, but it doesn't work on Windows.
> This is because _mocha points to node_modules/.bin/_mocha.cmd, which doesn't work in this case.

=(( didnt know about this

jedmao: @ai `node_modules/.bin/mocha` and `node_modules/.bin/_mocha` both don't work either. Only what I've specified works >_<

package.json
jedmao: Unfortunately, it looks like `npm-run-all` is [consuming CLI colors](https://github.com/mysticatea/npm-run-all/issues/8). Looking for a fix or alternatives.

jedmao: BTW – looks like `&&` is working fine in the Windows Command Prompt. I just did `dir && dir /w` and it works great. See http://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/ntcmds_shelloverview.mspx?mfr=true

jedmao: I should add that this `npm test` works flawlessly as well.

ai:
@jedmao OK, I will wait for that Windows fix from your discussion

ai:
Ping me when you will be ready

jedmao:
@ai there's nothing to fix on Windows. It works fine.

iamstarkov:
its great, sorry for bothering then

jedmao:
The Travis and AppVeyor checks below also verify there's nothing wrong on Windows :)

iamstarkov:
awesome

ai:
Broken tests do not fail CI :(

ai:
Also why you didn’t use `gulp-instanbul` to be closer to current build system?

jedmao:
I didn't use `gulp-istanbul` because I couldn't figure out how to get it working with isparta and there was no `gulp-isparta`. Looks like [you figured it out though](https://github.com/postcss/postcss/commit/106cfa2658d962d58b756ee1ecd72145b7e698fe). Nice work!

PR #511, opened 2015-08-24
Add postcss-rgb-plz to list of plugins

himynameisdave:
Adds the [postcss-rgb-plz](https://github.com/himynameisdave/postcss-rgb-plz) plugin to the list of plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,7 @@ See also [`precss`] plugins pack to add them by one line of code.
 * [`postcss-colorblind`] transforms colors using filters to simulate
   colorblindness.
 * [`postcss-hexrgba`] adds shorthand hex `rgba(hex, alpha)` method.
+* [`postcss-rgb-plz`] converts 3 or 6 digit hex values to `rgb`.
 
 ### Images and Fonts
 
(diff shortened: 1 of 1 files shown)


ai:
I can find a reason why this plgin can be useful for other developers.

himynameisdave:
okay..

That's cool man, you don't have to add it to the official list if you don't want. Regardless I'm going to be maintaining it. You never know what developers will need though, like I never would have thought [the rebeccapurple plugin](https://www.npmjs.com/package/postcss-color-rebeccapurple) would ever get > 10k downloads in a month but here we are.

ai:
@himynameisdave I am really curious, what is a point of using only `rgb()`? :) At least, it is more symbols. And less clear istead of HSL.

lydell:
I think the download count for rebeccapurple plugin is high because it is a dependency of cssnext, not because a lot of people nees it to transpile that color.

himynameisdave:
@lydell right forgot [cssnext](https://github.com/cssnext/cssnext) and how cool it is! :)

himynameisdave:
@ai first off I just wanna say you are right, and usually developers will prefer to use hex or something else anyway.

The only reason me (& my team) use rgb values is for consistency between alpha and non-alpha colors. Basically we don't want a codebase that, for example, jumps between `#fff` and `rgba( 255, 255, 255, 0.5 )` - we'd rather just have RGB colors. 

I built [`postcss-rgb-plz`](https://github.com/himynameisdave/postcss-rgb-plz) to specifically address this despite the fact that there are probably only a very small number of devs out there that will use it. I am aware of this, so as I stated above it's fine if this one doesn't make the list. I will still happily maintain it :)

ai:
I have a better solution for you. PostCSS has plugin to add alpha for hex colors
https://github.com/avanes/postcss-color-alpha

ben-eb:
I am in favour of merging this request, with the condition that it passes the plugin guidelines. Currently there are no tests written; PostCSS plugins *must* be tested.

Like you say it promotes consistency. I don't see why it shouldn't make the list.

himynameisdave:
@ai thanks but no thanks - we're writing/repairing super old legacy CSS in a lot of our work, none of which uses 8-digit hex colors. I could be wrong, but it doesn't look like it converts basic 3/6 digit hexes to rgb which is what mine does.

himynameisdave:
@ben-eb yes working on some tests right now. Thanks for seeing that side of it.

PR #515, opened 2015-08-25
Fix Error Node#_important was deprecated in parser

dmnsgn:
Test if node.important exists.


ai:
`important` and old `_important` are different properties.
First contains boolean value. Second contained exactly text from CSS.

So this commit do not fix it. I think you got warning just because if old plugins, which are still not use PostCSS 5.0.

Tests show, that this commit block any way to set boolean `important` to true.

dmnsgn:
Okay thanks I'll look into it.

PR #551, opened 2015-09-10
Add postcss-pseudo-content-insert

mikestreety:


ai:
Do you have a diffference between `postcss-pseudo-elements-content`?

TrySound:
@ai after-c, before-c, here just after and before

mikestreety:
Yeah, it's subtle, but I didn't want to have to write `-c` every time. It's more about an aid for forgetfulness.

:smile:

PR #556, opened 2015-09-12
Add postcss-copy-assets plugin

alexmchardy:


ai:
Why ypu do not use [postcss-url](https://github.com/postcss/postcss-url) for it?

@MoOx do I right, that postcss-url supports this feature?

MoOx:
I think postcss-url is doing too much things https://github.com/postcss/postcss-url/issues/43#issuecomment-133676539

alexmchardy:
I'll just add that postcss-url doesn't copy assets into a specific base build directory. It will copy assets into a directory relative to the postcss `to` option (the destination of the css file). See [postcss/postcss-url#34](https://github.com/postcss/postcss-url/issues/34).

The idea of this plugin is to copy all assets into a base build directory while keeping the part of the asset's path that makes sense. So for `url(../../images/icons/i.jpg)` the file is copied to `<base>/images/icons/i.jpg`. This means, when processing many css files, they're referenced assets can be gathered into one base build directory instead of scattered all around.

tinchoz49:
Hi people, I'm the author of [postcss-copy](https://github.com/geut/postcss-copy). I think we can avoid the plugin duplication (postcss-copy-assets is quite similar) working together. 

I took some feedback from some postcss-url issues like the [async one](https://github.com/postcss/postcss-url/issues/43#issuecomment-133676539) and [custom hash function](https://github.com/postcss/postcss-url/issues/44). The plugin works with the async API, has support for customizable hash function, templates for setting up relative paths e.g., ```assets/[path]/[name].[ext]``` or maybe   ```[path]/[name].[ext]``` or  ```[path]/[hash].[ext]```, etc and is well tested.

So there is some work on it and like I said before maybe we can unify efforts.

alexmchardy:
Hi @tinchoz49. I tried to better explain what led me to create `postcss-copy-assets` in shutterstock/postcss-copy-assets#1.

I'm in favor of a good tradeoff between reduced duplication and "do one thing and do it well".

PR #557, opened 2015-09-12
Add postcss-write-svg

jonathantneal:


ai:
Maybe it should be in Language extension?

ai:
Can you change a commit, because I am from phone and can't fix it.

ai:
Nope, category is correct :). I thought that it was Optimization :).

PR #573, opened 2015-09-21
Update short plugin pack

jonathantneal:
Many moons ago, we wanted this to be a plugin pack. Well, I finally got around to it. https://github.com/jonathantneal/postcss-short/issues/3


himynameisdave:
lol @ "many moons ago" :moon:

PR #574, opened 2015-09-21
README.md: use cssnano's homepage, instead of github repo

vitorgalvao:
This already happens to `cssnext`, so it seems appropriate.


ben-eb:
Thanks. :smile: :+1:

PR #581, opened 2015-09-25
add engine restriction since native Promise is used

leftstick:
While people are using lower version of node, `Promise is not defined` error occured

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "postcss",
   "version": "5.0.6",
   "description": "Tool for transforming styles with JS plugins",
+  "engines": {
+    "node": ">=0.12"
+  },
   "keywords": [
     "css",
     "postcss",


TrySound:
@leftstick Does it still let to install module on older node?

ai:
Node.js 0.10 is not supported. Having this polyfill is bad for client side build.

ben-eb:
@ai It's a node engine restriction. We're not polyfilling anything. :smile:

leftstick:
I don't think so. But haven't tried yet.

According to [documentation](https://docs.npmjs.com/files/package.json#engines), it should be a restriction at installing phase. In case someone are using older version of node

ai:
Oops, sorry :-)

ben-eb:
@leftstick Thanks! :+1:

PR #586, opened 2015-09-29
Detect source-maps created by mozilla/source-map fix #583

### LINKED ISSUE #583: Problem with source map detection in previous-map.es6
Hi i add autoprefixer-loader to my webpack config and get `Unsupported previous source map format` error. I traced problem to line: https://github.com/postcss/postcss/blob/master/lib/previous-map.es6#L70 and looks like it tests `mappings` when in internal sourcemap representation this prop called `_mappings`, so looks like we need to test `_mappings` too, or test mappings after `JSON.parse(JSON.stringify(...))`
Or can some1 who knows source map internal kitchen suggest some better decision?

asan:
All motivation in ticket #583

### DIFF
--- a/lib/previous-map.es6
+++ b/lib/previous-map.es6
@@ -67,7 +67,7 @@ export default class PreviousMap {
                     .fromSourceMap(prev).toString();
             } else if ( prev instanceof mozilla.SourceMapGenerator ) {
                 return prev.toString();
-            } else if ( typeof prev === 'object' && prev.mappings ) {
+            } else if ( this.looksLikeMap(prev) ) {
                 return JSON.stringify(prev);
             } else {
                 throw new Error('Unsupported previous source map format: ' +
@@ -90,4 +90,14 @@ export default class PreviousMap {
         }
     }
 
+    looksLikeMap(map) {
+        if ( typeof map !== 'object' ) return false;
+
+        if (map.mappings || map._mappings) {
+            return true;
+        } else {
+            return false;
+        }
+
+    }
 }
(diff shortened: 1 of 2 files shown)


ai:
Using private `_mapping` is look like hack. We should fix it, but maybe we have better way?

\cc @fitzgen

asan:
The other way to get `obj.mappings` is to try call `toJSON` or `JSON.stringify(JSON.parse(...))` but i`m not sure how safe is it(requires some try-catch statment imo).

lydell:
Can't an `instanceof` check be used, just like a few lines above?

asan:
@lydell in my case this sourcemap object is from sass-loader(webpack), and it uses different instance of SourceMapGenerator(from webpack-core), so we need some kind of duck typing here, like:  if it can be converted to json, and this json has field `mappings` - it sourcemap

ai:
@asan sorry for delay. I just return from my vacation.

Can you add some test cases? I will try to find a better detection and if I will not have result in next day, I accept this check. So, anyway we will ave release after a day after you adding a test. Wit for tests :).

asan:
Hm, interesting i try to simulate different instances of mozilla/source-map by requiring it in different case, it works ok on osx but fails on unix. Any ideas how to simulate sorcemaps required by different packages?

asan:
Ok i fix it using 'source-map' and  'isparta/node_modules/source-map' seems legit.
@ai any thoughts on improving sourcemap detection?

ai:
@asan thanks for tests. Sorry, I have few things on work, but in the end of the week I will make a release.

ai:
Released in 5.0.9

asan:
@ai tnx a lot!

PR #588, opened 2015-09-30
Add postcss-strip-zero-length-units to README

bighuggies:
Simple plugin to remove unnecessary units from declarations, e.g:

```css
.foo { width: 0px; }
```

becomes:

```css
.foo { width: 0; }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,6 +460,7 @@ See also plugins in modular minifier [`cssnano`].
   if it is missing.
 * [`postcss-pxtorem`] converts pixel units to `rem`.
 * [`postcss-remove-prefixes`] removes vendor prefixes.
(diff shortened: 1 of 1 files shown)


Semigradsky:
[postcss-convert-values](https://github.com/ben-eb/postcss-convert-values) ?

> This plugin reduces CSS size by converting values to use different units where possible; for example, 500ms can be represented as .5s.

TrySound:
@bighuggies With falsy options it will do the same and also will trim leading zero.

himynameisdave:
Yeah and [`postcss-convert-values`](https://www.npmjs.com/package/postcss-convert-values) is already widely used/downloaded, so this plugin seems kinda redundant. Nice work though!

bighuggies:
Ah, I totally missed that plug-in! No need for this then.

PR #591, opened 2015-10-05
Add oldie plugin pack

jonathantneal:


ai:
Awesome pack. Do not forget to write a tweet about it.

PR #603, opened 2015-10-20
New plugin

ayhankuru:


ai:
You forget to add link to plugin. Also you should mention plugin name in commit.

Wait second one (but I am going to sleep, so maybe I will accept it tomorrow in the morning).

ayhankuru:
@ai thanks fixed soon :+1:

TrySound:
@ayhankuru It's not plugin. It's wrapper of module with own parser. Won't be accepted.

ben-eb:
I am also :-1: on this.

ai:
I looked into plugin too. Maybe crass is a good tool, but it is not good way to use it.

Stringifing and reparsing may cause unexpected behavior.

It is better just to run crass in old way after PostCSS.

PR #618, opened 2015-10-29
Add postcs-shades-of-gray link

laureanoarcanio:


ai:
Can you fix commit?

Plugins order is abc (but `postcss-` prefixes is on top).
Links order is by length.

PR #629, opened 2015-11-05
Use Rollup to build postcss

Victorystick:
Have you considered using [Rollup](https://github.com/rollup/rollup)?

`postcss` would be bundled into a single file (`build/lib/postcss.js`) removing the duplicated `babel-helpers` and the `require` calls needed to load each individual CommonJS module. The ES module semantics are preserved, while optimising file size and efficiency compared to regular CommonJS modules.

The size of all files in `build/lib` is reduced from 164 kB  when compiled individually to (100 kB) using rollup.


ai:
@Victorystick many node packages (like custom parsers) requires separated files of PostCSS like:

```js
import Comment from 'postcss/lib/comment';
import Parser  from 'postcss/lib/parser';
```

As I understand, we can’t bundle all files to single file.

Victorystick:
I see. I wasn't aware of that. Feel free to close this then.

This wouldn't have been an issue if `postcss` were to include named exports for those components so that

```js
import { Parser } from 'postcss';
```
were possible. But CommonJS makes this difficult.

ben-eb:
@ai Perhaps we can revisit this for PostCSS 6?

ai:
@ben-eb I really do not like ship big bundle in npm. Sometimes developer should go inside files in npm package.

ben-eb:
OK. But you know, `import { Parser } from 'postcss';` is a much more elegant solution than using string paths.

ai:
@ben-eb yeap, it is a good point.

BTW, Parser will not be part of `postcss` in 6.0. But I understand your point.

Victorystick:
I know @Rich-Harris has argumented against using string paths as an API, since it prevents refactoring the directory structure of a project (which is arguably an implementation detail).

What we want is for developers to be able to select components at a more granular level than the entire package. ES modules makes this possible [without using string paths](http://rollupjs.org). I'd love to help out of I can.

ai:
I add separated issue for it: https://github.com/postcss/postcss/issues/630

PR #632, opened 2015-11-08
Add postcss-inline-svg plugin

TrySound:


ai:
Can you please put a dot in the end? :)

PR #640, opened 2015-11-19
Adds postcss-cachebuster plugin

glebmachine:
^_^


TrySound:
@glebmachine Imports don't have practicle use case in production, caz slow. Better use postcss-import. For caching images there are a few plugins already. So, duplicating isn't good.

ai:
Sorry, right now I rewrite docs and this merge will make a conflict. I will accept it later, when I finish rewriting.

PR #664, opened 2015-12-05
Add postcss-not-x-child plugin to the plugins list

hudochenkov:
Add [postcss-not-x-child](https://github.com/hudochenkov/postcss-not-x-child) plugin to the plugins list.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -252,6 +252,8 @@ See also plugins in modular minifier [`cssnano`].
   Useful for creating repetitive utilities.
 * [`postcss-input-style`] adds new pseudo-elements for cross-browser styling
   of inputs.
+* [`postcss-not-x-child`] adds `:not-first-child` and `:not-last-child`
+  selectors.
 * [`postcss-position`] adds shorthand declarations for position attributes.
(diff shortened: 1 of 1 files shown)


ai:
What is a reason for this plugin? WHy we can’t write `div:not(:first-child)`?

hudochenkov:
I've use a lot of `div:not(:first-child)` and `div:not(:last-child)` after [this advice](https://github.com/AllThingsSmitty/css-protips#use-not-to-applyunapply-borders-on-navigation) recently. These selectors looks a little messy. So I propose `:not-first-child` and `:not-last-child` shortcuts as better looking selectors and with improved readability.

hudochenkov:
Frankly I don't know if someone else will need these plugin. I was just thinking that every PostCSS plugin should be listed in that list.

ai:
Nope, I I think to that it is one-user plugin :). Maybe you even should not use, because it do not brings you any benefits, just confuse other developer.

I try avoid dangerous plugins from this list ;).

hudochenkov:
No problem. Thank you for constantly keeping PostCSS ecosystem and community on the right path :)

P. S. It was my first plugin and npm package ever, so it was a great experience for future development :)

PR #665, opened 2015-12-07
Replace postcss-svg with postcss-inline-svg

TrySound:
postcss-svg plugin does not correspond to guidelines:
- there are no tests,
- it's synchronous.


ai:
No tests and sync is a big problems, I agree. /cc @Pavliko

PR #666, opened 2015-12-08
add rollup-plugin-postcss

egoist:


egoist:
well, `Rollup` is considered typo...

TrySound:
@egoist You shouldn't use sync postcss api. It's only for debug. You should use promises which could be returned in transform method.

egoist:
gotcha.

egoist:
Now using the Promise.

TrySound:
@egoist postcss own return promise. You don't need to wrap it in new one.

egoist:
@TrySound 😂 sorry for that, fixed.

TrySound:
@egoist Can you show some real world usage example?

egoist:
@TrySound sure, later I'll add some to that repo.

ai:
@egoist sorry, I will compress Rollup link to one line (like Grunt or Brunch) because it is not so popular yet.

PR #669, opened 2015-12-11
Add postcss-sorting to plugins list

hudochenkov:
[postcss-sorting](https://github.com/hudochenkov/postcss-sorting) is a plugin to sort rules content with specified order.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -292,6 +292,7 @@ See also plugins in modular minifier [`cssnano`].
 * [`postcss-pxtorem`] converts pixel units to `rem`.
 * [`postcss-select`] select rules based off a selector list.
 * [`postcss-shorthand-expand`] expands shorthand properties.
+* [`postcss-sorting`] sort rules content with specified order.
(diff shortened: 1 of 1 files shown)


TrySound:
Seems like duplicating [css-property-sorter](https://github.com/Siilwyn/css-property-sorter). Also it does not follow guidelines. Plugin should use async methods where it's possible.

hudochenkov:
@TrySound it's absolutely not duplicating. css-property-sorter sort only properties. My plugin sort also at-rules, nested rules, variables; can group nodes. Also not breaks nesting (css-property-sorter does). css-property-sorter sorts only _properties_ (plugin name and readme very clear about this) in not nested rules, my plugin sort _all_ rule's content in nested (or not) rules.

About async I agree. I don't know how to replace sync with async yet.

TrySound:
@hudochenkov U can see in css-property-sorter. Are you sure that rules sorting not breaks rendered styles?

hudochenkov:
@TrySound as I tested on complex stylesheets it's not break styles. Style rendering done by PostCSS itself. Obviously edge cases may exists, no one can write software without bugs :)

PR #671, opened 2015-12-12
Fix warning toString method with specified index

TrySound:


ai:
Nice fix. But also we need to support `word` option too. I will make it tomorrow.

TrySound:
@ai Maybe just pass this as an options?

TrySound:
Done!

PR #681, opened 2015-12-22
add postcss-hash-classname plugin

ctxhou:


ai:
Could you fix a sorting? Plugins are sorted by ABC (`postcss-` prefix is not counting) and links are sorted by length.

ctxhou:
ok, I fix it now.

PR #707, opened 2016-01-21
New svg logo with left padding

soul-wish:
This new version of logo can be used in the places where we should have some additional space between logo image and description/readme text. For example: `README.md` file of [PostCSS Plugin Boilerplate](https://github.com/postcss/postcss-plugin-boilerplate) repo.


ai:
Thanks :) I will update all my repos

PR #709, opened 2016-01-21
Update api.md with correct method to add an atrule.

studiosciences:
When using append/prepend to add an atrule, name: should not include @.


ben-eb:
Thanks! :+1:

studiosciences:
My pleasure!

PR #713, opened 2016-01-24
Use package.json version in the processor

igncp:
Hi,

In this way there is only one source for the version, and it doesn't have to be changed manually for each bump. Hope it helps.

Cheers

### DIFF
--- a/lib/processor.es6
+++ b/lib/processor.es6
@@ -1,8 +1,9 @@
 import LazyResult  from './lazy-result';
+import { version } from '../package.json';
 
 export default class Processor {
 
-    version = '5.0.14';
+    version = version;
 
     constructor(plugins = []) {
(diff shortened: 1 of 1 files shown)


ben-eb:
We do this because otherwise you need to configure a JSON loader for webpack. :smile: 

https://github.com/postcss/postcss/commit/eb3f7dcdbcd4afdbee1c95af9310234837ff3ce2

ai:
Here is origin issue of that commit https://github.com/postcss/postcss-js/issues/3

igncp:
Right! You had it like this before... :sweat_smile:
Adding postcss to a webpack bundle might be a rare case (normally it is used as a loader) but if you want to support that case without explicit loaders, I guess that another approach would be to write a script to bump the version automatically... anyway it is not a big deal, great project!

PR #716, opened 2016-01-27
Only include the library directory to reduce npm package size

Siilwyn:
Decrease size of the postcss npm package by 57% (295 kB) by only including the library directory. Instead of ignoring a bunch files I included the library directory in the package.json.


ai:
Em, what about docs and type definitions? They are very useful for developers

Siilwyn:
I don't think much people go into the *node_modules* directory, open *postcss* and then read the docs offline. I could be wrong so if anybody is against removing the docs from the packaged version of postcss, speak up.
Though for anybody wanting to read the docs offline cloning this repository would a good option. Also the readme file is included which contains some examples of how to use postcss.

ben-eb:
I read docs offline. It's much easier than *opening a browser*, searching for the *repository*, and using the *web UI* to find the relevant file. I can do it all from `node_modules` using Atom.

TrySound:
@Siilwyn That's why readme.md can't be ignored in npm.

Siilwyn:
@TrySound exactly, I think the readme.md is enough for most people.

TrySound:
@Siilwyn readme is docs. postcss has more docs than only readme. So you should include it.

Siilwyn:
In hindsight, and after all of your replies (which I thank you for) I see why this isn't such a good idea. I still don't fancy the idea of including the TypeScript source but I understand why it is included.

:outbox_tray:

TrySound:
We too don't like typescript, but it is.

ai:
Also I am big fan of airplane development (because you have Internet for procastinate). It is one of the main reasons for me for offline docs :D.

PR #722, opened 2016-02-07
Move condition up the loop

igncp:
Hi,

Just spotted this `if` inside the loop, which could be moved outside and increase slightly the performance. Hope it helps.

Cheers


ai:
Nice, catch :)

PR #726, opened 2016-02-09
Remove Broken Link to Article in Readme

snide:
Looks like the site julian.io is no longer available. No idea if this is permanent, but figured I'd send in a PR in case no one noticed.

http://julian.io/some-things-you-may-think-about-postcss-and-you-might-be-wrong/

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,6 @@ PostCSS can transform styles in any syntax, not just CSS.
 
 ## Articles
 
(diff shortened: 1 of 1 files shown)


TrySound:
@snide Everything works fine. Check it again please.

snide:
Strange. Connection being refused. Might be a firewall issue at work. Works fine on my phone. Thanks for checking.

ben-eb:
Good to remember this link, too. http://isup.me/julian.io :smile:

PR #728, opened 2016-02-10
Track newlines within strings; fixes #727

### LINKED ISSUE #727: CSS syntax error reporting wrong line/column position
@Arcanemagnus originally filed this as an issue with stylelint (https://github.com/stylelint/stylelint/issues/718):

With this CSS file:

```css
.s["] {
}
/* "" */
```

There is a CSS syntax error that reports the column of the unclosed quote on line 1 as being 18, which is far past the end of the line. It seems like it is counting all characters between the first quote and the next one as part of the line? When the comment with the `"` is removed the proper value is returned. If that comment is brought to the same line as the initial `"` the issue disappears as well.

Based on a report initially filed here: https://github.com/AtomLinter/linter-stylelint/issues/79.
image

This is different from #725 because the error is thrown during the tokenize step (so it's not using `node.positionBy()`).

I think that the syntax error is causing weird problems with text editor plugins, because it claims that things are happening in positions that do not exist.

Do you think there's a way to make these positions more accurate? If not, maybe text editor plugins just need to do something different when receiving CSS syntax errors from PostCSS?

davidtheclark:
I believe that the issue causing #727 is that the tokenizer was not registering newlines within strings. So whatever came after the string was thought to be on the same line as the string's opening --- which in this instance was not the case. 

I copied some logic from the tokenization of comments and verified that it works with a test. I also verified that the particular issue in #727 (the syntax error having a weird position) is fixed by this.

### DIFF
--- a/lib/tokenize.es6
+++ b/lib/tokenize.es6
@@ -143,11 +143,26 @@ export default function tokenize(input) {
                 }
             } while ( escaped );
 
+            content = css.slice(pos, next + 1);
+            lines   = content.split('\n');
+            last    = lines.length - 1;
+
+            if ( last > 0 ) {
+                nextLine   = line + last;
+                nextOffset = next - lines[last].length;
+            } else {
+                nextLine   = line;
+                nextOffset = offset;
+            }
+
             tokens.push(['string', css.slice(pos, next + 1),
                 line, pos  - offset,
-                line, next - offset
+                nextLine, next - nextOffset
             ]);
-            pos = next;
+
+            offset = nextOffset;
+            line   = nextLine;
+            pos    = next;
             break;
 
         case AT:
--- a/test/tokenize.es6
+++ b/test/tokenize.es6
@@ -98,6 +98,13 @@ describe('tokenize', () => {
         test('"\\\\"', [ ['string', '"\\\\"', 1, 1, 1, 4] ]);
     });
 
+    it('changes lines in strings', () => {
+        test('"\n\n""\n\n"', [
+            ['string', '"\n\n"', 1, 1, 3, 1],
+            ['string', '"\n\n"', 3, 2, 5, 1]
+        ]);
+    });
+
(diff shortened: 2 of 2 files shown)


ai:
Thanks for welll documented issue and PR. Sorry, I spend all my opens ource hours today and any tokenizer changes should be double checked (also they should be backported to `postcss-safe-parser` and `postcss-scss`).

So I will look in this issue in Thuesday/Friday. Feel free to ping me on next week if I will forget.

I think this issue can wait, yeap?

davidtheclark:
@ai: Whenever you're able! Thanks for your hard work.

ai:
I looked to PR. You are a hero, because you are not afraid tokenizer :D.

Main problem, that `\n` is illegal in CSS Syntax spec. Maybe we should throw error just about `\n` in string https://www.w3.org/TR/css-syntax-3/#consume-a-string-token

davidtheclark:
Ah, you're right. For some reason I thought it was allowed but just didn't show up when rendered. But the spec is clear: "Newline: This is a parse error."

Well, I think I agree with you then that throwing a syntax error is the way to go.

ai:
I compare perfomance, with extra lines code doesn’t slow down parser.

I think we should parse multiline strings. It is same case as nested rules — PostCSS is not a linter (itself :) ). So it could parse more than standard CSS if it is not a problem.

davidtheclark:
Eh, ok. Either way is good by me!

ai:
Changes was backported to `postcss-scss`, `postcss-safe-parser` and `sugarss` with your name in commit author field.

ai:
Tomorrow I will look into SCSS comment issue and will release 5.0.16.

davidtheclark:
Great. Thanks, @ai.

davidtheclark:
This raises an idea for a good stylelint rule: `string-no-newline`

PR #746, opened 2016-02-25
Update TypeScript support

jedmao:
I recompiled [my typescript branch](https://github.com/jedmao/postcss/tree/typescript) with the newly-released [TypeScript 1.8.2](https://github.com/Microsoft/TypeScript/releases/tag/v1.8.2) and fixed the errors that popped up.

Also, TypeScript now supports a new `typings` field in the package.json file, which points to the type definitions for postcss. This is now the golden path and I like it very much compared to what we had before!

### DIFF
--- a/d.ts/at-rule.d.ts
+++ b/d.ts/at-rule.d.ts
@@ -32,8 +32,10 @@ export default class AtRule extends Container implements postcss.AtRule {
      */
     clone(overrides?: Object): AtRule;
     toJSON(): postcss.JsonAtRule;
-    append(...children: any[]): Container;
-    prepend(...children: any[]): Container;
+    append(...children: any[]): this;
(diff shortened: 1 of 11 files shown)


TrySound:
@jedmao Is there a way to make it a bit more compact?

jedmao:
@TrySound make what more compact?

ai:
@jedmao why you need to add `typing` field to `package.json` dynamically? Maybe we should put it to `packge.json` file directly?

ai:
Do you need a release for this?

TrySound:
Definitions.

jedmao:
@ai yes this will need a release. And yes, we can put it in the `package.json` file directly. I'll submit a PR for that.

PR #747, opened 2016-02-25
Move typings to package.json

jedmao:
In response to https://github.com/postcss/postcss/pull/746#issuecomment-188988230


ai:
I will release version tomorrow

PR #749, opened 2016-02-26
Add the postcss-inline-rtl plugin

jakob101:
Added the postcss-inline-rtl plugin.


TrySound:
@jakob101 Please squash and rename commit like `Add the postcss-inline-rtl plugin`.

PR #758, opened 2016-03-07
Enabling source maps to handle odd urls

### LINKED ISSUE #757: Enabling absolute urls in source maps
Hi,

I'm in the process of trying to implement a PR for postcss that would enable absolute urls (including protocols, eg: file, http, or https). In effect, given the following options

```js
from: 'file:///dir/a.css',
to: 'http://example.com/a.css'
map: {
  inline: false, 
  // note: this option is described below
  preservePaths: true
}
```

I'd like for the source map to correctly reference both the source and the target file. Unfortunately, postcss is currently implemented such that `from` and `to` are assumed to be file directories that a relative path can be inferred from. Suffice to say, the resulting output is fairly broken.

I've tried adding the following to the `relative(file) {...}` in `map-generator.es6`:

```js
if (this.mapOpts.preservePaths) {
  return file;
}
```

The intention being to enable the `map.preservePaths` flag to prevent the majority of the behaviour in the `relative(file) {...}` method, but unfortunately this hasn't had the immediate result that I hoped for. 

From logs that I've been outputting, the incoming values to `relative(file) {...}` is called with `file` equal to `/path/to/cwd/file:/dir/a.css`. 

My _assumption_ is that something higher up in the codebase would be manipulating the `from` argument. But - having taken a bit of a dig through the source code - I've been unable to understand why. Is anyone able to shed any light on how I could approach the problem of preventing complete urls from being manipulated during source map generation?

markfinger:
This adds two options to the `map` options available:
- `preservePaths` which forces early escapes in the `relative(...)` function. This enables cross-protocol urls to function as expected.
- `sourcePath` an optional override for the relative path generated during processing 

This would fix #757

### DIFF
--- a/docs/source-maps.md
+++ b/docs/source-maps.md
@@ -65,4 +65,8 @@ option as an object with the following parameters:
 
   If you have set `inline: true`, annotation cannot be disabled.
 
+* `from` string: by default, PostCSS will set the `sources` property of the map
+  to the value of the `from` option. If you want to override this behaviour, you
+  can use `map.from` to explicitly set the source map's `sources` property.
+
 [source maps]: http://www.html5rocks.com/en/tutorials/developertools/sourcemaps/
--- a/lib/map-generator.es6
+++ b/lib/map-generator.es6
@@ -165,6 +165,10 @@ export default class {
     }
 
     relative(file) {
+        if ( file.match(/https?:\/\/|file:\/\//) ) {
+            return file;
+        }
+
         let from = this.opts.to ? path.dirname(this.opts.to) : '.';
 
         if ( typeof this.mapOpts.annotation === 'string' ) {
@@ -181,6 +185,9 @@ export default class {
     }
 
     sourcePath(node) {
+        if ( this.mapOpts.from ) {
+            return this.mapOpts.from;
+        }
         return this.relative(node.source.input.from);
     }
 
--- a/test/map.es6
+++ b/test/map.es6
@@ -508,4 +508,23 @@ describe('source maps', () => {
         expect(result.css).to.include('a {\r\n}\r\n/*# sourceMappingURL=');
     });
 
+    it('preserves absolute urls in `to`', () => {
+        let result = postcss().process('a{}', {
+            from: '/dir/to/a.css',
+            to:   'http://example.com/a.css',
+            map:  { inline: false }
+        });
+        expect(result.map.toJSON().file).to.eql('http://example.com/a.css');
+    });
+
+    it('`map.from` should override the source map sources', () => {
+        let result = postcss().process('a{}', {
+            map: {
+                inline: false,
+                from:   'file:///dir/a.css'
+            }
+        });
+        expect(result.map.toJSON().sources).to.eql(['file:///dir/a.css']);
+    });
+
 });


markfinger:
I've contributed a test case which coverts my use-case, specifically, exposing a file path to postcss, but ensuring that the generated source map can reference assets across multiple protocols.

Also, I'm not particularly attached to the names (or explicit behaviours) of the changes that I've made. So if there are recommendations which are more in-line with the "postcss way", I'm happy to make them. Though I'd rather avoid a bikeshed, if possible.

ai:
Having a special `option` is not a best way. Maybe we should change file paths logic on having `https?://` in `to`?

markfinger:
@ai I've pushed a change that replaced `preservePaths` with a check for protocols.

ai:
@markfinger awesome. Could I ask you to join all commits to one? Just `git reset HEAD^` all commits, make one commit and `git push --force` it again.

markfinger:
@ai I've flattened the commits. The CI seems to be spitting out some weird failures though

ai:
Yeap. Travis has weird issue. But it is not relevant for this PR.

Awesome work. I am now in the middle of SugarSS development. Can I merge this PS after hew hours? Everything fine, I just want to focused on one task :).

markfinger:
Cool. Thanks, @ai. 

Do you mind updating this when the changes are published to npm?

ai:
@markfinger yeap, I will merge it after few hours and publish on npm in 5.0.20 (you are lucky to have cool version number for your fix :) ).

I will write here when I make a release.

ai:
@markfinger hm, why we need `sourcePath`? Could we omit this option too?

markfinger:
The use-case for `sourcePath` is to override the file that inserted into the source map's `sources` array.

As (I believe) postcss tries to read in previous source maps, the `from` should be a file path (feel free to correct me if I'm wrong). However, we can't dump an absolute fs path into the source map as browsers tend to bail at anything that's not a relative url, absolute, or a complete url (with protocol).

Basically, I want to provide the `from` as an absolute file path, the `to` as the url that the asset will be served from, and the `sourcePath` as the url that the text content is reachable from in a browser.

markfinger:
As before, if there's an alternative that's more inline with the "postcss way", I'm happy to make any changes.

ai:
If `from` is a file path, why we need to hack `sourcePath`?

markfinger:
Browsers don't interpret OS file paths as urls, so when you're trying to navigate from devtools to the file contained in `source`, it tends to break in varying ways across browsers.

For example, in one of my projects, Firefox is currently breaking heavily on mapping the source map to the generated file. The one hack that _sort of_ gets around it is to use a `file://...` protocol url so that navigation is a bit better.

markfinger:
Correction, browsers **do** interpret OS file paths as urls.

ai:
Maybe we should add `file://` support for `from`?

markfinger:
Yeah, I suppose you could strip the `file://` from the start. The only problem I can see with that is you would end having either two code paths scattered throughout `postcss` or you'd need to store it during init. Either way, I think you'll still end up in much the same situation as the `sourcePath` override.

markfinger:
My need for that option is because `postcss` overloads the intention of `from`: It's both the input filename and the source url for the source map.

In most other build tools, there are either distinct phases (read in / emit to code / emit to source) or there are multiple options provided for this use case. For example: babel-generator's [sourceMapTarget](https://github.com/babel/babel/blob/master/packages/babel-generator/README.md#options)

markfinger:
Another example in babel: https://babeljs.io/docs/usage/options/. There are distinct options for `filename` and `sourceMapTarget`.

ai:
In this case we can add `map.from` option to override main `from` option. I think this option name is more clear, than `sourcePath`.

markfinger:
Sure, sounds good. I'll make the change

ai:
Also you need to add docs for this option too :).

markfinger:
Ok, I've made the changes and added some docs for `map.from`

ai:
Awesome! Sorry, I was burned off in SugarSS release. I will think about this PR tomorrow. Maybe we need a minor release (not patch) because we add new option.

Are you OK with some delay for this issue?

markfinger:
Yeah, sure

markfinger:
Alternatively, I can split this up into 2 PRs, if it makes it easier for you

ai:
2 PRs looks like good idea, because 5.1 release could take some time.

markfinger:
Closing in favour of #759 and #760

PR #759, opened 2016-03-08
Absolute urls should be handled by source maps

markfinger:
re: postcss#757
re: postcss#758

### DIFF
--- a/lib/map-generator.es6
+++ b/lib/map-generator.es6
@@ -165,6 +165,10 @@ export default class {
     }
 
     relative(file) {
+        if ( /^(?:https?|file):\/\//.test(file) ) {
+            return file;
+        }
+
         let from = this.opts.to ? path.dirname(this.opts.to) : '.';
 
         if ( typeof this.mapOpts.annotation === 'string' ) {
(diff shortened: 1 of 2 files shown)


lib/map-generator.es6
```diff
@@ -165,6 +165,10 @@ export default class {
     }
 
     relative(file) {
+        if ( file.match(/https?:\/\/|file:\/\//) ) {
```
lydell: `/^(?:https?|file):\/\//.test(file)`

lydell: The most important point in my above comment:

Be sure to use `^` to only match at the beginning.

markfinger: Yeah, I noticed the caret. Personally, I don't think it'd be too much of an issue as the colon after the protocol wont match any realistic OS path -even OSX tries not to support it anymore as a non-delimiter character.

That being said, the caret makes it more explicit and probably offers some minor perf improvements, so I don't mind rebasing the pull again.

markfinger:
Ok. It's been updated with @lydell's request.

ai:
Merged to 5.1 branch

PR #760, opened 2016-03-08
Add `map.from` option

markfinger:
This allows overrides of a source map's `sources` property.

re: postcss#758 

@ai this should be added to the next minor release


ai:
Merged manually to 5.1 branch

PR #761, opened 2016-03-09
Add <no source> mapping

TrySound:


TrySound:
Should I add the same position in original?

ai:
> Should I add the same position in original?

Hm. Tricky question. Can you should browser output without a `origin` position?

ai:
Merged to 5.1 branch

PR #762, opened 2016-03-09
Fix missing type causing TypeScript compiler error with noImplicitAny.

jjrv:
TypeScript compiler run with `tsc --noImplicitAny` prints:

`node_modules/postcss/d.ts/postcss.d.ts(26,19): error TS7006: Parameter 'css' implicitly has an 'any' type.`

Adding the `any` type explicitly fixes the error.

### DIFF
--- a/d.ts/postcss.d.ts
+++ b/d.ts/postcss.d.ts
@@ -23,7 +23,7 @@ declare module postcss {
     interface Plugin<T> extends Transformer {
         (opts?: T): Transformer;
         postcss: Transformer;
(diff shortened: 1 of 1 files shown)


ai:
Looks fine. @jedmao can you check this PR as TS master of this project? :)

jedmao:
This might be OK for a short-term fix, but there are actually hundreds of implicit anys in the typescript branch that need to be declared. I'll need some time to do this for the long-term fix.

PR #777, opened 2016-03-23
accept function for previous map option

montmanu:


ai:
Want is your user case for this feature?

montmanu:
To dynamically generate the `prev` option's path.. useful when the src file does not already include an annotation and the map is located in a different directory than the src file

ai:
Why you can do something like `{ map: { prev: func() } }`?

montmanu:
I don't think that option currently accepts a function.. that's why I opened this PR and wrote the patch.

ai:
Why you can't execute function on PostCSS#process call? Why you need to call it later inside PostCSS?

montmanu:
I'm not interacting with postcss directly.. I'm interacting with it via a plugin.

ai:
What plugin?

montmanu:
Sorry.. should have mentioned that: grunt-postcss

ai:
I merge it manually in 5.1 branch.

montmanu:
ok, thanks!

Issue #339, opened 2015-05-19
decl.value should be forced as string


MoOx:
Some plugin might add things like integer as value in the AST.
When setting declaration value, we should force the value to be a String.

Related: this https://github.com/postcss/postcss-calc/pull/10 (where you would find more material about why and who to handle this)

This will prevent bugs in all plugins that do `decl.value.indexOf(...`

ben-eb:
Agreed. Would save plugin developers from having to coerce it themselves (such as here https://github.com/ben-eb/postcss-minify-font-weight/blob/master/index.js#L5-L9)

evilebottnawi:
@ai bump

ai:
Ouh. I didn’t like a virtual properties by getter/setter, but I think we have no other solution.

I will fix it after CSSConf.

@evilebottnawi @MoOx can you help me, please? Can you invastigate the best way to define virtual property, that will looks like common JS property. For example, it should be disableable in `console.log(decl)`.

ben-eb:
@ai Can we not do something like this:

```js
export default class Declaration extends Node {
    // ...
    set value (value) {
        this.raw._value = value;
    }

    get value () {
        return this.raw._value;
    }
}
```

ai:
@ben-eb we can, but `value` will not be displayed in `console.log` and many users don’t read docs and will use some strange things like `decl.raw._value`.

ben-eb:
Does it matter if they set that indirectly? As long as the AST returns a string type?

ai:
They will use raw property and will get same type issue ;-).

Any way, our API should be simple to developers. So it should be inspectable. I think `defineProperty` had options to make it inspectable.

ben-eb:
Ah, the joys of no private properties in JS. :smile:

ai:
Some extra property with long name is OK. Question is only to see `node.value` in inspect.

ai:
We can add `Node#setContext` and set `Result` to `node.context.result` to use it in `node.warn`. But does it is a good way?

Especially in current love on pure function and FP.

Or maybe we can use this API:

```
node.warn(result, 'text');
```

What guys do you think?

gucong3000:
fixed
https://github.com/postcss/postcss/pull/784

gucong3000:
```CSS
#qunit-fixture {
    position: absolute;
    top: -10000px;
    left: -10000px;
    width: 1000px;
    height: 1000px;
}
```

```JS
function cssPipe(stream) {
[...]

PR #784, opened 2016-04-08
decl.value should be forced as string

gucong3000:
fixed
https://github.com/postcss/postcss/issues/339

### DIFF
--- a/lib/container.es6
+++ b/lib/container.es6
@@ -281,6 +281,9 @@ export default class Container extends Node {
             } else if ( nodes.prop ) {
                 if ( typeof nodes.value === 'undefined' ) {
                     throw new Error('Value field is missed in node creation');
+                } else if ( typeof nodes.value !== 'string' ) {
+                    warnOnce('decl#value should be forced as string');
+                    nodes.value = String(nodes.value);
                 }
                 nodes = [new Declaration(nodes)];
             } else if ( nodes.selector ) {
--- a/test/container.es6
+++ b/test/container.es6
@@ -802,7 +802,8 @@ describe('Container', () => {
         it('decl.value should be forced as string', () => {
             let rule = parse('a { a: 1; b: 2 }').first;
             rule.append({ prop: 'c', value: 3 });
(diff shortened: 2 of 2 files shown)


ai:
User can still set other type by `node.value = 1`. I will change check in 5.1 release.

gucong3000:
There are two plugins throw exceptions, such as cssnext, cssgrace, and maybe more.
My changes can fix this problem.

test/container.es6
```diff
@@ -799,6 +799,12 @@ describe('Container', () => {
             expect(rule.toString()).to.eql('a { a: 1; b: 2;\n c: 3 }');
         });
 
+        it('decl.value should be forced as string', () => {
+            let rule = parse('a { a: 1; b: 2 }').first;
+            rule.append({ prop: 'c', value: 3 });
+            expect(rule.toString()).to.eql('a { a: 1; b: 2; c: 3 }');
```
ben-eb: This test doesn't do anything, because in the final stringify step, PostCSS will convert everything to a string. You need to check that each `decl` in the rule has a value of String type.

lib/container.es6
```diff
@@ -281,6 +281,9 @@ export default class Container extends Node {
             } else if ( nodes.prop ) {
                 if ( typeof nodes.value === 'undefined' ) {
                     throw new Error('Value field is missed in node creation');
+                } else if ( typeof nodes.value !== 'string' ) {
+                    warnOnce('decl#value should be forced as string');
```
ben-eb: I don't understand why we need to print a warning here.

ai:
Yeap, this PR is useful. I will accept it to 5.1 branch later (I have uncommit changes there right now).

lib/container.es6
gucong3000: I'm not sure we need a warning.

test/container.es6
gucong3000: @ben-ed  I modified the code as your wish.

ai:
I manually get this PR to `zagan` branch

PR #785, opened 2016-04-10
Added postcss-font-weight-names to plugin list

dan-gamble:


TrySound:
@dan-gamble Seems like duplication of https://github.com/jonathantneal/postcss-font-weights

dan-gamble:
Whoops, looks like it to be fair. Oh well, also noticed mine doesn't handle `font`!

PR #798, opened 2016-05-01
Reverse token adding in spacesFromEnd

davidtheclark:
I ran into a strange bug while trying to fix this: https://github.com/stylelint/stylelint/issues/1148

Turns out the rule's `raws.between` seemed to have its tokens reversed. From the source code 

```css
a /*x*/
{}
```` 

I'd expect the rule's `raws.between` to be `" /*x*/\n"`. But instead I was getting `"\n/*x*/ "`.

I was able to reproduce this with a test in PostCSS. And I found what I think may have been a logic flaw in the `spacesFromEnd` function that is used to create `raws.between`. I adjusted that logic to get my test to pass, and all existing tests continue to pass. (The build failure seems to be related to code coverage reporting.)

### DIFF
--- a/lib/parser.es6
+++ b/lib/parser.es6
@@ -365,11 +365,12 @@ export default class Parser {
     }
 
     spacesFromEnd(tokens) {
-        let next;
+        let lastTokenType;
         let spaces = '';
         while ( tokens.length ) {
-            next = tokens[tokens.length - 1][0];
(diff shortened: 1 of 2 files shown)


ai:
I added test case to `postcss-parser-tests` https://github.com/postcss/postcss-parser-tests/commit/80f84dcd696ef7666e159aca74902176f504a9e2

ai:
Released in 5.0.20

PR #805, opened 2016-05-17
Added postcss-less-parser plugin

matthew-dean:
This uses Less.js as a custom parser, so it correctly interprets all Less syntax, unlike the existing postcss-less plugin. It evaluates the LESS and then converts individual LESS nodes to PostCSS nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,8 @@ you can write a parser and/or stringifier to extend PostCSS.
   *(but does not compile SCSS to CSS)*.
 * [`postcss-less`] allows you to work with Less
   *(but does not compile LESS to CSS)*.
+* [`postcss-less-parser`] allows you to work with Less
+  *(and DOES compile LESS to CSS using true Less.js evaluation)*.
 * [`postcss-js`] allows you to write styles in JS or transform
   React Inline Styles, Radium or JSS.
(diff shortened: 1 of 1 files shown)


TrySound:
Parser shouldn't evaluate anything. It just generates AST. So the name is not correct.

matthew-dean:
Well, it's a custom parser. The fact that it evaluates what you are parsing in order to create usable nodes is a necessity of using Less correctly. Technically, it would have been just called "postcss-less" but someone registered that name on NPM before I did. If you have a better suggestion, let me know.

TrySound:
@matthew-dean postcss-less is only parser. So it ideologically couldn't be named like this. It's something like converter, I think. Or adapter. Evaluator?:)

ai:
Yeap, name definitely not clear. What do you think of it?

matthew-dean:
postcss-less-engine?

ai:
:+1: I like it

matthew-dean:
Okay, I'll update the repo.

ai:
Send new PR when you finish it

matthew-dean:
Before I finalize it, just want to run by two more possible names: `postcss-less-compiler` / `postcss-less-render`.

Either of those better than `postcss-less-engine` in your opinion?

ai:
`engine` is better :)

PR #806, opened 2016-05-17
Added postcss-esplit plugin

VitaliyR:
Hello. I've created and polished [postcss-esplit](http://github.com/vitaliyr/postcss-esplit) plugin - and decided to add it to readme. 

The idea inside it that we were used in our project blesscss tool which just doing the same job. But it doesn't have support for source maps. Also we though to move our project to use postcss so I decided to wrote a plugin for postcss which would do just the same thing blesscss doing but with support of sourcemaps.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ If you have any new ideas, [PostCSS plugin development] is really easy.
 * [`lost`] is a feature-rich `calc()` grid system.
 * [`cssnano`] is a modular CSS minifier.
 * [`rtlcss`] mirrors styles for right-to-left locales.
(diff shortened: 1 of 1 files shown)


ai:
IE 9 support is not so popular, sorry

VitaliyR:
@ai say it to Chinese coders :)

ai:
We have a long fallback plugins list in special file https://github.com/postcss/postcss/blob/master/docs/plugins.md#fallbacks

README.md now containing highlighted plugins

VitaliyR:
@ai should I add it there?

ai:
Yeap

VitaliyR:
@ai 👍

PR #807, opened 2016-05-17
Added postcss-esplit plugin

VitaliyR:
Hello. I've created and polished [postcss-esplit](http://github.com/vitaliyr/postcss-esplit) plugin - and decided to add it to readme. 

The idea inside it that we were used in our project blesscss tool which just doing the same job. But it doesn't have support for source maps. Also we though to move our project to use postcss so I decided to wrote a plugin for postcss which would do just the same thing blesscss doing but with support of sourcemaps.

What do you think? 

(Duplicate pulls/806, but changes are in correct file)


ai:
Plugins is sorted by name (but `postcss-` name prefixes is on first)

VitaliyR:
@ai didn't noticed it. Fix

PR #808, opened 2016-05-18
Use CommonJS TypeScript export

danielmoore:
This PR changes the type definition export from and ES6 export (`export default`) to a CommonJS export (`export =`). This is necessary because of the `add-module-exports` babel plugin, which appends `module.exports = exports['default']` to the main file. Because of this, TypeScript's translation of `import postcss from 'postcss'` becomes `var postcss = require('postcss').default` and does not work. Babel users don't have this problem because Babel equates `module.exports` with `exports.default` when it doesn't see the `___es6Module` flag.

## Test Cases

I've included an integration test that compiles a simple piece of TypeScript against the library and attempts to run it. 

With `export default postcss`, the following cases do **not** work.

```ts
```

```
  1) TypeScript Typings links correctly:
     AssertionError: expected [Function] to not throw an error but [TypeError: Cannot read property 'plugin' of undefined] was thrown
      at expectToRun (typings.es6:23:5)
      at Context.<anonymous> (typings.es6:122:33)
```

```ts
 import * as postcss from 'postcss';
```

```
  1) TypeScript Typings has no compile errors:
     AssertionError: One or more errors were reported.
- Property 'plugin' does not exist on type 'typeof "/node_modules/postcss/d.ts/postcss"'.
      at expectDiagnosticsEmpty (typings.es6:33:15)
      at Context.<anonymous> (typings.es6:120:15)
```
```ts
import postcss = require('postcss');
```

```
  1) TypeScript Typings has no compile errors:
     AssertionError: One or more errors were reported.
- Property 'plugin' does not exist on type 'typeof "/node_modules/postcss/d.ts/postcss"'.
      at expectDiagnosticsEmpty (typings.es6:33:15)
      at Context.<anonymous> (typings.es6:120:15)
```

With `export = postcss`, the following case does **not** work.

```ts
```

```
  1) TypeScript Typings has no compile errors:
     AssertionError: One or more errors were reported.
- Module '"/node_modules/postcss/d.ts/postcss"' has no default export.
      at expectDiagnosticsEmpty (typings.es6:33:15)
      at Context.<anonymous> (typings.es6:120:15)

  2) TypeScript Typings links correctly:
     AssertionError: expected [Function] to not throw an error but [TypeError: Cannot read property 'plugin' of undefined] was thrown
      at expectToRun (typings.es6:23:5)
      at Context.<anonymous> (typings.es6:122:33)
```

With `export = postcss`, the following cases **do** work.

```ts
import * as postcss from 'postcss';
```

```ts
import postcss = require('postcss');
```

### DIFF
--- a/d.ts/postcss.d.ts
+++ b/d.ts/postcss.d.ts
@@ -1242,4 +1242,4 @@ declare module postcss {
     interface JsonComment extends JsonNode {
     }
 }
-export default postcss;
+export = postcss;
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -21,17 +21,21 @@ gulp.task('build:lib', ['compile'], () => {
     return gulp.src('lib/*.js').pipe(gulp.dest('build/lib'));
 });
 
+gulp.task('build:typings', ['clean'], () => {
+    return gulp.src(['d.ts/**/*.d.ts']).pipe(gulp.dest('build/d.ts'));
+});
+
 gulp.task('build:docs', ['clean'], () => {
     let ignore = require('fs').readFileSync('.npmignore').toString()
         .trim().split(/\n+/)
-        .concat(['.npmignore', 'index.js', 'lib/*', 'test/*',
+        .concat(['.npmignore', 'index.js', 'lib/*', 'test/*', 'd.ts/*',
                  'node_modules/**/*'])
         .map( i => '!' + i );
     return gulp.src(['**/*'].concat(ignore))
         .pipe(gulp.dest('build'));
 });
 
-gulp.task('build', ['build:lib', 'build:docs']);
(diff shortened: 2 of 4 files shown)


ai:
I will look into it after few days. Anyway we will release it only in 5.1 release :).

BarryThePenguin:
I'm exploring the potential of using typescript in lost (referenced above). I've run into this problem too. Is there anything I can do to help?

ai:
@BarryThePenguin I will accept this PR when I finish a migration to JSDoc. Just question of time.

Sorry. I hope type definitiala is not so critical.

ai:
Hi. I am ready to merge it :). But could I ask you to rebase commit?

ai:
Seems like we will move TS files to separated package in 6.0, so we don’t need test here.

PR #809, opened 2016-05-20
Update plugins.md

michelemazzucco:


ai:
1. Use better commit message
2. Keep origin plugins order
3. Move URL to end of section as in other links.

PR #812, opened 2016-05-25
Fix TypeScript module resolution

### LINKED ISSUE #787: PostCSS import fails in TypeScript with node resolution
I have a TypeScript project with `"moduleResolution": "node"` in its `tsconfig.json` as well as `postcss` in its `package.json` `devDependency`. When I write the following, it fails at runtime with the message `TypeError: Cannot read property 'parse' of undefined`:

```typescript
import * as postcss from 'postcss';
const root = postcss.parse('');
```

This is due to a mismatch between the exported JS from `./node_modules/postcss/lib/postcss.js` and the supplied typings at `./node_modules/postcss/d.ts/postcss.d.ts`. At the end of `postcss.d.ts` the last line says:

```typescript
export default postcss;
```

Which means given `import * as p from 'postcss'`, TypeScript would emit any expression `p.x` as `p_1.default.x`. However, when requiring PostCSS directly in node, it becomes clear that that is not the correct shape of the exported module at runtime:

```
  $ node
> const postcss = require('postcss')
undefined
> postcss
{ [Function: postcss]
  plugin: [Function],
  stringify: [Function: stringify],
  vendor:
   { prefix: [Function: prefix],
     unprefixed: [Function: unprefixed] },
  parse: [Function: parse],
  list:
   { split: [Function: split],
     space: [Function: space],
     comma: [Function: comma] },
  comment: [Function],
  atRule: [Function],
  decl: [Function],
  rule: [Function],
  root: [Function] }
```

This problem seems to be in the same vein of the [Babel default export behavior problem](https://phabricator.babeljs.io/T2212) that hit many people upgrading from Babel 5 to Babel 6, except in TypeScript.

jedmao:
Fixes #787 

Sorry it took so long to figure out, @kourge.

### DIFF
--- a/d.ts/at-rule.d.ts
+++ b/d.ts/at-rule.d.ts
@@ -1,5 +1,5 @@
 import Container from './container';
-import postcss from './postcss';
+import * as postcss from './postcss';
 export default class AtRule extends Container implements postcss.AtRule {
     /**
      * Returns a string representing the node's type. Possible values are
--- a/d.ts/comment.d.ts
+++ b/d.ts/comment.d.ts
@@ -1,4 +1,4 @@
-import postcss from './postcss';
+import * as postcss from './postcss';
 import Node from './node';
 export default class Comment extends Node implements postcss.Comment {
     /**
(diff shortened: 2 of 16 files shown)


ai:
I will release it in 5.1, so it could take few week, sorry.

I need to convert all docs to JSDocs.

Is it OK?

jedmao:
Yeah, Man. Take your time. No rush here.

Issue #790, opened 2016-04-13
Solution for plugin node cache


ai:
Right now plugins often add custom `_autoprefixerDisabledCache` properties to node’s objects. Of course, it looks dirty.

Maybe we need to add some unique `Node#id`. Plugins will be able to use this ID in local cache object:

```js
var cache = {};
…
cache[node.id] = myValue;
```

delorge:
Shall this cache be avaiable as postcss.nodeCache or smth?

ai:
@delorge I think that `.id` name is better, because this uniq ID could be used not only in caching. But I am open for use cases :).

delorge:
> var cache = {};
> …
> cache[node.id] = myValue;

@ai as I can see, cache should be avaiable everywhere, am I correct? Or maybe we need to add get/set cache methods for Node class? To be honest I'm not really familiar with such use cases :)

ai:
@delorge no it is private in-plugin cache. For example, Autoprefixer has some caches and this caches should be used only inside Autoprefixer.

delorge:
@ai aye, got it. I have a solution but it breaks 30+ tests :D I guess that's is an issue as well, because tests expect harcoded values, hence I'm not sure how do I test random values (of course I do, but I need an advice how to do that right).

ai:
@delorge what tests you talk about?

If you talk about `postcss-parser-tests`, I regenerate JSONs on PostCSS AST update :)

delorge:
@ai I mean postcss unit tests. For example, one of node tests: 
```
test('toJSON() converts custom properties', t => {
    let root = new Root();
    root._cache = [1];
    root._hack = {
        toJSON() {
            return 'hack';
        }
    };
    t.deepEqual(root.toJSON(), {
        type:   'root',
        id: root.id,
        nodes:  [],
        raws:   { },
        _hack:  'hack',
        _cache: [1]
    });
});

root.id is my addition, because otherwise test fails.
```

ai:
Don’t worry about them :D.

You can fix `parser` and `stringify` tests by adding `id` cleaning here https://github.com/postcss/postcss-parser-tests/blob/master/jsonify.js

ai:
If you have same code. Could you test it in [benchmark](https://github.com/postcss/benchmark). It is important to not have lower performance.

delorge:
@ai alright, I will check parser later, thank you :)

p.s. will try to test in benchmark as well

delorge:
[...]

PR #824, opened 2016-06-20
implemented unique node id

delorge:
solution for #790

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "babel-preset-es2015-loose": "7.0.0",
     "concat-with-sourcemaps": "1.0.4",
     "eslint-config-postcss": "2.0.2",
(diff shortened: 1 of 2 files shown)


ai:
1. I released `postcss-parser-test`, please add it.
2. Don’t broke indentation ;).

delorge:
investigating problems with last commit, I can confirm that `postcss-parser-test` has nothing to do with toJSON() node method, so I decided to change appropriate tests. @ai could you please check them out?

delorge:
so what will we do about this one? toJSON method isn't really about cleaning node itself but its parent.

ai:
I will look on next week. Anyway we need to finish JSDoc.

delorge:
okay then :)

PR #831, opened 2016-06-24
Add Camelcaser to plugins.

GMchris:
Added a url to my plugin which transforms css selectors to camelcase.

https://github.com/GMchris/postcss-camelcaser/commits?author=GMchris


ai:
Could you fix plugin order and you miss a link

PR #834, opened 2016-06-27
added postcss-material-shadow-helper plugin

HarrisRobin:
This is my first postcss plugin, let me know if I did anything wrong.

This is a simple plugin to provide a helper to create shadows based on google material design specs. The values are taken from here:
https://medium.com/@Florian/freebie-google-material-design-shadow-helper-2a0501295a2d#.3a8cnueb0


ai:
I think this plugin is more for https://github.com/himynameisdave/postcss-plugins#submitting-a-new-plugin

Anyway that list is more popular and it is easy to add plugin there.

PR #836, opened 2016-06-29
Added postcss-ie8 link to the plugins in section "Other"

michelgotta:


ai:
Plugins is ordered by name (but postcss- prefix is on top to promote good naming).

PR #838, opened 2016-07-02
Fixed issue with PostCSS API link in the docs

soul-wish:
😉


marcustisater:
Good catch, thanks for PR.

soul-wish:
Not at all! That was really quick reply and accepting 👍

PR #841, opened 2016-07-07
Added my plugin autorem.

malchata:
I wrote a plugin named autorem that converts px units in CSS to rem units using PostCSS. I'd love to see it added to the PostCSS plugin registry!


ai:
Thanks for work. What is difference from other `rem` plugins?

malchata:
Damn, I guess nothing. I wrote this plugin a while ago when I didn't notice another px to rem converter, but either I failed to notice one when I checked the list back then, or one was developed in the interim. So, I guess it's up to you whether or not to include mine. Your call, man. Thanks for your speedy reply.

ai:
Yeap, good work, but I try to keep list clean from plugins without difference

malchata:
Looking at the pxtorem plugin, I'd say this one is a little more straightforward, but yeah. I'm with you. Thanks for your consideration. Take care!

PR #851, opened 2016-07-25
Update postcss.d.ts for Input interface

lexich:
Example
```typescript
import { plugin, Container, Result } from "postcss";
export default plugin("postcss-test", function() {
   return function(css: Container) {
      const input = css.source.input;
      console.log(input.css);       // haven't access to css property in typescript
                                    // but valid in javascript
   };
});
```


ai:
I will release it tomorrow

PR #854, opened 2016-07-26
Add postcss-responsive-properties plugin

alexandr-solovyov:
Add postcss-responsive-properties plugin to 'Language Extensions' section


ai:
@alexandr-solovyov c’om, look into other plugins — copy their style.

Put URL to this end (they are sorted by link length). Plugin sorted by abc (but `postcss-` prefixed plugin is on the top).

PR #858, opened 2016-07-31
Add postcss-responsive-properties plugin

alexandr-solovyov:


ai:
Why you put URL right after plugin name, if all other plugins have URL separated? ;)

ai:
Also you broke 80 columns limit and forget to place a dot to the end 2a73f7d

Be careful, always copy project style, when you contribute.

PR #862, opened 2016-08-05
Add css-declaration-sorter

Siilwyn:
Adding my PostCSS plugin [css-declaration-sorter](https://github.com/Siilwyn/css-declaration-sorter/) to the readme. I'd love if it can be included in this list.

With the last update this module supports SCSS and Less, has stable sorting and comment handling. Also it's very fast since it only touches the declarations, still working on [the benchmarks](https://github.com/Siilwyn/css-declaration-sorter/tree/benchmark).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ If you have any new ideas, [PostCSS plugin development] is really easy.
 * [`precss`] contains plugins for Sass-like features, like variables, nesting,
   and mixins.
 * [`postcss-sorting`] sorts the content of rules and at-rules.
+* [`css-declaration-sorter`] sorts CSS declarations in a certain order.
 * [`postcss-utilities`] includes the most commonly used shortcuts and helpers.
 * [`short`] adds and extends numerous shorthand properties.
 
@@ -96,6 +97,7 @@ If you have any new ideas, [PostCSS plugin development] is really easy.
 * [`rtlcss`] mirrors styles for right-to-left locales.
 
 [PostCSS plugin development]: https://github.com/postcss/postcss/blob/master/docs/writing-a-plugin.md
+[`css-declaration-sorter`]:   https://github.com/Siilwyn/css-declaration-sorter
 [`postcss-inline-svg`]:       https://github.com/TrySound/postcss-inline-svg
 [`react-css-modules`]:        https://github.com/gajus/react-css-modules
 [`postcss-autoreset`]:        https://github.com/maximkoretskiy/postcss-autoreset


ai:
Sorry, we should highlight only one plugin.

And I am not happy that we have three different solutions for it (stylefmt, postcss-sorting and your plugin). Why you can’t unite with @hudochenkov and Stylelint authors?

Siilwyn:
Well `stylefmt` has a different purpose from the other two and is more similar to `perfectionist`. For `postcss-sorting` I have no idea why @hudochenkov didn't join forces as `css-declaration-sorter` was already there.

I do have a hunch that he has a different vision on this subject than me and that's probably the reason he created a seperate plugin. `postcss-sorting` can also clean up empty lines, sort selectors and sort prefixed declarations. While `css-declaration-sorter` only moves declarations in source files which makes it roughly two times faster than `postcss-sorting`. In my opinion this also keeps the plugin much simpler by design and configuration following the Unix philosophy: “do one thing and do it well”.

ai:
Is the performance is only difference between your plugins right now?

Siilwyn:
Not the only thing, as I stated above `postcss-sorting` does more than just sorting declarations. Also the predefined configurations are different, the configurations of `css-declarations-sorter` are approved by their authors and as I said in my opinion much simpler.

I would be open to a conversation on if we can merge these two plugins together, but in the meantime I would really appreciate if `css-declaration-sorter` is also listed.

ai:
It is hard question, it is like two your child come to you and ask “who you love more” :D.

Anyway, I think, right now, that we should promote postcss-sorting. My reasons:

1. It has better name according PostCSS Plugins Guidelines.
2. I think that this type of plugins should sort at-rules. Users need a solutions, not a one-type-of-node-tools. Many users use Sass. Some uses PreCSS. All of them need to sort rules and at-rules. Also our main competitor here, CSSComb, sort mixins as well.

This is why I think, that we need to select postcss-sorting. And we can’t select both, because it is highlighted list (we still have full plugins list in `docs/plugins.md`).

ai:
Maybe you should add your link to `postcss-sorting` README.md? Like:

> This plugin sorts at-rules, rules and other node types. If you want sort only declarations, you could do it faster, by [css-declaration-sorter]

Siilwyn:
What are your thoughts @hudochenkov?

hudochenkov:
> I have no idea why @hudochenkov didn't join forces as css-declaration-sorter was already there.

@Siilwyn, I almost finished my plugin, when your plugin was released. I've checked your plugin, but it only can sorting properties in non-nested rules. My plugin is an improved version of CSSComb's functionality. That is why I released my plugin anyway.

Right now I'm working on stylelint plugin, which will use `postcss-sorting` for checking the order of rules content. And afterward, I will integrate `postcss-sorting` in stylefmt, which is reformatting style sheets based on stylelint config. So `postcss-sorting` won't be just a plugin, it will be an engine for features in high-level tools.

We can join forces and make `postcss-sorting` faster if you like. Maybe you have suggestions?

P. S. BTW, I run my plugin a lot in an editor. And even on large stylesheets, it works pretty fast, almost instant. It's not numbers; it's user perception.

ai:
@hudochenkov I like the idea of integration with Stylelint in this way!

jeddy3:
@ai FYI, the `stylelint-declaration-block-order` plugin is discussed in https://github.com/stylelint/stylelint/issues/1500.

Siilwyn:
@hudochenkov thank you for clearing that up, that's a bit of unfortunate timing.

For the aim of being a part of high-level tools I have the same idea for `css-declaration-sorter` but even a bit smaller as I mentioned before.

The performance may indeed not be directly noticeable if used without any other processing, my idea of performance is that it should be as fast as possible so that when building the time of all build steps together is as short as possible. Because `postcss-sorting` also cleans newlines and processes at-rules it will always be slower than `css-declaration-sorter`, if I find the time I'll look if there are other performance wins for `postcss-sorting` to get.

PR #868, opened 2016-08-22
Added `postcss-attribute-case-insensitive` plugin

Semigradsky:
Specification: https://www.w3.org/TR/selectors4/#attribute-case
Plugin documentation: https://github.com/Semigradsky/postcss-attribute-case-insensitive


ai:
Please fix order. Plugins are ordered by name (but `postcss-` prefixed plugins on top). URLs are ordered by length.

PR #873, opened 2016-08-29
Forced Variable Plugin update to plugins.md 

alekhrycaiko:
added postcss-forced-variables to reporters section.


ai:
Plugins is sorted by name (with `postcss-` on top). Links are sorted by length.

PR #874, opened 2016-08-29
postcss-forced-variables update to plugins.md

alekhrycaiko:
I've updated the link ordering to match length & alphabetized my plugin name as mentioned in the prior PR.


alekhrycaiko:
Need to remove prior commit, closing.

PR #877, opened 2016-09-04
Various terminal highlighter fixes

lydell:
I basically fixed the same bugs that I’ve fixed in babel-code-frame before ;)

See each commit for details.

### DIFF
--- a/lib/css-syntax-error.es6
+++ b/lib/css-syntax-error.es6
@@ -3,14 +3,6 @@ import supportsColor from 'supports-color';
 import terminalHighlight from './terminal-highlight';
 import warnOnce          from './warn-once';
 
-function pad(number) {
-    let result = '';
-    for ( let i = 0; i < number; i++ ) {
-        result += ' ';
-    }
-    return result;
-}
-
 /**
  * The CSS parser throws this error for broken CSS.
  *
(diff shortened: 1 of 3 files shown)


lib/css-syntax-error.es6
```diff
@@ -162,8 +162,9 @@ class CssSyntaxError {
         if ( color ) css = terminalHighlight(css);
 
         let lines = css.split(/\r?\n/);
-        let start = Math.max(this.line - 3, 0);
-        let end   = Math.min(this.line + 3, lines.length);
+        let validLine = Math.min(Math.max(this.line, 0), lines.length);
```
ai: How `error.line` could be outside `0..lines.length`?

lydell: Ideally never. But if we ever introduce a bug in the tokenizer or parser, or – more likely – if somebody else’s custom parser does, it is possible. Take this line from tokenize.es6 as an example:

```js
throw input.error('Unclosed ' + what, line, pos - offset);
```

It can be very easy to mess up the `line` count or accidentally write `pos + offset` or something.

So this is all about robustness, so that the error reporting system doesn’t hide the _real_ error :) But I can remove this commit if you want :)

ai: Could you, please? :) Because I afraid that code was too complicated. We will add it back on first report :).

lydell: Done :)

ai:
Awesome :)

PR #890, opened 2016-09-19
Add postcss-filter-tint plugin to plugin list

alexlibby:
This plugin adds a tint to any element, such as images - feedback welcome!


ai:
Could you make other PR with fixed order? Plugins are ordered by name (by `postcss-` prefix is on the top), links are ordered by length.

PR #893, opened 2016-09-25
Update .npmignore

fson:
Remove docs from the published npm package.


ai:
This `docs/` could be using during offline development. But thanks for look-out :).

PR #894, opened 2016-09-26
Avoid calls to path.relative when the input file name is not available

mlocati:
Does it make sense to call `path.relative` when the input style does not have associated a path (eg it's `<input css 1>`)?
I'm not 100% sure about this fix, but in my brand new [PHP port of PostCSS](https://github.com/mlocati/postcss) the integration test with bootstrap.css (tokenize it and re-create it) has a huge performance improvement (about 2x fast)...

### DIFF
--- a/lib/map-generator.es6
+++ b/lib/map-generator.es6
@@ -166,6 +166,7 @@ export default class MapGenerator {
 
     relative(file) {
         if ( /^\w+:\/\//.test(file) ) return file;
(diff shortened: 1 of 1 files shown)


ai:
Fix looks good 👍.

2 times boost? Awesome!

I will try this PR few hours after (need to go to shop right now).

ai:
Released in 5.2.1

PR #897, opened 2016-09-29
Improve syntax highlight (sync with babel-code-frame)

lydell:
Before:

screenshot from 2016-09-29 21-43-09

After:

screenshot from 2016-09-29 21-42-38

See each commit for more details.

I also have a more crazy branch where I:

- Post-process the `tokens` array to create `,` tokens in some cases, to be able
  to highlight some commas.
- Try to highlight property names, without false positives for pseudo-classes.

The code of that crazy branch is really ugly. I don't think it's worth it. Here's what it looks like:

screenshot from 2016-09-29 21-46-54

### DIFF
--- a/lib/css-syntax-error.es6
+++ b/lib/css-syntax-error.es6
@@ -160,7 +160,7 @@ class CssSyntaxError {
 
         let css = this.source;
         if ( typeof color === 'undefined' ) color = supportsColor;
-        if ( color ) css = terminalHighlight(css, this.line, this.column);
+        if ( color ) css = terminalHighlight(css);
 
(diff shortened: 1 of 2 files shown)


ai:
Awesome! You was really fast.

Sure we could left crazy things. It is not so important feature to be 100% accurate.

I will accept and release tomorrow (sorry, in the train right now).

lydell:
I amended the last commit to also highlight the second `url()`. Here's the latest output:

screenshot from 2016-09-29 21-57-37

ai:
Released in 5.2.4.

PR #910, opened 2016-11-02
Added postcss-rtl plugin to plugins list

vkalinichev:
Added `postcss-rtl` plugin that generates RTL-styles and combines both-directional styles in one file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ If you have any new ideas, [PostCSS plugin development] is really easy.
 
 * [`lost`] is a feature-rich `calc()` grid system.
 * [`cssnano`] is a modular CSS minifier.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,6 +93,7 @@ If you have any new ideas, [PostCSS plugin development] is really easy.
 
 * [`lost`] is a feature-rich `calc()` grid system.
 * [`cssnano`] is a modular CSS minifier.
+* [`postcss-rtl`] combines both-directional (left-to-right and right-to-left) styles in one CSS.
```
RyanZim: Perhaps:

> combines both-directional (left-to-right and right-to-left) styles in one CSS file.

would be clearer.

ai:
Please fix plugins order (by ABC, `postcss-` prefix on top)

PR #922, opened 2016-11-23
remove Node 0.12 requirement in guidelines

sylvainpolletvillard:
As discussed on Gitter

Node 0.12 ends maintenance support at the end of this year. I think we should drop this requirement in the plugin guidelines and replace it by the active LTS, which would be an intemporal statement. Currently this means Node v4

I'm not sure if it is useful to update the boilerplate too, or let the plugin author decide which older versions of Node he wants to support.


RyanZim:
:+1: For updating the boilerplate.

PR #930, opened 2016-12-09
add postcss-bidirectional to README

gasolin:
`postcss-bidirectional` is planed to be used for firefox devtools debugger to provide cross browser RTL support.

https://github.com/gasolin/postcss-bidirection/issues/1


ai:
Now we has plugins list in `docs/plugins.md`. Could you make other PR and add your awesome plugin there?

BTW, do you have any link about `postcss-bidirectional` and Firefox? I want to promote it in twitter :)

PR #932, opened 2016-12-19
To determine the name of the processor

GitScrum:
It is necessary to determine the exact name of the processor.
Now it's not obvious - https://github.com/post-org/post-load-plugins/blob/master/src/post-processor.js

### DIFF
--- a/lib/processor.es6
+++ b/lib/processor.es6
@@ -32,7 +32,7 @@ class Processor {
          * const processor = postcss([autoprefixer, precss]);
          * processor.plugins.length //=> 2
          */
-        this.plugins = this.normalize(plugins);      
+        this.plugins = this.normalize(plugins);
         /**
          * @member {string} - To determine the name of the processor
          *
--- a/lib/processor.es6
+++ b/lib/processor.es6
@@ -40,7 +40,7 @@ class Processor {
          * see https://github.com/post-org/post-load-plugins#nodejs-with-postcss
          *
          */
-        this.processor = 'postcss';
+        this.name = 'postcss';
     }
 
     /**
--- a/lib/processor.es6
+++ b/lib/processor.es6
@@ -33,6 +33,15 @@ class Processor {
          * processor.plugins.length //=> 2
          */
         this.plugins = this.normalize(plugins);
+        
+        /**
+         * @member {string} - To determine the name of the processor
+         *
+         * @example
+         * see https://github.com/post-org/post-load-plugins#nodejs-with-postcss
+         *
+         */
+        this.processor = 'postcss';
     }
 
     /**
--- a/lib/processor.es6
+++ b/lib/processor.es6
@@ -32,8 +32,7 @@ class Processor {
          * const processor = postcss([autoprefixer, precss]);
          * processor.plugins.length //=> 2
          */
-        this.plugins = this.normalize(plugins);
-        
+        this.plugins = this.normalize(plugins);      
         /**
          * @member {string} - To determine the name of the processor
          *


ai:
What is use case for it?

GitScrum:
@ai this need for create presets plugins
1. example `post-preset-modules`
```js
export default () => (ctx, res) =>[import(`${res.processor.name}-modules`)(/* options */)(ctx, res)];
```

2.example `post-preset-mini`
```js
export default () => (ctx, res) => {
  if (res.processor.name === 'postcss') {
    return [require('postcss-csso')(/* options */)(ctx, res)];
  }

  if (ctx.processor.name === 'posthtml') {
    return [require('posthtml-minifier')(/* options */)(ctx, res())];
  }
}
```

I use it to configure and download plugins to process. See [post-load-plugins](https://github.com/post-org/post-load-plugins) and [example](https://github.com/post-org/post-load-plugins#nodejs-with-postcss)

ai:
@GitScrum and why you need to create a plugin, which could be used in PostCSS and PostHTML?

GitScrum:
@ai have you seen the plugin [post-load-plugins](https://github.com/post-org/post-load-plugins) ? One plugin to create a configuration and download plug-ins you think are bad ? Seal up all its use is not to dwell on PostCSS and PostHTML.

> and why you need to create a plugin, which could be used in PostCSS and PostHTML?

For convenience - no need to search, choose plug-ins if you need to mark up the letter. Can connect `post-preset-email` and work.

I'm sure I'll use `post-preset-mini` in [post-email](https://github.com/GitScrum/post-email) - readme not ready.

ai:
Why not:

```js
postLoadPlugins(…).css.process(…)?
```

`.css` method could call `postcss()` inside.

ai:
If `.css` will return a `Processor` you could write even:

```
postcss([
  extraPlugin,
  postLoadPlugins(…).css
])
```

GitScrum:
@ai 
> Why not:
> `postLoadPlugins(…).css.process(…)?`
> .css method could call postcss() inside.

There is no need to determine process if I know him.
The main point of the plugin - run in unknown environment.

> If .css will return a Processor you could write even:
> `postcss([extraPlugin, postLoadPlugins(…).css])`

could you elaborate ?

ai:
> There is no need to determine process if I know him. The main point of the plugin - run in unknown environment.

What does it mean “unknown environment”?

> could you elaborate ?

User may want to load some plugins from `post-load-plugins`, but other set manually.

GitScrum:
> What does it mean “unknown environment”?
was meant to be unknown plug-in which process it will be ie plugin doesn't know about entourage but can define it

> User may want to load some plugins from post-load-plugins, but other set manually.
Yes but this has nothing to do with the plugin that can identify the processor.
In `post-load-plugins` I will recommend you to use extends as an option or configuration in the package.json

ai:
> was meant to be unknown plug-in which process it will be ie plugin doesn't know about entourage but can define it

Sorry, I didn’t got it. Could you show some examples?

GitScrum:
`post-load-plugins` ? )) he knows nothing about the processes in which it will run, but it can define them, while not very nice [post-processor](https://github.com/post-org/post-load-plugins/blob/master/src/post-processor.js)

ai:
He know nothing according current API? Why we can’t change from:

```js
postcss(postLoadPlugins(…))
```

to:

```js
postcss(postLoadPlugins(…).css)
```

GitScrum:
> He know nothing according current API?

knows nothing, because it can be run in `posthtml` and maybe the list will grow.

```js 
posthtml(postLoadPlugins(…))
```

ai:
Why we can’t change to `postcss(postLoadPlugins(…).postcss)`?

GitScrum:
The plugin should load only what is to the process.
I aggregate the plugins and the configuration for them based on process name and connect through the process.

Perhaps you mean that I upload all at once ?

```js
postLoadPlugins(…) => {
  postcss: function which returns array ? right ?
  posthtml: function which returns array ? right ?
}
```

suppose that we have a situation

```js
postunknown(postLoadPlugins(…).???)
```
how to understand what process name I need ?

ai:
I think user will just copy some example from `post-load-plugins` docs.

If `post-load-plugins` really become popular we could add special support for it.

GitScrum:
ok, thanks.

PR #944, opened 2017-01-05
[WIP] Porting csstree tokenizer

hzlmn:
It is work in progress patch with attemp to merge `csstree` tokenizer in `postcss` .

**Description**:
 
Mostly because `csstree` tokenizer much more detailed and works with lower structures, most of "heavy lifting" done in parser part. 

Because of that, if we fully integrate `csstree` we need rewrite most of parser parts to support new tokens system that works with current "array-based" structure.

The point is to minimalise efforts for updating parser. We can use `csstree` scanner that will operate character structures and then plug it to updated `tokenizer` where it produces on each `readNextToken` call token compatible with current token structure and parser.

Obviously need feedback of @lahmatiy as I may miss something.

If this structure works fine for you @ai, I will continue work.  If not, I don't see reasons to integrate `csstree` tokenizer as it can be effortless to just upgrade current `tokenizer`.

**Roadmap**:
- [x] Port initial structures from `csstree`
- [x] Fix linting issues
- [ ] Integrate `csstree` output with current token types (in progress)
- [ ] Update parser for support streaming (in progress)
- [ ] Fix tests related to new architecture
- [ ] More?

### DIFF
--- a/lib/character.es6
+++ b/lib/character.es6
@@ -1,0 +1,18 @@
+/* Single chacracter structure */
+export class Character {
+    constructor(type, value, start, end, position) {
+        this.type = type;
+        this.value = value;
+        this.start = start;
+        this.end = end;
+        this.position = position;
+    }
+
+    /* Get character position */
+    getPosition() {
+        return {
+            line: this.position.line,
+            column: this.position.column
+        };
+    }
+}
--- a/lib/new-tokenizer.es6
+++ b/lib/new-tokenizer.es6
@@ -32,7 +32,7 @@ const RE_BAD_BRACKET = /.[\\\/\("'\n]/;
 export class Tokenizer {
     constructor(input, options = {}) {
         this.scanner = new Scanner(input.css.valueOf(), options);
-        this.currentChar = {};
+        this.currentChar = null;
         this.currentToken = null;
         this.buffer = [];
     }
--- a/lib/new-tokenizer.es6
+++ b/lib/new-tokenizer.es6
@@ -1,0 +1,159 @@
+/* eslint no-unused-vars: 0, no-undef: 0 */ // turning off only for now
+
+import { Scanner } from './scanner';
+import { Character } from './character';
+
+const SINGLE_QUOTE      = '\''.charCodeAt(0);
+const DOUBLE_QUOTE      =  '"'.charCodeAt(0);
+const BACKSLASH         = '\\'.charCodeAt(0);
+const SLASH             =  '/'.charCodeAt(0);
+const NEWLINE           = '\n'.charCodeAt(0);
+const SPACE             =  ' '.charCodeAt(0);
+const FEED              = '\f'.charCodeAt(0);
+const TAB               = '\t'.charCodeAt(0);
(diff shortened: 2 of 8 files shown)


ai:
I thought you will take `csstree` package, but you want to rewrite it from scratch?

I think it will be better to have one tokenizer for both project. Is it possible?

hzlmn:
Hah, you told backporting, so what i actually thought is to take tokenizer part from `csstree` srcs and port it to `postcss`. 😄  

If purpose is to use `csstree` then it is a question to @lahmatiy , because current version do not expose tokenizing ability and it involves modifying sources.

Btw i don't like idea of adding dependency to heavy full featured parser only for tokenizing part, but as i see that project is work in project, so if splitting it to `tokenizer` , `parser` packages is an option, it can be easier.

ai:
OK, let’s start from backporting until we will have some feedback from @lahmatiy

lahmatiy:
Hi, all!

Yep, `csstree` doesn't expose tokenizer since I thought it's an internal part and doesn't need to others. But recently I realized that it can be useful for other projects and tasks. So I'm going to expose tokenizer. But I'm not sure it's to be a separate package right now because of additional complexity for current state of `csstree`. Although it's possible in future.

You're free to backport tokenizer if you want. I can't deny you to do this. But you should take into account that you'll need tests and related things too. Also I'm not sure your changes are good for performance and you'll need tune performance by yourself. If you're happy with it - no problems ;)

I believe csstree's tokenizer is not perfect yet and will be improved in future. So maybe using it will be better choice.

In any case, it is worth taking a decision after you try to integrate tokenizer as is it now. If everything is ok, you'll back to make a decision dependency vs backport. Otherwise the only choice will be to backport tokenizer with the necessary modifications.

hzlmn:
@lahmatiy  Thanks for your feedback and your work on `csstree`. 

Due to your active work on improving  it, I think it's waste of time to backport it to `PostCSS` and as you mentioned could also be much effort to support as well. When `csstree` API will be ready, I can integrate it to postcss.

For now @ai, I can update current tokenizer or we can wait until `csstree` will be "stable".

ai:
@hzlmn sure, let’s update current tokenizer.

@lahmatiy I think tokenizer unification will be great for both project. You will get much more users. PostCSS will get faster parser (but right now main performance problem is multiple AST walks). Tell me if you need help with API, I will try to use Cult of Martians to bring developers.

lahmatiy:
@hzlmn [v1.0.0-alpha10](https://github.com/csstree/csstree/releases/tag/v1.0.0-alpha10) is out and `csstree` exposes tokenizer now (as `Scanner` class).

`csstree` in alpha stage since AST format is not stable yet. But it doesn't affect tokenizer in any way. I don't think its API will change much. So you can start to experiment with it. Feedback is welcome.

As I mention above, I don't like to extract tokenizer to package right now. But it's ready for it. If you success with your experiments and you decide to use it in PostCSS, it will be a reason to make a separate package.

@ai Thank you for your participation. Tokenizer API is stable enough and nothing prevents to try to use it. I need a feedback first. Thank you one more time.

PR #949, opened 2017-01-09
Fix TypeScript module resolution

### LINKED ISSUE #948: Fix TypeScript module resolution
TypeScript now has [better documentation explaining the publishing process of definition files](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html). Use this information to update file paths and module resolution so that it works properly in vscode and other supported editors.

jedmao:
Fixes #948

This PR doesn't add any updated PostCSS updates to the TypeScript definitions, but it has been compiled with the most recent TypeScript compiler and updated to follow the most recent guidance given by the TypeScript team with respect to how to publish definition files.

Refer to the [Publishing TypeScript](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) page for more information.

_Note: I did modify the gulpfile so that `postcss.js` and `postcss.d.ts` can exist side-by-side. This is the recommended way to do it._


ai:
Thanks. Released in 5.2.9.

PR #957, opened 2017-01-20
Revert "Fix TypeScript module export"

### LINKED ISSUE #954: No longer able to import module in typescript
As of version 5.2.10, we are not sure how to use postcss with typescript.

We import the module via:
`import * as postcss from 'postcss';`

And we could call like:
`postcss().process(mycss).then((result) => { ... }`

However, as of 5.2.10, when attempting to compile we get error like:
> Cannot invoke an expression whose type lacks a call signature. Type of 'typeof "postcss"' has no compatible call signatures.

Attempting to change the import to:
`import postcss from 'postcss';`

Will compile okay, but causes issues when running:
> TypeError: postcss_1.default is not a function

What are we missing to get v5.2.10 to work with typescript? For the moment, we are using 5.2.9.

jedmao:
Reverts #950 

Fixes #954 

Sorry for the wait, everybody. I had to do some testing to reproduce the problem on my end first so that I could verify this was the issue. Just need to revert this change.

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -1303,5 +1303,5 @@ declare module "postcss" {
         interface JsonComment extends JsonNode {
         }
     }
-    export default postcss;
+    export = postcss;
 }


ai:
We are ready for release?

jedmao:
Yes. I tested this out with my `postcss-nested-props` plugin to make sure it works and it does. The reason I "thought" this commit worked before is because I wasn't getting in-editor errors, but I needed to run the tests to see that it would fail. Definitely got it right this time.

ai:
Released in 5.2.11.

PR #960, opened 2017-02-05
Fix typo in generated() deprecation message

garetmckinley:
In the deprecation message associated with the `generated()` function, the word deprecation was spelled `depreacted`. This is incredibly minor, but it's something I've run into a fair amount of times recently when searching for `react` inside my library's bundle.


ai:
Released in 5.2.12

PR #962, opened 2017-02-06
Remove js-base64 from dependencies

lahmatiy:
PostCSS is currently using `js-base64` for encode/decode base64 content of source maps. But there is a "standart" way in node.js world to do it using Buffer API.
Yep, Buffer doesn't supported by browsers. But it will be included if build by bundlers. I tested with browserify, everything ok. Moreover `js-base64` also uses Buffer when possible, so `buffer` module includes in bundle by browserify anyway.
I believe changes doesn't break much, but minus one redundant dependency.

### DIFF
--- a/lib/map-generator.es6
+++ b/lib/map-generator.es6
@@ -1,4 +1,3 @@
-import { Base64 } from 'js-base64';
 import   mozilla  from 'source-map';
 import   path     from 'path';
 
@@ -126,7 +125,7 @@ export default class MapGenerator {
 
         if ( this.isInline() ) {
             content = 'data:application/json;base64,' +
-                       Base64.encode( this.map.toString() );
+                       new Buffer( this.map.toString() ).toString('base64');
 
         } else if ( typeof this.mapOpts.annotation === 'string' ) {
             content = this.mapOpts.annotation;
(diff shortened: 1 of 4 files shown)


ai:
Great idea. Is it OK to release it in 6.0 branch, because of breaking changes

lahmatiy:
As you wish ;)
I don't think it could break anything.
May be need more testing with bundlers other than Browserify.

ai:
I agree that it looks safe.But for my opinion it is “changing of public API” :). Anyway I plan this major release on next month. Is it OK or it is time sensitive?

lahmatiy:
Don't worry, no rush at all.
Just interesting a little bit why do you think it changes the public API? But it's not so important, so you shouldn't answer ;)

ai:
It is always better to make major release, even if update will not hurt anyone, that make patch release and broke some environments :D. I have so many bad feedback on unexpected compatibility problems on minor/patch releases :(.

So I agree that most of users will not saw API changes, but because API changes theoretically exists, it will be better to make it as major just to be sure.

ai:
(For me every patch release is good thing, because often patch release increase download counter :D)

ai:
I merged it to 6.0 branch manually

PR #980, opened 2017-02-23
add postcss-image-set-polyfill to plugins list

SuperOl3g:
Hi there!
I've made a small css func polyfill, hope somebody will find it very useful. 
So I propose add it to plugins list.


ai:
What is different with [postcss-image-set](https://github.com/alex499/postcss-image-set)?

SuperOl3g:
[alex499 version](https://github.com/alex499/postcss-image-set) just fallback to first image for all no-support browsers.
My version fully simulate `image-set` function logic.

ai:
Wow, great!

SuperOl3g:
Thanks)

PR #970, opened 2017-02-14
add postcss setup instructions for webpack 2

hidace:
I was able to get post-css working with Webpack 2 using this simple configuration (however, wasn't able to get post-css and webpack 2 working with the original instructions for webpack). 
Please feel free to confirm. Thanks!

PR #988, opened 2017-03-11
Add config for Webpack 2

mrmartineau:
This PR is intended to fix the hacks in #970. I have included multiple ways to compile PostCSS with Webpack: inline (default webpack behaviour) or separate CSS (using the extract-text-plugin).

I also added a little more context to the config because I thought that user's might need the extra info; see the `entry`, `output` and `plugins` keys in the two code blocks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,20 +178,83 @@ You can start using PostCSS in just two steps:
 
 Use [`postcss-loader`] in `webpack.config.js`:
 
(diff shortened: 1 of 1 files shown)


ai:
Too many details (bundled vs. extract) for common PostCSS docs. This details should be only in `postcss-loader`

mrmartineau:
@ai Shall I remove the extra content and resubmit?

ai:
Absolutely! :)

PR #989, opened 2017-03-11
Add config for Webpack 2

mrmartineau:
This PR is intended to fix the hacks in #970. I have included multiple ways to compile PostCSS with Webpack: inline (default webpack behaviour) or separate CSS (using the extract-text-plugin).

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,14 +181,7 @@ Use [`postcss-loader`] in `webpack.config.js`:
 Use this if you are bundling your CSS into your JS (the default Webpack behaviour)
 
 ```js
-const path = require('path');
-
 module.exports = {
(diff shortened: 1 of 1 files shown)


ai:
You still have two configs :D

mrmartineau:
@ai I know, I meant to keep two because I think its important to know how to extract the CSS into a new file. If you still don't think that's appropriate, perhaps we can link to a gist or something..

ai:
I agree that it is important information. But this docs just overview.

We can add docs to `postcss-loader` and set link from section. In this docs we could keep only one basics config example.

mrmartineau:
sure, ok. I'll resubmit a fresh and clean PR later

PR #996, opened 2017-03-26
Add postcss-ant to list of plugins

corysimmons:


ai:
We need better description. Right now it tells about features for developers. But we need to describe a benefits for users. Why they should take care about “autosizes and more”?

corysimmons:
tbh, I've never been sure how to describe it in a vague enough way to cover all its use-cases that won't take a paragraph.

"Returns any size you could ever need in any situation" ?

ai:
Why users even need some size? ;) What use cases?

corysimmons:
Why do users need precise "sizes" in design-oriented CSS?

Grid widths. Positioning elements precisely.

PR #997, opened 2017-03-28
Update plugins.md

kunukn:
Add reference to postcss-alter-property-value plugin


ai:
1. Fix plugins order.
2. `Alter` → `alters`. It is just a verb after your plugin name.
3.  `css` → `CSS`. It is abbr.

PR #1006, opened 2017-04-21
postcss-nested-import

eriklharper:
Hello!  I [created a new postcss plugin that allows for `@import` statements to work within nested css statements](https://github.com/eriklharper/postcss-nested-import).  I'd like to get this reviewed by the PostCSS community and also would like your help in determining why my tests are failing, which I think is due to how postcss processes the results, and not from a bug in my plugin code.  I used the PostCSS plugin boilerplate, so you can clone my project, run `yarn` and then `jest` to see the result of the test which looks like this:

```
FAIL  ./index.test.js
  ● replaces @import with the contents of the file being imported inside a declaration block.

    expect(received).toEqual(expected)

    Expected value to equal:
      ".vendor { background: silver; } .vendor-font { font-size: 14px; }"
    Received:
      ".vendor {
        background: silver
    }
    .vendor-font {
        font-size: 14px
    }"
```

It appears that when postcss runs and outputs a result, it adds line breaks to the code and also removes the semicolons from the end of the statements for some reason.  Not sure why this is happening, but advisement would be much appreciated.

Thanks!
Erik Harper

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -287,6 +287,7 @@ loosely resembles the original.
 * [`postcss-combine-duplicated-selectors`] automatically join identical css selectors.
 * [`postcss-filter-mq`] Filter all matching or non-matching media queries.
 * [`postcss-import`] inlines the stylesheets referred to by `@import` rules.
+* [`postcss-nested-import`] inlines stylesheets referred to by `@import` rules inside nested rule blocks.
 * [`postcss-partial-import`] inlines standard imports and Sass-like partials.
 * [`postcss-reference`] emulates Less’s [`@import (reference)`].
 * [`postcss-remove-root`] removes all instances of `:root` from a stylesheet.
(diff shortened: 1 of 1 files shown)


ai:
Thanks for new plugin 👍

eriklharper:
Thanks, @ai that was very prompt!  Could you perhaps tell me what is causing that issue with my test above? ☝️

ai:
I do not see any issue

eriklharper:
interesting, perhaps its something in my environment..

michael-ciniawsky:
@eriklharper You need to add `\n` to the expected output for the tests to pass 😛 

```js
'.vendor {\n background: silver;\n }\n .vendor-font {\n font-size: 14px;\n }\n'
```

maybe a space here and there, the string comparison is strict about indentation and lines breaks

eriklharper:
why does the output strip the semicolons?

ai:
@eriklharper right now fix output as in Jest expected output and wait for PostCSS 6.0 where we fixed `Node#raws` API.

PR #1007, opened 2017-04-21
Add Browsers support in map-generator.es6

h0tc0d3:
Buffer we have only in Node.JS.

### DIFF
--- a/lib/map-generator.es6
+++ b/lib/map-generator.es6
@@ -123,20 +123,12 @@ export default class MapGenerator {
     b64Encode(str) {
         let b64str;
         if (Buffer) { // In Node.JS
-            if (Buffer.from) {
-                b64str = Buffer.from(str).toString('base64');
-            } else {
-                b64str = new Buffer(str).toString('base64');
-            }
+            b64str = Buffer.from ?
+                Buffer.from(str).toString('base64') :
+                new Buffer(str).toString('base64');
         } else { // In Browsers
-            /* global btoa */
-            /* eslint no-undef: "error" */
-            b64str = btoa(encodeURIComponent(str).replace(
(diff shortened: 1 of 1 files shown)


ai:
Should we move it to function on the top of file to make method cleaner.

lib/map-generator.es6
```diff
@@ -120,12 +120,30 @@ export default class MapGenerator {
         }
     }
 
+    b64Encode(str) {
+        let b64str;
+        if (typeof Buffer !== 'undefined') { // In Node.JS
+            b64str = new Buffer(str).toString('base64');
```
michael-ciniawsky:
```js
Buffer.from(str).toString('base64')
``` 

`New Buffer` is [deprecated](https://nodejs.org/dist/latest-v7.x/docs/api/buffer.html#buffer_new_buffer_array) and will throw in node >= v8

lib/map-generator.es6
```diff
@@ -122,8 +122,12 @@ export default class MapGenerator {
 
     b64Encode(str) {
         let b64str;
-        if (typeof Buffer !== 'undefined') { // In Node.JS
-            b64str = new Buffer(str).toString('base64');
+        if (Buffer) { // In Node.JS
+            if (Buffer.from) {
+                b64str = Buffer.from(str).toString('base64');
+            } else {
+                b64str = new Buffer(str).toString('base64');
```
michael-ciniawsky: Is this needed ? 🙃

h0tc0d3: I do not know for sure. Added just in case. I had a case with Uint8Array where not will be from and subarray mthods.  

> Class Method: Buffer.from(array)
> #
> Added in: v5.10.0

PR #1015, opened 2017-05-08
Adds Initial Contributing Guidelines.

solodynamo:


ai:
It is a regular advice for any project.

For example, the most popular PR is new plugins to `docs/plugins.md`. This doc doesn’t cover it.

I suggest you to write the most popular cases first and only then start to make docs.

ai:
It is CSS parser. Most of developers never will change parser or AST ;).

For all this years I had only few PR for source map, parser or AST. So we should not focused on regular PR as in other projects.

ai:
Also we should cover “where can I help” topic. If user what to contribute to PostCSS, but don’t know how it can be useful.

solodynamo:
Okay , will come up with something incorporating most of the above suggestions :)

PR #1017, opened 2017-05-15
Improve postcss.d.ts so that type guards work

kimamula:
This PR enables type guards on the `type` property of `Node`.

```ts
function doSomethingWithNode(node: postcss.Node): string {
  switch (node.type) {
    case 'root':
      // node is inferred as Root here
      node.toResult();
      return 'root';
    case 'atrule':
      // node is inferred as AtRule here
      node.params = '';
      return 'atrule';
    case 'rule':
      // node is inferred as Rule here
      node.selector = '';
      return 'rule';
    case 'decl':
      // node is inferred as Declaration here
      node.important = false;
      return 'decl';
    case 'comment':
      // node is inferred as Comment here
      node.text = '';
      return 'comment';
    // default is not necessary as TypeScript knows the above cases are collectively exhaustive
  }
}
```

Additionally, the `parent` property of `Root` is altered from `Container` to `void`, as it "Should always be undefined for a Root node" according to its comment.

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -612,7 +612,8 @@ declare module "postcss" {
              */
             origin(line: number, column: number): InputOrigin;
         }
-        type Node = Container | Declaration | Comment;
+        type ChildNode = AtRule | Rule | Declaration | Comment;
+        type Node = Root | ChildNode;
         interface NodeBase {
             /**
              * Returns the input source of the node. The property is used in source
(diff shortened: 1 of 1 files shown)


ai:
@jedmao what do you think?

kimamula:
Added another commit to strictly distinguish `Root` from the other `Node` types.

jedmao:
Excellent changes. Thanks @kimamula!

PR #1018, opened 2017-05-15
Add postcss-hamster plugin in README.md

h0tc0d3:


ai:
`README.md` contains only the most popular and important plugins. Could you its changes from PR?

PR #1043, opened 2017-06-22
Fixed issue with unicode escapes.

### LINKED ISSUE #1042: Tokenizer: Unicode escapes with trailing whitespaces
I've noticed that postcss doesn't work with [hexadecimal unicode css escapes](https://www.w3.org/TR/css-syntax-3/#escape-diagram).

Declaration `\62 olor: red` is incorrectly parsed into
```js
{"decl": "\\62", "value": "red", /*...*/}
```
Expected is:
```js
{"decl": "\\62 olor", "value": "red", /*...*/}
```
One of the results is the fact [stylelint property-blacklist](https://stylelint.io/user-guide/rules/property-blacklist/) and many other rules can be easily tricked. And it's very unlikely that this issue can be resolved on the stylelint side.

Digging deeper I've found that issue is in tokenizer. The test string is `\0a \09 `. Calling `tokenize({css: '\\0a \\09 '})` yields
```
[ 'word', '\\0', 1, 1, 1, 2 ]
[ 'word', 'a', 1, 3, 1, 3 ]
[ 'space', ' ' ]
[ 'word', '\\0', 1, 5, 1, 6 ]
[ 'word', '9', 1, 7, 1, 7 ]
[ 'space', ' ' ]
```
expected result is all (up to 6) hex numbers and a trailing whitespace are glued to the backslash:
```
[ 'word', '\\0a ', 1, 1, 1, 4 ]
[ 'word', '\\09 ', 1, 5, 1, 8 ]
```

I understand that this change can be breaking: for example, existing plugins could possibly not expect whitespaces or newlines in property names, media queries, etc. So if you decide that it just doesn't worth it and nobody writes code this way, I'll be okay with that.

hzlmn:
Hello, sir.
This simple patch closes #1042 .

### DIFF
--- a/lib/tokenize.es6
+++ b/lib/tokenize.es6
@@ -21,7 +21,7 @@ const AT                =  '@'.charCodeAt(0);
 const RE_AT_END      = /[ \n\t\r\f\{\(\)'"\\;/\[\]#]/g;
 const RE_WORD_END    = /[ \n\t\r\f\(\)\{\}:;@!'"\\\]\[#]|\/(?=\*)/g;
 const RE_BAD_BRACKET = /.[\\\/\("'\n]/;
-const RE_HEX_ESCAPE  = /[0-9A-Fa-f]+/g;
+const RE_HEX_ESCAPE  = /[a-z0-9\s]/i;
 
 
 export default function tokenizer(input, options = {}) {
@@ -223,6 +223,7 @@ export default function tokenizer(input, options = {}) {
         case BACKSLASH:
             next   = pos;
             escape = true;
+            let hexEscape;
             while ( css.charCodeAt(next + 1) === BACKSLASH ) {
                 next  += 1;
                 escape = !escape;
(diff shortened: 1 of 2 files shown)


ai:
Wow! You were fast for such complicated task.

I will accept and release it tomorrow.

lib/tokenize.es6
```diff
@@ -21,6 +21,7 @@ const AT                =  '@'.charCodeAt(0);
 const RE_AT_END      = /[ \n\t\r\f\{\(\)'"\\;/\[\]#]/g;
 const RE_WORD_END    = /[ \n\t\r\f\(\)\{\}:;@!'"\\\]\[#]|\/(?=\*)/g;
 const RE_BAD_BRACKET = /.[\\\/\("'\n]/;
+const RE_HEX_ESCAPE  = /[a-z0-9\s]/i;
```
subzey: Is it `a-z`, not `a-f`?

hzlmn: Oops good point, thanks.

lib/tokenize.es6
```diff
@@ -234,6 +235,11 @@ export default function tokenizer(input, options = {}) {
                             code !== CR      &&
                             code !== FEED ) ) {
                 next += 1;
+                if (RE_HEX_ESCAPE.test(css.charAt(next))) {
+                    while (RE_HEX_ESCAPE.test(css.charAt(next + 1))) {
```
subzey: Looks like `\61 a` would be parsed as a whole entity, `"\\61 a"`, while it [should stop after the first whitespace](https://jsfiddle.net/subzey/6bdmL6a9/): `"\61 "`, `"a"`.

Also, there can be no more than 6 hex digits (the escape seq [ends even if there's no whitespace](https://jsfiddle.net/subzey/tu50edsv/))

hzlmn: I know, it was done, because otherwise `escapes.css` fails in parsing. You can checkout branch and check my first commit. The key point here to produce understandable tokens for parser so it's not fully spec compatible.

PR #1072, opened 2017-10-01
use declare instead of declare module

chriseppstein:
The `declare module "foo"` syntax is used to declare values ambient types for other modules. When declaring types for your own module you're supposed to just use `declare` for each exported member individually. Which for postcss isn't that many because it's mostly in a namespace.

The reason this is important is that the syntax being used currently can cause problems with typescript projects that resolve to several different versions of postcss because of the `export =` is a global in typescript and causes a conflict between the several implementations of the `declare module` blocks. or something. I don't know. This seems to work for me, but I think it will require a 5.x release to fully address the issue. At least this is what the typescript experts I know have told me.

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -1,1285 +1,1283 @@
-declare module "postcss" {
+/**
+ * @param plugins Can also be included with the Processor#use method.
+ * @returns A processor that will apply plugins as CSS processors.
+ */
+declare function postcss(plugins?: postcss.AcceptedPlugin[]): postcss.Processor;
+declare function postcss(...plugins: postcss.AcceptedPlugin[]): postcss.Processor;
+declare namespace postcss {
+    type AcceptedPlugin = Plugin<any> | Transformer | {
+        postcss: TransformCallback | Processor;
+    } | Processor;
     /**
-     * @param plugins Can also be included with the Processor#use method.
-     * @returns A processor that will apply plugins as CSS processors.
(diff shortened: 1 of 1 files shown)


chriseppstein:
Note: this change looks big because it changes the indent level, but if you turn off whitespace on the diff it's actually very small:

https://github.com/postcss/postcss/pull/1072/files?diff=unified&w=1

ai:
@jedmao do you approve?

(I doesn't know TypedScript will and @jedmao is our TS expert 😅)

ai:
> Note: this change looks big because it changes the indent level, but if you turn off

Didn't know it. Awesome!

Can we revert indent changes in this PR? Or previous indent was wrong?

chriseppstein:
@ai This change removed an enclosing block so the code inside got outdented by one level.

jedmao:
👍 thanks @chriseppstein!

chriseppstein:
@ai is the any chance that we can get this delta backported and released as `5.2.18`? without both versions released it won't fix the issue I'm seeing in my project :(

ai:
I will try to backport it and release tomorrow

ai:
5.2.18 was released with fix backport (I didn’t find good way to make 5.2.18 in Git, so I just make release without Git commit and Git tag)

ai:
6.0.13 was released too

ai:
Sorry for delay :) It is my first backport release :D

PR #1075, opened 2017-10-08
README: image-set W3 Draft URL fixed

exarus:
Fixed both in Chinese and English version.


ai:
Thanks. Especially for fixing Chinese version too.

PR #1092, opened 2017-12-07
Updated walk() to print out more information on exception

nikhilgaba:
This is a modification to the walk() method found for the Container object. The difference from the original is that it now attempts to print out more information to help a developer locate where the problem occurred.


nikhilgaba:
I blew some stuff up on my end, I will create a new PR and we can continue on this :)

PR #1093, opened 2017-12-08
Updated walk() to print out more information on exception

nikhilgaba:
This is a modification to the walk() method found for the Container object. The difference from the original is that it now attempts to print out more information to help a developer locate where the problem occurred.

### DIFF
--- a/lib/container.es6
+++ b/lib/container.es6
@@ -106,11 +106,24 @@ class Container extends Node {
      */
     walk(callback) {
         return this.each( (child, i) => {
-            let result = callback(child, i);
-            if ( result !== false && child.walk ) {
-                result = child.walk(callback);
+            try {
+                let result = callback(child, i);
+                if ( result !== false && child.walk ) {
+                    result = child.walk(callback);
+                }
+                return result;
+            } catch (exception) {
+                // If an error occurs, attempt to print
+                // out information that will narrow down
+                // where the problem can be found.
+                throw new Error(
+                    `${exception.message}
+                    File: ${child.source.input.file}
+                    CSS Rule: ${child.selector} (lines: 
+                        ${child.source.start.line}
+                        -${child.source.end.line})`
+                );
             }
-            return result;
         });
     }
 


ai:
@nikhilgaba It is an interesting idea. But we have 4 problems:

1. We need to copy error class.
2. Maybe instead of changing message, we should change .stack. Because we are not sure how the error will be caught. Maybe there is some error.message === someText above.
3. We need an option to disable it.
4. We can enable it by default only in a major release.

If you are ready to pass this way, I will help you on the way :).

ai:
How I see a process of adding it to PostCSS:

1. We change a way to add this data to error (class coping, replacing `stack`, adding extra properties to error object)
2. We add this magic to `postcss.plugin`.
3. We add an option and disable it by default.
4. We release it in 6.1. It will be switched `off` by default.
5. We will ask postcss-loader, gulp-postcss and postcss-cli to turn the option `on` in their next major release.
6. In 7.0 we will switch it `on` by default.

lib/container.es6
```diff
@@ -106,11 +106,24 @@ class Container extends Node {
...
-            let result = callback(child, i);
-            if ( result !== false && child.walk ) {
-                result = child.walk(callback);
+            try {
+                let result = callback(child, i);
+                if ( result !== false && child.walk ) {
+                    result = child.walk(callback);
+                }
+                return result;
+            } catch (exception) {
```
michael-ciniawsky: `exception` => `err`

lib/container.es6
```diff
@@ -106,11 +106,24 @@ class Container extends Node {
...
+                // If an error occurs, attempt to print
+                // out information that will narrow down
+                // where the problem can be found.
+                throw new Error(
```
michael-ciniawsky: Could this be a new Error Type (Instance) instead please ? So third party implementations can handle it correctly via e.g `if (err.name === 'WalkerError') handler...` equal to `err.name === CssSyntaxError`

```js
// But not WalkerError please, it's a really bad name :)
class WalkerError extends Error {
   constructor (err) {
    super()
   
    this.name = 'WalkerError'
    this.message = ... // handle/format message (file, message, code frame etc..)
    this.stack = ... // handle/format stack trace

    Error.captureStackTrace(this, this.constructor)
  }
}
```

```js
catch (err) {
  throw new WalkerError(err)
}
```

ai:
> 5. Error must be captured not only in walk, but also in postcss.plugin and in Promise if plugin return it to postcss.plugin.

Forget it. I was wrong in this request (there is no way to detect source node or of `walk`).

ai:
I suggest:

```js
} catch (exception) {
  if (result.opts.errorSource) {
    exception.postcssNode = child
    exception.stack = "…"
  }
  throw exception
}
```

michael-ciniawsky:
☝️ How will the thrown ```{Error}``` look like then ? How to catch && handle it correct ?

```js
postcss(plugins).process(css, options)
  .then((result) => result)
  .catch((err) => {
     // for e.g
     err.name === '?' 
       ? new PostCSSError(err) // Syntax Error, Plugin Error, API Error etc...
       : new Error(err) // Normal (JS) Error 
  })
```

ai:
@nikhilgaba what do you think about my plan?

nikhilgaba:
@ai Plan seems quite sound. I just haven't had a chance to look into the implementation.

ai:
Don't worry, I will use power of our [Cult of Martians](http://cultofmartians.com) to improve your PR.

ai:
I merged and fixed it manually https://github.com/postcss/postcss/commit/b2530e0f7a106a23eb17168093509ec65f94bb53

ai:
Released in 7.0

PR #1095, opened 2017-12-08
Provide Tokenizer, Comment and Parser as named exports

### LINKED ISSUE #1094: [Feature request] Provide Tokenizer, Comment and Parser as named exports
Hi,

Thanks for your great contributions to Open Source! <3

Here's a PR for this issue: https://github.com/postcss/postcss/pull/1095

# Problem

Currently you have to import internal modules to implement your own custom parser for PostCSS.
For example:

```javascript
import tokenizer from 'postcss/lib/tokenize';
import Comment from 'postcss/lib/comment';
import Parser from 'postcss/lib/parser';
```

Since PostCSS is set up to accept a custom parser as a configuration option, I would be very happy if you provide these as public importable modules, rather than having consumers need to tap in to the internals of PostCSS to actually implement a custom parser and/or tokenizer.

# Proposed solution

Here's what I would like to do:

```javascript
import {tokenizer, Comment, Parser} from "postcss";

class MyCustomParser extends Parser {
    // ...
}
```

To do that, you could add:
```javascript
export {tokenizer, Comment, Parser}
```

in the bottom of `/lib/postcss.js` right below the default export. The Typescript typings would need to be updated too to reflect this change.

I hope you'll consider this. Thanks!

wessberg:
This fixes #1094 

This PR provides `tokenizer`, `Parser` and `Comment` as named exports of PostCSS.
Additionally, it extends the Typescript type declarations with typings for all three types.

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -1329,7 +1329,7 @@ declare namespace postcss {
      * Comments inside selectors, at-rule parameters, or declaration values will
      * be stored in the Node#raws properties.
      */
-    interface Comment extends NodeBase {
+    class Comment extends Node {
         type: 'comment';
         /**
          * Returns the comment's parent node.
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -60,36 +60,46 @@ declare namespace postcss {
         endOfFile(): boolean;
     }
 
+    type TokenKind = "[" | "]" | "(" | ")" | "{" | "}" | ";" | ":" | "space" | "brackets" | "string" | "at-word" | "word" | "comment";
+    type StringToken = ["string", string, number, number, number, number];
+    type WordToken = ["word", string, number, number, number, number];
+    type SpaceToken = ["space", string];
+    type AtWordToken = ["at-word", string, number, number, number, number];
+    type CommentToken = ["comment", string, number, number, number, number];
+    type BracketsToken = ["brackets", string, number, number, number, number];
+    type SyntaxToken = [ "[" | "]" | "{" | "}" | ":" | ";" | "(" | ")", string, number, number];
+    export type Token = StringToken|WordToken|SpaceToken|BracketsToken|AtWordToken|CommentToken|SyntaxToken;
+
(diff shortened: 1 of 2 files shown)


ai:
Will it continue work with `postcss = require ("postcss")`

wessberg:
This works:
```javascript
import postcss, {Parser, tokenizer} from "postcss";
postcss();
```

But this doesn't:
```javascript
const postcss = require("postcss");
postcss();
```

Since the commonjs transpiled version binds the default export to the "default" property of the `module.exports` object (so you'd have to do `require("postcss").default`). It does so because there are named exports and it has to provide a namespace object.

This is why it breaks now that there are named exports.

Four ways forward as I see it:

1) Provide PostCSS as a named export also. As a benefit, this also means that people using Typescript + Rollup doesn't have to force `allowSyntheticDefaultImports": true"` in their `tsconfig.json`.

2) Accept that for clients using commonjs, they have to do `require("postcss").default`. Clients using ES-modules will still be able to do `import postcss from "postcss";`.

3) Reject this PR and only export `postcss`. Do nothing else

4) Reject this PR and only export `postcss`. Put the Parser and Tokenizer into a separate library under the postcss organization on Github (such as `postcss/postcss-parser`) and export the Parser and Tokenizer from there. Depend on it from postcss/postcss.

ai:
Note that we already have `add-module-exports` Babel plugin.

Keeping `require` API is critical since users will kill me if I will break API often than once PR year.

wessberg:
Yeah, I guess they changed it in Babel 6 to improve interop with ES-modules.

I totally understand. No one likes breaking changes. So let's not do that.

Would you consider providing the parser (and tokenizer, etc) from a separate library, like you do with https://github.com/postcss/postcss-safe-parser and depending on it from `postcss`?

wessberg:
Great news! I'm closing this PR then.

vipcxj:
@ai @wessberg 
So if I want to write a parser using typescript, where can I find the type define file?

ai:
Sorry, I have zero experience in TupeScript

PR #1098, opened 2017-12-17
use the system eol setting

DanielRuf:
this fixes the failing unit tests of postcss-cssnext on Windows

### DIFF
--- a/lib/stringifier.es6
+++ b/lib/stringifier.es6
@@ -1,12 +1,13 @@
+const endOfLine = require('os').EOL;
 const defaultRaw = {
     colon:         ': ',
     indent:        '    ',
-    beforeDecl:    '\n',
-    beforeRule:    '\n',
+    beforeDecl:    endOfLine,
+    beforeRule:    endOfLine,
     beforeOpen:    ' ',
-    beforeClose:   '\n',
-    beforeComment: '\n',
-    after:         '\n',
+    beforeClose:   endOfLine,
+    beforeComment: endOfLine,
+    after:         endOfLine,
     emptyBody:     '',
     commentLeft:   ' ',
     commentRight:  ' '


ai:
Oh, this changes could be done only in the major update. Maybe we should use some option in `toString()`?

ai:
What is the real problem for you here? Maybe I can find a solution without breaking changes?

DanielRuf:
The problem is that the testsuite of postcss-cssnext produces 3 failing tests, two of them are because of the line ending settings which are forced to `\n` by postcss (by default).

ai:
There are many ways to solve `postcss-cssnext` issue :). Changing EOL will break tests in *many* projects :(.

DanielRuf:
Hm, but the testfiles automatically get the system file endings and the generated ones use `\n` even on my Windows machine.

Or is some options value needed that the maintainer has to set in the testsuite?

ai:
> testfiles automatically get the system file endings

I think they just don’t test is on Windows. It is better to fix testfiles.

But I like you fix. I will keep it until 7.0 release (I can’t release too many major updates in the year).

DanielRuf:
I guess there is no consequent and uniform decision abroad the postcss plugins. At least there are different settings and setups (some have appveyor and travis and many do not).

DanielRuf:
> But I like you fix. I will keep it until 7.0 release (I can’t release too many major updates in the year).

So should I keep this open for the future and reference here from postcss-cssnext?

ai:
> So should I keep this open for the future

Yeap :) It is a good idea for 7.0.

gucong3000:
I ‘d like to see a software output same result on any platform

DanielRuf:
> I ‘d like to see a software output same result on any platform

Well, Windows changes the line endings so the testsuite breaks there. It is a fact that the testsuite is failing when you run it on Windows at the moment. The systems use different line endings when they generate files. So this should be solved by this PR.

Not sure why the downvote as I try to fix an actual issue and make it more independent.

gucong3000:
> Windows changes the line endings
Windows will do any thing for line endings, Maybe your `git` did it.

try this:
```bash
git config --global core.autocrlf input
```

DanielRuf:
Tests should be independent and no specific line ending be required in general.

Also setting this globally is not what the default is and as I already wrote, it is the file generation while running the testsuite, this has nothing to do with the Git config in general.

And it should not require such specific settings.

Many also include the tests in own projects but do not need a Git setup.

There is a reason for `eol` constants in many languages ;-)

And I guess @ai as owner can decide what makes sense and what not. Also this change does not negatively affect you directly. Or do you face any specific issues with the proposed change?

DanielRuf:
Mixed line endings would otherwise create other issues imo.

gucong3000:
Maybe we should keep EOL as file is. Just like IDEs and editors.

ai:
Sorry, I decided to use only `\n` on every platform because of:

1. Different new lines will make tests much more complex (some most of plugins compare output string).
2. Even Microsoft Notepad now supports UNIX new lines.

I can change my mind if you will show some good practical reasons.

And I send you a hug. I know how it is sad when somebody reject your PR. Your code is great and idea is interesting.

Issue #1066, opened 2017-09-07
Incorrect behavior when building the AST


iGetPass:
Hi.

`.Compo/* incorrect */nent`
PostCSS the output of this selector:
`.Component`

You can correct this behavior?

ai:
Yeap, it is unexpected behavior.

Quick fix: you can use `rule.raws.selector.value` to get selector before comment stripping.

hzlmn:
@ai Did you have a chance to look on it, if not i can fix it?

ai:
Did had time :(.

@hzlmn it will be awesome if you will send PR.

hzlmn:
Okay, i will try to come with something later today.

hzlmn:
@ai Quick question, should it be parsed as a single rule node with value 
equal to `"Compo/* incorrect*/sent` or it should be three separate nodes corresponding?

ai:
Good question :-/.

In the current API we don’t have nodes in `Rule#selector`. So maybe the best option here will be not trim comment without space at least from one side.

hzlmn:
@ai Seems reasonable, but to be honest I hardly think that there is a case when such selector could be useful :)

ai:
Released in 6.0.17

PR #1109, opened 2018-01-24
Fixed selector parsing with comment inside

hzlmn:
Fixes https://github.com/postcss/postcss/issues/1066 . 

As was discussed, if comment does not have at least space from one of side interp it as a part of word.

> Note,  some of `postcss-parser-tests` should be updated as them match old parsing rules.

### DIFF
--- a/lib/parser.es6
+++ b/lib/parser.es6
@@ -369,19 +369,28 @@ export default class Parser {
         let length = tokens.length;
         let value  = '';
         let clean  = true;
+        let next, prev;
+        const pattern = /^(.|#)([\w-])+$/i;
+
         for ( let i = 0; i < length; i += 1 ) {
             token = tokens[i];
             type  = token[0];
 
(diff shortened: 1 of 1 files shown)


hzlmn:
All passed except one for `:not(div/*)*/)`. What should parser produce in such case?

ai:
@hzlmn same as before `:not(div)`. Otherwise, this PR will break API.

hzlmn:
@ai So comment stripping should be performed for all selectors types except class, id or type or when it should be stripped and when not? Don't you find it a bit confusing that certain selectors will be parsed differently like if i get `.Compo/*test*/nent` as a selector value i expect  parsing`.a/*test*/[href*="test"]` to get the same.

ai:
Let's keep comments only inside identificator (letters and digits). Not between difficult like different tokens.

ai:
Released in 6.0.17

PR #1113, opened 2018-02-15
use primitive object type in TS definition file

sylvainpolletvillard:
Webstorm was complaining when I use litteral objects within PostCSS API.
According to TS guidelines, we should not use the non-primitive `Object` type : https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html


ai:
Released in 6.0.18

PR #1115, opened 2018-02-16
Improved typescript declarations for source maps

hzlmn:
`source-map` provides ts typings so would be nice to get benefits from it.


ai:
Released in 6.0.19

PR #1129, opened 2018-03-07
Update README.md

oleersoy:


ai:
Sorry. This note will make the simple example much more complicated :(.

Maybe we can change file names to avoid error at all?

PR #1143, opened 2018-04-20
tests: use npm ci

DanielRuf:
`npm ci` is much faster and is meant for CI. This PR is meant for evaluating the new ci command and testig the performance and compare the logs of Travis CI.

https://docs.npmjs.com/cli/ci
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,8 @@ node_js:
   - "8"
   - "6"
   - "4"
+before_install:
+  - npm install grunt-cli -g
 install:
   - YARN_IGNORE_ENGINES=true yarn
+  - npm ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ node_js:
   - "4"
 before_install:
   - npm install grunt-cli -g
+  - npm install npm@latest -g
 install:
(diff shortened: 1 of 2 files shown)


ai:
Hi. Thanks for PR.

But I have a question. Why we need to use `npm ci` if we already use `yarn`?

DanielRuf:
I'll check tomorrow if the changes makes much sense and how much it would improve thr builds.

ai:
Yarn is a alternative for npm. So we don't call npm at all. We use only yarn.

I really recommend to look to yarn as well. It is faster than npm, API is better and team work better.

DanielRuf:
> I really recommend to look to yarn as well. It is faster than npm, API is better and team work better.

The npm team learned from them and improved npm 5 and 6 and they are not interchangeable (lockfiles, tools, ecosystem, ...). But thanks for the recommendation (we do not prefer lockins / changes here).

ai:
So you know about the Yarn 😊

What is the point to installing npm in this case? Also what is a point in installing `grunt-cli`? PostCSS uses Gulp.

DanielRuf:
The grunt-cli seems to be by accident, sorry,

PR #1151, opened 2018-05-28
Add new syntax

gucong3000:


ai:
Wow!!! Even object-notation is supported.

You need to write about it to styled-component community.

gucong3000:
> You need to write about it to styled-component community.

https://github.com/styled-components/styled-components/issues/1763

PR #1158, opened 2018-06-19
Update Chinese doc to the newest version

wenliangdai:


ai:
Should we use Chinese Wikipedia article for Abstract Syntax Tree? Also, maybe there is a translation for Abstract Syntax Tree.

wenliangdai:
@ai I was thinking most people should understand what Abstract Syntax Tree is. But you are right, I'll make this translated now.

ai:
Thanks for quick update :+1:

PR #1170, opened 2018-07-16
Update clone and cleanRaws docs

straker:
ESlint required an `@return` for JSDoc comments, so I had `cleanRaws` return itself like some other functions for chaining.

### DIFF
--- a/lib/node.es6
+++ b/lib/node.es6
@@ -161,17 +161,17 @@ class Node {
   }
 
   /**
-   * Returns a clone of the node.
+   * Returns an exact clone of the node.
    *
-   * The resulting cloned node and its (cloned) children will have
-   * a clean parent and code style properties.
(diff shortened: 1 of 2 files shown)


ai:
Changing `return` is breaking change and I just released 7.0 a few hours ago. Let’s keep the old `return`.

straker:
Ok, np.

ai:
Also, you need to add new words to the dictionary in `yaspeller` section of `package.json`

straker:
Is there something I need to change to fix the typo errors about cleanRaws and keepBetween?

straker:
Perfect, was just about to ask about that

ai:
Great 👍 Thanks. I will release it a few days later.

ai:
Released in 7.0.1

PR #1176, opened 2018-08-16
Update TypeScript definition of LazyResult

plesiecki:
Hello @ai 

Do you mind adding `error` to ts defintion of `LazyResult`?

I find it useful for testing.

Peace :v:

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -324,6 +324,11 @@ declare namespace postcss {
      * this Result instance.
      */
     opts: ResultOptions;
+    
+    /**
+     * @returns Exists when an error occurs
+     */
+    error?: CssSyntaxError;
   }
   /**
    * Provides the result of the PostCSS transformations.


ai:
Hi. What is a `error`, can you tell me more (I didn't find property or method with this name).

plesiecki:
That's interesting. It's the same thing you can access from `then`'s second argument or from `catch`. You can see it here https://repl.it/@plesiecki/ShamelessGorgeousSystemsanalysis

ai:
Link doesn't show it. Maybe you forgot to save?

plesiecki:
Well I guess you have to click "run" at the top ;)

plesiecki:
```js
postcss().process()
```
and the result looks like this
```js
LazyResult {
  stringified: false,
  processed: false,
  error: TypeError: Cannot read property 'toString' of undefined
    at new Input (/home/runner/node_modules/postcss/lib/input.js:58:20)
    at parse (/home/runner/node_modules/postcss/lib/parse.js:16:15)
    at new LazyResult (/home/runner/node_modules/postcss/lib/lazy-result.js:70:16)
    at Processor.<anonymous> (/home/runner/node_modules/postcss/lib/processor.js:137:12)
    at Processor.process (/home/runner/node_modules/postcss/lib/processor.js:119:23)
    at evalmachine.<anonymous>:3:11
    at Script.runInContext (vm.js:74:29)
    at Object.runInContext (vm.js:182:6)
    at evaluate (/run_dir/repl.js:133:14)
    at ReadStream.<anonymous> (/run_dir/repl.js:116:5),
  result: 
   Result {
     processor: Processor { version: '7.0.2', plugins: [] },
     messages: [],
     root: undefined,
     opts: {},
     css: undefined,
     map: undefined } }   
```

ai:
Found it. Hm, it is a private API. It could be changed in any patch release. 

Show me your use cases, maybe we will move it to public API.

plesiecki:
It looks like something along those lines:
```js
test('should error', async t => {
    const err = await postcss(plugins).process(brokenCSS).catch(err => err)

    // assertions on err
})
```
Right now to access an error I pass it from `catch`.

Well I can live with that ;)

What do you think?

plesiecki:
Forgot to mention in typescript right now I can't do this this way:
```js
test('should error', async t => {
    const error = await postcss(plugins).process(brokenCSS)

    // assertions on error
})
```
because of the current state of d.ts

plesiecki:
Thanks for immediate response :heart: 

I rewrote my test and I noticed another interesting thing.
In latest ava@1.0.0-beta.7 it goes like this:
```js
const {message} = await t.throwsAsync(postcss(plugins).process(cssWithSyntaxError))
```
Now AVA complains about `CssSyntaxError`:
> Promise rejected with exception that is not an error

AVA uses `is-error` module to test reason of promise rejection.

What do you think about extending from `Error` in `CssSyntaxError`?

Should I open separate issue for the above?
Feel free to close this PR.

ai:
> What do you think about extending from Error in CssSyntaxError?

Really good idea. Send PR.

Issue #1178, opened 2018-08-30
IE11 support?


Pomax:
I noticed the changelog for v6 -> v7 includes ["Remove IE and “dead” browsers from Babel."](https://github.com/postcss/postcss/commit/85c33b9d9a8a10a1b7ac2afee84a64d8eb5e05c1) but IE11 is (as much as I would _really_ want it to be) most emphatically not a dead browser. It is [the only version of IE that is still supported by Microsoft](https://support.microsoft.com/en-ca/help/17454/lifecycle-faq-internet-explorer), but thanks to it shipping with Windows 10, it is literally going to be an actively supported browser until Windows 10's EOL, which is 13 October 2020.

Looking at http://browserl.ist/, IE11 still has more market share than Safari and Edge combined, so even in terms of real world use, it's still relevant. Can IE11 support be restored? As much as we all hate it? =(

ai:
@Pomax we removed IE support from very rare case running PostCSS in the browser.

Most of the users don’t do it (PostCSS compile CSS during deploy in Node.js).

You need to run PostCSS in browser only for rare cases. For example, online tools like CodePen (but even CodePen run PostCSS on the servers by AJAX) or [Autoprefixer online](http://autoprefixer.github.io/). These online tools are used only by web developers and it is extremely unlikely that they will use IE.

ai:
If you know the popular case of running PostCSS in the browser with many IE users, feel free to open the issue again

Pomax:
ahh, the changelog and commit were not clear on that. Thanks.

ai:
@Pomax Yeap 😣 maybe you can send PR to fix ChangeLog text. I am not native speaker, it is hard to me.

Pomax:
sure: https://github.com/postcss/postcss/pull/1181

PR #1181, opened 2018-09-01
Update CHANGELOG.md

Pomax:
small rephrase to make sure its clear the ie/dead browser removal only applies to the client-side postcss babel step, not postcss "in general" (as per https://github.com/postcss/postcss/issues/1178#issuecomment-417553553)


ai:
Thanks

PR #1191, opened 2018-09-25
feat: add ignoreUnclosed option to nextToken method

shellscape:
This PR stems from [an issue in postcss-less](https://github.com/shellscape/postcss-less/issues/117) and a [short discussion on twitter](https://twitter.com/sitnikcode/status/1044319395760263168) with @ai.

### Use Case

PostCSS v7 has a wonderful parser class structure that enabled syntaxes to be created by inheriting from the base `Parser` class and selectively overriding methods to handle syntax nuances - all _without having to reimplement a parser or tokenizer._ That's a **massive** advantage to developers. However, there are some strict rules hardcoded into the tokenizer. Because the tokenizer is optimized for performance, it cannot be made a class to be inherited from. A mechanism is then needed to work around/with the hardcoded tokenizer rules.

In some situations it would be preferable to fetch the next token from a tokenizer while ignoring any `unclosed` errors for only the next token request. This is different than instantiating the tokenizer with the `ignoreErrors` options, as it only ignores `unclosed` for a single `tokenizer.nextToken()` request. 

This PR implements an `ignoreUnclosed` option. The naming of that option was chosen because that option is _only_ used on lines where the `else` cases result in the `unclosed(...)` method being called, and an "Unclosed X" error is thrown. Hence the naming is more specific, but should aid in usability and providing a distinction between the top-level tokenizer option `ignoreErrors` and this new `nextToken` method option.

Example Usage:

```js
// default value for `nextToken` opts is `{}`
// default value for ignoreUnclosed is `false`
this.tokenizer.nextToken({ ignoreUnclosed: true });
```

### DIFF
--- a/lib/tokenize.es6
+++ b/lib/tokenize.es6
@@ -45,9 +45,12 @@ export default function tokenizer (input, options = {}) {
     return returned.length === 0 && pos >= length
   }
 
-  function nextToken () {
+  function nextToken (opts = {}) {
     if (returned.length) return returned.pop()
     if (pos >= length) return
+
+    let defaults = { ignoreUnclosed: false }
+    opts = Object.assign({}, defaults, opts)
 
     code = css.charCodeAt(pos)
     if (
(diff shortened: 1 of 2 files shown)


shellscape:
🎉 thank you!

ai:
Released in 7.0.3

PR #1196, opened 2018-10-02
Use files field in package.json and remove .npmignore

pgilad:
This way discards a lot of unnecessary files bundled by the current `.npmignore` solution.


ai:
You deleted documentation, which we put to npm package for a purpose (to be able to develop PostCSS plugins without Internet).

ai:
I prefer `.npmignore` in all my projects. If I will have `files` in one of the projects, but still have `.npmignore` in other projects it will just increase mistakes. Especially, when `files`/`npmignore` has a difference — npm package size will be the same.

This is why I close PR. It will make maintenance harder without getting any practical result.

pgilad:
Sorry about the documentation flip, my mistake, but fixed. Anyway, there are currently lots of redundant files being bundled with `postcss`, that's why I noticed it and decided to go with a `whitelist` instead of `blacklist`..

This is what's currently bundled with `postcss`:
image

Anyway I guess having a single method of working (i.e npmignore or files) in all projects is a worthwhile argument, although I think this is a public enough of a repo to distinguish it from "your" projects. Thanks for the feedback

ai:
Oops, `gulpfile.js` is wrong.

Did you check the actual npm package? All other files should be not in npm package.

I release npm package from `build/` and have special scripts to cut unnecessary files. I even cut unnecessary section in `package.json`.

ai:
I removed `gulpfile.js` from npm package 893b4d0

pgilad:
image

pgilad:
👍

pgilad:
@ai I think you added it to the wrong place, it should be added to `.npmignore` file

ai:
@pgilad nope, PostCSS release process is very different from other npm packages.

```sh
npx gulp build
cd ./build/
npm publish
```

This is why I files `gulp build` script to avoid `gulpfile.js` being copied to `build/`.

pgilad:
🙈

ai:
Released in 7.0.5.

PR #1197, opened 2018-10-02
Add prettier for auto-code formatting

pgilad:
Added Prettier, this required a minor tweak to eslint rules (`space-before-function-paren` and `template-curly-spacing`, both minor spacing rules). 
Sorry for the massive changes, the only relevant files that have changed (outside of `yarn prettier`) are:

- `package.json`: Added `prettier` and `prettier-verify` scripts. Also added them to `lint-staged`. Also updated eslint config. Also run `prettier-verify` in `test` script in order to fail the build in travis.
- `.prettierrc.yml`: prettier config
- `.prettierignore`: This file is redundant, because we `prettier` only relevant files (with glob), but just in case

Hopefully this PR can be considered, code styling is a solved problem, and even though prettier isn't perfect, I think the alternative of forcing users to manually control/edit style (even though some may have relevant lint rules) is just not worth their time.


ai:
I do not like Prettier. Tools help people. So we should think about “do people need a tool” instead of using trends.

If somebody who writes PostCSS code a lot will need a Prettier, I will be open for adding it. Right now only I am writing a code. And it is not a problem for me to make code style decisions.

pgilad:
Thanks I'll see about your suggested  contributions

PR #1203, opened 2018-11-10
Introduce Input.hasBom property

vkrol:


ai:
Do we need to keep BOM in `Root#toString`? Or you are OK with cutting it?

What is a reason why you taking care of BOM? (to find the best option)

vkrol:
> Or you are OK with cutting it?

I think so 🤔

> What is a reason why you taking care of BOM? (to find the best option)

I want to create the unicode-bpm stylelint rule 
https://github.com/stylelint/stylelint/issues/3792.

ai:
OK. Seems good. I am right now on vacation. Will try to accept it during a few weeks.

ai:
Released in 7.0.6

vkrol:
@ai thanks!

Issue #1175, opened 2018-08-11
Differentiate variable from declaration?


Siilwyn:
Wondering if it would be a good idea to differentiate between a `Declaration` and a variable in CSS.
E.g.: `--bg-color: blue;` would return the `Variable` instead of `Declaration`.

So instead of having to check the declaration property for `--` it would be provided by the parser.

Context:
- https://github.com/AleshaOleg/postcss-sass/issues/54
- https://github.com/Siilwyn/css-declaration-sorter/pull/51/

ai:
It will be breaking change. But I am OK, if you will find a way without breaking changes.

hzlmn:
One possible backward compatible solution is to add extra property to declaration for variables like with !important.

ai:
Done f3df413

https://postcss.org/api/#declaration-variable

PR #1210, opened 2018-12-12
Added extra variable param to decl

hzlmn:
Simplest patch for #1175


hzlmn:
1. filtering variables from declarations. it mostly what done in most cases on user side.
2. tests will be provided if PR make sense for you and `postcss-parser-tests` should be updated as well.
3. why not? yes they are not exactly the same but i don't see possible problems with it.

ai:
> yes they are not exactly the same but i don't see possible problems with it.

For instance, Custom Properties should not be compiled to values anymore (like we do with `postcss-simple-vars`).

Issue #1213, opened 2019-01-04
[TypeScript] TransformCallback's Result argument as non-optional


aoberoi:
Hello there!

I noticed that the `TransformCallback` (return value from `postcss.plugin()`) in the type definitions receives an optional `result` argument.

https://github.com/postcss/postcss/blob/3a44f17aadb81f984e760d05cceac2e6e5b6471d/lib/postcss.d.ts#L36

I cannot think of a reason why `result` would be undefined. But leaving its type this way forces my plugin to deal with the possibility. I have to add an ugly block of code near the top of my transform callback that looks like this, just so that that I don't have to check for `undefined` later:

```javascript
    if (result === undefined) {
      throw new Error('Cannot run without a result defined');
    }
```

I propose that we change the definition to the following:

```typescript
interface TransformCallback {
    /**
     * @returns Asynchronous plugins should return a promise.
     */
     (root: Root, result: Result): Promise<any> | any;
  }
```

This should not break existing TypeScript implementations since a function which doesn't use the `result` argument would still implement this interface. You'll also notice that the return value is changed, which also shouldn't break anything, since both the existing return type and the one I've proposed uses a union with `any`, which is effectively still `any`. I just think this reads clearer, since asynchronous plugins should return a Promise, and otherwise the return value doesn't really matter.

If this sounds good to the maintainers, I'm happy to make the PR.

cc @jedmao (seems to be the resident typescript expert)

jedmao:
You're right that a callback function shouldn't have optional args, as they are all passed regardless. You just opt into the ones you want by only asking for `n` number of args.

As for `void | Function | any`. Yeah, that's really just the same as `any`, but I wrote the union to be more explicit. If you look at the [`postcss.plugin`](http://api.postcss.org/postcss.html#.plugin) docs, you'll see two examples of return types: `void | Promise<void>`, but technically there's no reason you couldn't return other things too. Probably should have written this as `void | Promise<void>` without the `any` in the first place so people don't get the wrong idea. If they return anything else it's not like we're going to do anything with that information. And now here we are, wondering if anyone's relying on that return type for some reason? So it does have the potential to break some plugins if we remove `any` from that union, in that they won't compile w/o changing the return type to suit.

In conclusion, yes, we can make the `result` non-optional and `void | Promise<void> | any` for explicitness and backwards compatibility.

aoberoi:
Thanks for the fast feedback @jedmao.

[...]

PR #1217, opened 2019-01-05
Updates Node#source to optional type, adds comments

### LINKED ISSUE #1214: [TypeScript] node.source (and maybe its properties) should be of optional types
The comments around the declaration themselves state that this value might not be defined:

https://github.com/postcss/postcss/blob/3a44f17aadb81f984e760d05cceac2e6e5b6471d/lib/postcss.d.ts#L616-L625

In both the other places where the type `NodeSource` is used (in `JsonNode` and `NodeNewProps`), its optional. I believe it should be in `NodeBase` as well.

---

I'm curious about whether the `input` property inside `NodeSource` should be optional. Also, since the `file` property within `Input` is only conditionally initialized, should that also be optional?

aoberoi:
the type definitions are updated to reflect where properties are optional. the comments are used to clarify the conditions in which those properties will have values, or not.

fixes #1214

i also took this opportunity to improve on #1215's work by adding a clarifying comment, as mentioned at the end of [this comment](https://github.com/postcss/postcss/issues/1213#issuecomment-451367502)

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -31,7 +31,10 @@ declare namespace postcss {
   }
   interface TransformCallback {
     /**
-     * @returns Asynchronous plugins should return a promise.
+     * @returns Plugins should return a Promise that resolves when all modifications
+     * to work is complete. They may return synchronously, but that style of plugin is
+     * only meant for debugging and development. In either case, the resolved or
+     * returned value is not used - the "result" is the output.
      */
     (root: Root, result: Result): Promise<any> | any;
   }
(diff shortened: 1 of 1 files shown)


aoberoi:
cc @jedmao

jedmao:
@aoberoi I don't know how passionate you are about TypeScript, but you might be interested in https://github.com/jedmao/postcss/tree/typescript. It's an old branch that's now pretty far behind current PostCSS master, but I had the entire PostCSS library converted over into TS, including tests. That's where the original `d.ts` file was generated from. It'd be super awesome of someone were to update it to work with the latest version of TS.

ai:
> Are we willing to convert PostCSS into a TypeScript project yet?

Nope 😄 I found that pure JS projects is easier to maintain. I will be glad to remove Babel if I will have time.

aoberoi:
@ai totally fair.

Still, ya gotta love the irony when the creator of a source code processing tool tell you he wants to get rid of source code processing tools in the project. :smile:

PR #1219, opened 2019-01-07
Updates type definitions for ProcessOptions

aoberoi:
It seems that quite a few of the type definitions related to `ProcessOptions` were not in sync with the latest source. This change should bring it back up to date.

Notably:
*  `ProcessOptions` no longer extends `Syntax`. This used to be a way of saying the object had both `parse` and `stringify` options, but that wasn't true. The object does have `parser` and `stringifier` options.
*  The `safe` option was removed. It's not referenced in the [documentation](http://api.postcss.org/global.html#processOptions).
*  `Parse` and `Stringify` are renamed to `Parser` and `Stringifier`. This is consistent with the documentation, and reflects the fact that these are functions (verbs instead of nouns).
*  `Parser` options has `from` and `map` properties, instead of taking `SourceMapOptions` (which is the type for the `map` option). Fun use of the `Pick<T,K>` helper!
*  Fixes the argument optionality of `Parser` and `Stringifier`. Adds `Builder` to specify the argument that use to be an `any`.
*  General clean up of comments and types.

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -244,7 +244,7 @@ declare namespace postcss {
     /**
      * Source map options
      */
-    map?: SourceMapOptions;
+    map?: SourceMapOptions | true;
   }
   interface Syntax {
     /**
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -181,13 +181,11 @@ declare namespace postcss {
      *
      * If you have set inline: true, annotation cannot be disabled.
      */
-    annotation?: boolean | string;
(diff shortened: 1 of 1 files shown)


aoberoi:
cc @jedmao

ai:
@aoberoi tell me when you will finish fixes. I will release them.

aoberoi:
@ai I think those are all the changes I’ll have for a while. Feel free to release :smile:

ai:
@aoberoi released in 7.0.8

PR #1220, opened 2019-01-07
Added check to see if css is defined

adover:


lib/input.es6
```diff
@@ -18,7 +18,7 @@ class Input {
    * @param {object} [opts] {@link Processor#process} options.
    */
   constructor (css, opts = { }) {
-    if (css === null || (typeof css === 'object' && !css.toString)) {
+    if (!css || css === null || (typeof css === 'object' && !css.toString)) {
```
aoberoi: if we use this change as-is, `css` being an empty string would error. that doesn't seem to be what you want, given the title. if you want to check if the value is defined, you should probably use `css === undefined` instead.

if you do actually want the empty string to generate an error, then i suggest you remove the `css === null` part of this condition, because `!css` would already be true for that case.

adover:
@ai @aoberoi thanks for the feedback. I noticed this once I broke other things :) I'll raise an issue to explain more the issue I'm trying to fix

PR #1223, opened 2019-01-11
Fix end tokens having wrong sourcemap entry

### LINKED ISSUE #1222: Sourcemaps contain invalid mappings (even without any plugins)
PostCSS seems to generate or emit sourcemaps with mappings to non-existent positions. [In this visualisation](http://sokra.github.io/source-map-visualization/#base64,CmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9Cgo=,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInNyYy9zdHlsZS5jc3MiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IjtBQUNBO0NBQ0Msc0JBQXNCO0NBQ3RCIiwiZmlsZSI6Im91dC5jc3MiLCJzb3VyY2VzQ29udGVudCI6WyJcbmJvZHkge1xuXHRiYWNrZ3JvdW5kLWNvbG9yOiByZWQ7XG59XG5cbiJdfQ==,CmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9Cgo=), you can see that there is a mapping from/to line 3 col 23. But that line has only 23 columns (which are 0-indexed). 
This could be a off-by-one error or caused by incorrect usage of the source-map library.

I have some issues with sourcemaps in https://github.com/parcel-bundler/parcel, because when concatenating multiple css files, the last position from one CSS asset would then actually point to a different CSS asset. Chrome seems to handle them fine, but testing is becomes harder.

Input file:
```css

body {
	background-color: red;
}

```

Generated map:
```json
{"version":3,"sources":["src/style.css"],"names":[],"mappings":";AACA;CACC,sBAAsB;CACtB","file":"out.css","sourcesContent":["\nbody {\n\tbackground-color: red;\n}\n\n"]}
```

Code:
```js
const fs = require("fs").promises;
const postcss = require("postcss");

(async () => {
	const inputFile = "src/style.css";
	const contents = await fs.readFile(inputFile);

	const root = postcss.parse(contents, { from: inputFile });

	const result = root.toResult({
		to: "out.css",
		map: { inline: false, annotation: false, sourcesContent: true }
	});

	let { css, map } = result;
	console.log(map.toJSON());
	map = map.toString();

	await fs.writeFile("out.css", css);
	await fs.writeFile("out.css.map", map);
})();
```

mischnic:
There was a bug where end node parts (not sure how they're actually called) like ";" or "}" had off-by-one-errors (explanation in separate comments)

[Before](http://sokra.github.io/source-map-visualization/#base64,CmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9Cgo=,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInNyYy9zdHlsZS5jc3MiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IjtBQUNBO0NBQ0Msc0JBQXNCO0NBQ3RCIiwiZmlsZSI6Im91dC5jc3MiLCJzb3VyY2VzQ29udGVudCI6WyJcbmJvZHkge1xuXHRiYWNrZ3JvdW5kLWNvbG9yOiByZWQ7XG59XG5cbiJdfQ==,CmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9Cgo=)
[After](http://sokra.github.io/source-map-visualization/#base64,CmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9Cgo=,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInNyYy9zdHlsZS5jc3MiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IjtBQUNBO0NBQ0MscUJBQXFCO0FBQ3RCIiwiZmlsZSI6Im91dC5jc3MiLCJzb3VyY2VzQ29udGVudCI6WyJcbmJvZHkge1xuXHRiYWNrZ3JvdW5kLWNvbG9yOiByZWQ7XG59XG5cbiJdfQ==,CmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9Cgo=)

Fixes #1222

### DIFF
--- a/lib/map-generator.es6
+++ b/lib/map-generator.es6
@@ -231,10 +231,10 @@ class MapGenerator {
         if (node.source && node.source.end) {
           this.map.addMapping({
             source: this.sourcePath(node),
-            generated: { line, column: column - 1 },
+            generated: { line, column: column - 2 },
             original: {
               line: node.source.end.line,
-              column: node.source.end.column
+              column: node.source.end.column - 1
             }
           })
         } else {
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -23,6 +23,6 @@ it('generates source map without fs', () => {
   expect(postcss([() => true]).process('a{}', { from: 'a.css', map: true }).css)
     .toEqual('a{}\n/*# sourceMappingURL=data:application/json;base64,' +
              'eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbImEuY3NzIl0sIm5hbWVzIjpbXSw' +
-             'ibWFwcGluZ3MiOiJBQUFBLEdBQUciLCJmaWxlIjoiYS5jc3MiLCJzb3VyY2' +
+             'ibWFwcGluZ3MiOiJBQUFBLEVBQUUiLCJmaWxlIjoiYS5jc3MiLCJzb3VyY2' +
              'VzQ29udGVudCI6WyJhe30iXX0= */')
 })
--- a/test/map.test.js
+++ b/test/map.test.js
@@ -66,6 +66,12 @@ it('generate right source map', () => {
     column: 0,
     name: null
   })
+  expect(map.originalPositionFor({ line: 1, column: 2 })).toEqual({
+    source: 'a.css',
+    line: 1,
+    column: 0,
+    name: null
+  })
   expect(map.originalPositionFor({ line: 2, column: 2 })).toEqual({
     source: 'a.css',
     line: 2,
@@ -75,6 +81,18 @@ it('generate right source map', () => {
   expect(map.originalPositionFor({ line: 3, column: 2 })).toEqual({
     source: 'a.css',
     line: 2,
+    column: 2,
+    name: null
+  })
+  expect(map.originalPositionFor({ line: 3, column: 14 })).toEqual({
+    source: 'a.css',
+    line: 2,
+    column: 14,
+    name: null
+  })
+  expect(map.originalPositionFor({ line: 4, column: 2 })).toEqual({
+    source: 'a.css',
+    line: 3,
     column: 2,
     name: null
   })


lib/map-generator.es6
```diff
@@ -231,10 +231,10 @@ class MapGenerator {
         if (node.source && node.source.end) {
           this.map.addMapping({
             source: this.sourcePath(node),
-            generated: { line, column: column - 1 },
+            generated: { line, column: column - 2 },
```
mischnic: `column` is the 1-based position in the generated css. But `column += str.length` happens before the end tag is mapped (now, column is the first column index of the next node), therefore the end token is at `column - 1 - 1` (if it has length 1).

I'm not 100% sure this is always correct (but should be if all "end tokens" are of length 1).

ai:
Wow, it was fast

ai:
Released in 7.0.10.

mischnic:
> Released in 7.0.10.

Thank you for your quick response.
But there I discovered an issue: the last semicolon might be missing, so [this](http://sokra.github.io/source-map-visualization/#base64,Cgpib2R5IHsKCWNvbG9yOiByZWQ7CgliYWNrZ3JvdW5kLWNvbG9yOiByZWQKfQoK,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInNyYy9zdHlsZS5jc3MiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6Ijs7QUFFQTtDQUNDLFVBQVU7Q0FDVixvQkFBb0I7QUFDckIiLCJmaWxlIjoib3V0LmNzcyIsInNvdXJjZXNDb250ZW50IjpbIlxuXG5ib2R5IHtcblx0Y29sb3I6IHJlZDtcblx0YmFja2dyb3VuZC1jb2xvcjogcmVkXG59XG5cbiJdfQ==,Cgpib2R5IHsKCWNvbG9yOiByZWQ7CgliYWNrZ3JvdW5kLWNvbG9yOiByZWQKfQoK) could happen (the last `d` of `red` is separate).
Could there be other cases where something similar might happen?
This would the the semicolon:
```diff
diff --git a/lib/map-generator.es6 b/lib/map-generator.es6
index 8029cbf..dc13536 100644
--- a/lib/map-generator.es6
+++ b/lib/map-generator.es6
@@ -227,7 +227,8 @@ class MapGenerator {
         column += str.length
       }

-      if (node && type !== 'start') {
+      if (node && type !== 'start' &&
+           (node === node.parent.last ? node.parent.raws.semi : true)) {
         if (node.source && node.source.end) {
           this.map.addMapping({
             source: this.sourcePath(node),
```

ai:
Send another PR, I will release 7.0.11.

Yeap, semicolon can be missed.

PR #1230, opened 2019-01-17
Update README.md

JustClear:


ai:
Hi. You plugin is awesome, but this list is only for popular or non-standard plugins.

But there is a another list for all plugins. Add it to `docs/plugins.md` in another PR.

JustClear:
Oh... Thanks

PR #1245, opened 2019-02-22
Visitor to postcss

BondarenkoAlex:


test/parse.test.js
```diff
@@ -1,5 +1,5 @@
 let parse = require('../lib/parse')
-let Root = require('../lib/root')
+let Root = require('../lib/root').default
```
ai: This is breaking changes. We need to avoid it.

lib/lazy-result.es6
```diff
@@ -3,6 +3,13 @@ import stringify from './stringify'
 import warnOnce from './warn-once'
 import Result from './result'
 import parse from './parse'
+import {
+  isVisitorMode,
+  listeners,
+  isComplete,
+  isClean,
+  walkVisitor
+} from './root'
```
ai: It will be better to have separated files for this helps

test/parse.test.js
BondarenkoAlex: Ok

lib/lazy-result.es6
BondarenkoAlex: Ok

jonathantneal:
I’m so excited for this. Question: Observers will default to node `enter`, but I have thought Babel defaults observers to `exit`. Would someone verify this?

BondarenkoAlex:
> I’m so excited for this. Question: In this PR, observers will default to node `enter`, but I have thought Babel observers default to `exit`. I think we should be consistent. Would someone verify this?

The default value is "enter".
>Note: Identifier() { ... } is shorthand for Identifier: { enter() { ... } }
https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-visitors

jonathantneal:
Consider my feedback resolved. This is going to have such an amazing impact on performance, @BondarenkoAlex.

lib/node.es6
```diff
@@ -419,6 +423,22 @@ class Node {
     return pos
   }
 
+  [resetNodeWalk] () {
```
ai: Давай вынесём её в функцию.

Методы плохо работают в JS (например, плохо сжимаются):

```
function resetNodeWalk (node) {
    if (node.root().isVisitorMode === false) {
      return
    }

    node[isClean] = false

    if (node[isComplete]) {
      node[isComplete] = false

      if (node.parent) {
        resetNodeWalk(node.parent)
      }
    }
  }
}
```

BondarenkoAlex: Ок. Нет проблем.

lib/lazy-result.es6
```diff
@@ -375,6 +386,33 @@ class LazyResult {
...
+    }
+
+    root[isVisitorMode] = true
+    while (!root[isClean]) {
+      root[isClean] = true
+      root[walkVisitor]((node, index, isPostOrder) => {
+        let { type } = node
+        let visitorsByType = plugins[type] || {}
+        let order = !isPostOrder ? 'enter' : 'exit'
+        let visitorsByOrder = visitorsByType[order] || []
```
CyberAP: End of the statement is not reachable
```suggestion
        let visitorsByOrder = visitorsByType[order]
```

lib/root.es6
```diff
@@ -1,4 +1,69 @@
...
+  // typeNode have view "decl" or "decl.exit" or "decl.enter"
+  // return { decl: {enter: cb}}
+  let type = typeNode
+  if (!type.includes('.')) {
+    type = `${ type }.enter`
+  }
+
+  type = type.toLowerCase()
+
+  let arr = type.split('.')
```
CyberAP: A little more readability for computed properties.
```suggestion
  let [decl, phase] = type.split('.')
  return ({
    [decl]: {
      [phase]: callback
    }
  })
```

lib/root.es6
```diff
@@ -1,4 +1,69 @@
...
+  }
+}
+
+/* General view of node type names */
+function normalizeVisitorPlugin (typeNode, callback = () => {}) {
+  // typeNode have view "decl" or "decl.exit" or "decl.enter"
+  // return { decl: {enter: cb}}
+  let type = typeNode
+  if (!type.includes('.')) {
+    type = `${ type }.enter`
```
CyberAP: Any need for template literals here?
```suggestion
    type += '.enter'
```

lib/root.es6
```diff
@@ -1,4 +1,69 @@
...
+}
+
+function validateNameTypeNode (typeNode) {
+  let type = typeNode
+
+  if (!isString(type)) {
+    throw new Error('typeNode must be a string')
+  }
+
+  let arr = type.split('.')
```
CyberAP: Could be significantly simplified without nested ifs.

```suggestion
  let arr = type.split('.')
  let [_, phase] = arr
  let validPhases = ['enter', 'exit']
  
  if (!phase || !validPhases.contains(phase)) {
    throw new Error('The plugin must subscribe to either the enter or exit node')
  }
```

CyberAP:
Is it absolutely required to mark the whole tree as dirty every time a change happens and iterate all over again?
https://github.com/postcss/postcss/blob/019c5c18a7306fe3aec3635e0eea8e9188779df7/lib/reset-node-walk.es6#L12-L18
Could it be possible to trigger listeners of that specific node type just in place if the property changes?  (`resetNodeWalk` is likely a place for it)

Could it have race conditions in that case?

lib/root.es6
BondarenkoAlex: Oh, Yes.

lib/root.es6
CyberAP: Made a typo here, should be `includes` instead of `contains`. @BondarenkoAlex

lib/root.es6
BondarenkoAlex: Readability. Imho.

Issue #1249, opened 2019-03-14
postss-bem version is too low


holidaypenguin:
# 中文
在 [language-extensions](https://github.com/postcss/postcss/blob/master/docs/plugins.md#language-extensions) 中的`postcss-bem` 版本过低，建议用[postcss-bem-fix](https://github.com/supermonkeyz/postcss-bem-fix)替换他的位置

# English
The version of `postcss-bem` in [language-extensions](https://github.com/postcss/postcss/blob/master/docs/plugins.md#language-extensions)  is too low. It is recommended to replace his location with [postcss-bem-fix](https://github.com/supermonkeyz/postcss-bem-fix).

ai:
@holidaypenguin sure. Do you want to send PR?

holidaypenguin:
Can I do it?

ai:
@holidaypenguin yeap! Press the Edit button for this file on GitHub and send a pull request. Let's save your name in the project history.

PR #1250, opened 2019-03-15
Fix broken link (#1249)

holidaypenguin:
Update postcss-bem to postcss-bem-fix to support the latest version of postcss.


ai:
You need also remove postcss-bem link from links list

holidaypenguin:
@ai Done

PR #1258, opened 2019-04-25
fix: Remove docs folder from package

retyui:
via: https://github.com/postcss/postcss/issues/1257


ai:
`docs/` are already in ignore. Try `npm pack` on initial `ose` branch. Maybe we already removed docs from npm package.

PR #1262, opened 2019-05-06
Change the engine of documentation

isolovev:


package.json
```diff
@@ -52,13 +53,14 @@
...
     "jest": "^24.7.1",
-    "jsdoc": "^3.5.5",
     "lint-staged": "^8.1.6",
     "postcss-parser-tests": "^6.5.0",
     "rewire": "^4.0.1",
     "rimraf": "^2.6.3",
     "size-limit": "^1.3.1",
+    "stream-array": "^1.1.2",
     "strip-ansi": "^5.2.0",
+    "vinyl-fs": "^3.0.3",
```
ai: Can we do the task without extra dependencies?

package.json
```diff
@@ -363,7 +365,14 @@
       "Mischkulnig",
       "Tidelift",
       "Solovev",
```
ai: Should we remove something from dictionary since we do not use noisy JSDoc template?

package.json
isolovev: I'll try

package.json
isolovev: It is in the CHANGELOG.md

isolovev: I'll remove it

docs.js
```diff
@@ -0,0 +1,37 @@
+let documentation = require('documentation')
+let streamArray = require('stream-array')
+let vfs = require('vinyl-fs')
+
+let generateDocs = require('./helpers/generateDocs')
```
ai: Documentation is a small task, it is not cool to have so many separated files for it.

You can move everything to `generate-docs.js` or create `docs/` and move everything there.

helpers/constants.js
```diff
@@ -0,0 +1,7 @@
+let { join } = require('path')
+
+/**
+ * API folder
+ * @type {string}
+ */
+exports.API_FOLDER = join(process.cwd(), 'api')
```
ai: It, definitely, should not be in separated file

.jsdocrc
```diff
@@ -1,13 +0,0 @@
-{
```
ai: Should you clean `.npmignore` from `.jsdoc` (try to find any `.jsdoc` mentions in the project to be sure that you cleaned everything).

Do not forget to add new files to `.npmignore` to keep PostCSS npm package small.

package.json
ai: Yeap, you do not need to remove this specific word, just to clean a dictionary from other words.

package.json
isolovev: I removed `stream-array`. Vinyl-fs is very large. Is it worth adding at the main code?

ai:
You resolved some conversations without fixing the issue (like comments about updating `.npmignore`).

It could force you to forget something. It is a best practice to resolve conversation only after you pushed a fix.

package.json
ai: The real question here is what is a reason for this dependencies. If it is a small syntax sugar to avoid a few LOC, we need to remove them.

If it avoids us from having 2-3 big files, in this case, it is better to keep them.

docs.js
ai: I restored the conversation since we still have `helpers/` and `generate-doc.js` files.

isolovev:
I'm sorry. I've not added the fix.

package.json
isolovev: It avoids having 5 big files

.jsdocrc
isolovev: I added `generate-docs.js` at `. npmignore`

docs.js
isolovev: I added at single file

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+      renderError(error)
+    }
+
+    let html = data
+      .replace(
+        /regular blockmt1 quiet rounded/g,
+        'blockmt1 quiet rounded bold block h4 mt2'
+      )
+
+    fs.writeFile(indexFile, html, errorWrite => {
```
ai: Since we support Node > 6, you can use `let writeFile = util.promisify(fs.writeFile)`. As result we will be able to use `async/await`.

package.json
ai: Why do we need VFS and streams for this task?

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+
+class StreamArray extends Readable {
+  /**
+   * Create a new instance of StreamArray
+   *
+   * @access private
+   * @param {Array} list
+   */
+  constructor (list) {
+    if (!Array.isArray(list)) {
```
ai: Do we need this check?

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+ * @param {String} key Key of object
+ * @return {*}
+ */
+function concatValue (obj, key) {
+  let isEnum = obj.propertyIsEnumerable(key)
+  return typeof key === 'string' && isEnum ? [obj[key]] : []
+}
+
+class StreamArray extends Readable {
+  /**
```
ai: We are using JSDoc only for public API. I think we can trim them.

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+    path: []
+  }
+}
+
+/**
+ * Returns an array of values of the enumerable properties of an object
+ * @param {Object} obj Object that contains the properties and methods
+ * @return {Array}
+ */
+function objectValues (obj) {
```
ai: Native `Object.assign` is available in Node.js > 6

generate-docs.js
```diff
@@ -0,0 +1,232 @@
+let fs = require('fs')
+let { join } = require('path')
+let { Readable } = require('stream')
+let documentation = require('documentation')
+let vfs = require('vinyl-fs')
+
+let API_FOLDER = join(process.cwd(), 'api')
+
+let config = {
```
ai: It will be more readable to move config directly to `build()` call. Anyway it is very small.

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+  extension: 'js',
```
ai: It is very strange that it is not default value. Are you sure that we need it?

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+  shallow: true,
+  github: true
+}
+
+documentation
```
ai: Good candidate to use async/await

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+  return new StreamArray(output)
+    .pipe(vfs.dest(API_FOLDER))
+    .on('end', addMarginForNote)
+}
+
+/**
+ * Render error
+ * @param {Error} error
+ */
+function renderError (error) {
```
ai: It is better to keep error processing in one place. If you will use Promise in fs method, you can put this code into single `catch` statement without separated function

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+          let instanceEssence = essence.members.instance
+            .filter(item => item.memberof === essence.name)
+
+          doc.members.instance = []
+            .concat(doc.members.instance, instanceEssence)
+            .map(item => Object.assign(item, {
+              namespace: item.namespace.replace(essence.name, doc.name)
+            }))
+            .filter((item, index, list) =>
+              item.memberof === essence.name ||
```
ai: This statement is too complicated. It is better to rewrite it to `=> { }` multiline function

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+        let essence = docs
+          .find(essenceItem => essenceItem.name === augment.name)
+
+        if (essence) {
+          let instanceEssence = essence.members.instance
+            .filter(item => item.memberof === essence.name)
+
+          doc.members.instance = []
+            .concat(doc.members.instance, instanceEssence)
+            .map(item => Object.assign(item, {
```
ai: Can we use object destruction here since we need only latest Node 10-12?

generate-docs.js
```diff
@@ -0,0 +1,232 @@
...
+                )
+              )
+            )
+            .sort((a, b) => a.name > b.name ? 1 : -1)
+        }
+      })
+    })
+
+  docs
+    .sort((a, b) => {
```
ai: Why do we need to sort array of we do not have the order?

generate-docs.js
ai: Fix: Node >= 10

generate-docs.js
isolovev: Done

generate-docs.js
isolovev: Done

generate-docs.js
isolovev: Done

generate-docs.js
isolovev: It is done for sorting documentation at alphabet

generate-docs.js
isolovev: I removed

generate-docs.js
isolovev: I done

generate-docs.js
isolovev: I fixed

generate-docs.js
ai: Oh, I forgot that `sort` is mutable.

generate-docs.js
isolovev: I deleted the big part but saved for autocomplete

generate-docs.js
isolovev: I removed

generate-docs.js
isolovev: I added

generate-docs.js
```diff
@@ -0,0 +1,167 @@
+let fs = require('fs')
+let path = require('path')
```
ai: Let’s import only necessary methods here too `{ join, extname }`

generate-docs.js
```diff
@@ -0,0 +1,167 @@
...
+let { promisify } = require('util')
+let documentation = require('documentation')
+
+let mkdir = promisify(fs.mkdir)
+let writeFile = promisify(fs.writeFile)
+
+let API_FOLDER = path.join(process.cwd(), 'api')
```
ai: `const` if you use CAMEL_CASE

generate-docs.js
isolovev: It a good idea

isolovev:
I want to add Parcel for bundling JS, CSS and HTML files but links at fonts prevent is dream

generate-docs.js
ai: Never use `process.cwd()` in this case. CLI tool can be called from different dir `../generate-docs.js`. Use `__dirname` (I will replace it)

PR #1273, opened 2019-05-22
Update api.md

Ore4444:
Jumping through postcss.d.ts is what made it click for me


ai:
Hm. Why it is better than JSDoc?

Ore4444:
@ai i didn't say it's better, but it's another good way.
I personally found it more convenient, but it's a matter of taste. 
Don't you think people should be a aware of the option?

ai:
`api.md` is a “redirect” file. We have a JSDoc there. Now we put it on the website and keep the file for old links.

This is why it not a list of advice. It is for people who clicked on “PostCSS JSDoc” link. It will be strange to suggest other options on this page.

Ore4444:
Makes perfect sense
Thanks for the explanation

PR #1277, opened 2019-06-04
fix(types): make `Rule.selectors` non-optional

buschtoens:
`Rule.selectors` is guaranteed to be an array. The array itself _may_ be empty.

https://github.com/postcss/postcss/blob/d245ada946919df91009fd1745a7075bf5e84e43/lib/rule.es6#L22-L40

https://github.com/postcss/postcss/blob/d245ada946919df91009fd1745a7075bf5e84e43/lib/list.es6#L12-L53


ai:
Released in 7.0.17

PR #1279, opened 2019-06-16
Add postcss-simple-trig to list

Rplus:
postcss-simple-trig
https://github.com/Rplus/postcss-simple-trig
> A PostCSS plugin to calculate trigonometric functions: sin/cos/tan


ai:
Great plugin 👍

PR #1282, opened 2019-06-28
Json map

Melluish:


ai:
Thanks. I will release your fix in the next 7.0 release when @BondarenkoAlex will finished docs

PR #1284, opened 2019-07-03
Fix source map loading

### LINKED ISSUE #1272: Could not load existing sourcemap of "node_modules/postcss/lib/map-generator.js".
Hey, I tried adding postcss/postcssJs to my code and during build it gives me this error.

error1

I've implemented postcss-js this way and I use parcel-bundler to build my project.

```
const postcssJs = require("postcss-js");
const autoprefixer = require("autoprefixer");
const prefixer = postcssJs.sync([autoprefixer]);

        const style = prefixer({
            gridArea: X
        });

Y.style.gridArea = style

```

Am I doing something wrong or is it actually broken in some way?

kekexunxun:
Fixes #1272

### DIFF
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -139,7 +139,7 @@ class MapGenerator {
     let eol = '\n'
     if (this.css.indexOf('\r\n') !== -1) eol = '\r\n'
 
-    this.css += eol + '/*# sourceMappingURL=' + content + ' */'
+    this.css += eol + '/\*# sourceMappingURL=' + content + ' */'
   }
 
   outputFile () {


ai:
ESLint says that this `\` is unnecessary. How exactly it solves the problem?

kekexunxun:
this.css += eol + "/*# sourceMappingURL=" + content + "*/";
In this code, "# sourceMappingURL=" is omitted because it comes after "/*". So I try to use "\" to mention that this is not a Comment. So I think It's necessary to use "\".
I tried to split "/" and "*", it worked fine.
image
So I think it's necessary to split "/" and "*". And as I mentioned in #1272 , I'm still confused why that plugin caused this problem. I'll try to find out when I have enough time.

ai:
> In this code, "# sourceMappingURL=" is omitted because it comes after "/*".

Who omitted `sourceMappingURL` and in what case? By source maps spec `sourceMappingURL` should be in the CSS comment to be parsed by the browser.

JS as well understands that `"/**/"` is not a comment but string.

kekexunxun:
image
image
image

kekexunxun:
without `"sanitize-html": "^1.20.1"`, it works fine
image

ai:
Maybe we have an error in `sanitize-html` and it parse JS wrong?

louh:
This is a known bug in Parcel. See here for more information: https://github.com/parcel-bundler/parcel/pull/2427

There's a workaround which requires using the `--no-source-map` option in the CLI to avoid running into the bug.

PR #1292, opened 2019-09-05
fix(types): return `Warning[]` from `warnings()`

buschtoens:
https://api.postcss.org/Result.html#warnings
https://api.postcss.org/LazyResult.html#warnings


ai:
Thanks. Released in 7.0.18.

PR #1307, opened 2019-12-05
correct type for `annotation`

keithamus:
Setting `annotation: true` has a different behaviour to omitting the `annotation` key, but this is not allowable by the type system. 

This PR fixes this, so that `annotation: true` is an allowed option.


ai:
Thanks. Released in 7.0.24.

PR #1310, opened 2019-12-16
parse: avoid re-resolution of "already absolute" file paths

tomrav:
On windows, when calling `postcss.parse` providing it with a posix-style absolute file path (e.g. running against an in-memory fs) leads to the `from` field being resolved to a windows-style path.

For example, the file from path `/a/b/c` gets resolved to `C:\\a\\b\\c`, which doesn't exist in my case.

### DIFF
--- a/lib/input.es6
+++ b/lib/input.es6
@@ -41,7 +41,7 @@ class Input {
     }
 
     if (opts.from) {
-      if (/^\w+:\/\//.test(opts.from)) {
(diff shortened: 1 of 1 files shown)


ai:
Thanks. I will release the fix today.

tomrav:
Thanks for the quick fix! 
My twitter handle is RavivTom, though I'm far from active there these days.

ai:
Released in 7.0.25

PR #1316, opened 2019-12-30
Add the stringifier argument to NodeBase.prototype.toString()'s type

### LINKED ISSUE #1313: NodeBase.toString()'s types don't include a parameter
According to [the API docs](http://api.postcss.org/Node.html#toString), the `Node.toString()` function takes a parameter of type `stringifier | syntax`. However, this isn't reflected in the TypeScript interface:

https://github.com/postcss/postcss/blob/4d8a9859c6dbbbdfa159bd04939f6184ebedfe67/lib/postcss.d.ts#L603-L606

nex3:
Closes #1313

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -603,7 +603,7 @@ declare namespace postcss {
     /**
      * @returns A CSS string representing the node.
      */
-    toString(): string;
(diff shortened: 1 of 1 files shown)


ai:
Oops. I forgot to release. Released at 7.0.26.

Thanks for the fix.

PR #1318, opened 2020-01-07
[Docs]: Added reference to contributing guidelines

jamesgeorge007:
It is definitely a good idea to provide a reference to the `contributing guidelines` from the docs that people may miss out if otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,10 +179,6 @@ More articles and videos you can find on [awesome-postcss](https://github.com/jj
 
 * [Mastering PostCSS for Web Design](https://www.packtpub.com/web-development/mastering-postcss-web-design) by Alex Libby, Packt. (June 2016)
 
-## Contributing
-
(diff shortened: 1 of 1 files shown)


ai:
I didn't add it, because it could be out of context for most of repo visitors.

99% came here to find how to use PostCSS bin their project. In this case “Usage” section is much more important.

Initially I hoped that GitHub will show contribution docs to the users who plan to do contribution in PR UI. Is it stopped to work?

Should we move the link below?

jamesgeorge007:
If someone wishes to contribute to any project the first thing he/she would look out for is the `contributing docs` from which they could get to know regarding the local environment setup and related stuffs. I would say this is definitely worth adding as part of the docs in regard to the previous statement.

ai:
Lens move link below in README

jamesgeorge007:
@ai is this what you intended for?

ai:
Yeap. Thanks :)

PR #1320, opened 2020-01-17
Fix circular dependencies

mapgrid:
While building a recent project with Rollup, we ran in to [issues with circular dependencies](https://github.com/rollup/plugins/issues/164) in PostCSS. This PR solves circular dependencies across 3 files, and adds consistency to how files are included.

#### A sampling of current circular dependencies
```
./container.js -> ./rule.js -> ./container.js
./container.js -> ./parse.js -> ./parser.js -> ./rule.js -> ./container.js
./css-syntax-error.js -> terminal-highlight.js -> input.js -> css-syntax-error.js
./parse.js -> parser.js -> at-rule.js -> container.js -> parse.js
...
```

#### `container.es6`

In `container.es6` there are 3 late-style `require` in the `normalize` function. This PR removes the `require` completely and replaces it with a register dependency pattern.

#### `root.es6`

In `root.es6` there are 2 late-style `require` in the `toResult` function. This PR removes the `require` completely and replaces it with a register dependency pattern.

#### `terminal-highlight.es6`

`terminal-highlight.es6` does not use `require` to include `input.es6` in the terminalHighlight function. None the less, in a different environment, this would cause the same error. This PR replaces the import statement with a register dependency pattern.

#### Improvement

- All code is included with `import` or by registering a dependency
- 5 `require` statements removed
- No public API is changed
- PostCSS can be bundled with Rollup or other bundlers

### DIFF
--- a/lib/at-rule.es6
+++ b/lib/at-rule.es6
@@ -1,4 +1,4 @@
-import Container from './container'
+import Container, { registerAtRule } from './container'
 
 /**
  * Represents an at-rule.
@@ -90,3 +90,5 @@ class AtRule extends Container {
 }
 
 export default AtRule
+
+registerAtRule(AtRule)
--- a/lib/container.es6
+++ b/lib/container.es6
@@ -2,6 +2,20 @@ import Declaration from './declaration'
 import Comment from './comment'
 import Node from './node'
 
+let parse, Rule, AtRule
+
+export function registerParse (dependant) {
+  parse = dependant
+}
+
+export function registerRule (dependant) {
+  Rule = dependant
+}
+
+export function registerAtRule (dependant) {
+  AtRule = dependant
+}
+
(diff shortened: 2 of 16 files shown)


ai:
Hi. Can you backport (or maybe move) changes to `ose` branch as well?

Looks good, but I will check it deeper a few days later (need to rest after the conference)

mapgrid:
I can do that. I'm curious why the revert from `import` to `require` in the `ose` branch?

ai:
> I'm curious why the revert from import to require in the ose branch?

1. Babel `import` !== ESM imports. Current code will not work in real ESM without Babel.
2. In `ose` we stopped to use Babel, but we need to support Node.js 10 without enabled ESM.

mapgrid:
Superseded by #1321.

PR #1321, opened 2020-01-19
Fix circular dependencies

mapgrid:
While building a recent project with Rollup, we ran in to [issues with circular dependencies](https://github.com/rollup/plugins/issues/164) in PostCSS. This PR solves circular dependencies across 3 files, and adds consistency to how files are included.

#### A sampling of current circular dependencies
```
./container.js -> ./rule.js -> ./container.js
./container.js -> ./parse.js -> ./parser.js -> ./rule.js -> ./container.js
./css-syntax-error.js -> terminal-highlight.js -> input.js -> css-syntax-error.js
./parse.js -> parser.js -> at-rule.js -> container.js -> parse.js
...
```

#### `container.js`

In `container.js` there are 3 late-style `require` in the `normalize` function. This PR removes the `require` completely and replaces it with a register dependency pattern.

#### `root.js`

In `root.js` there are 2 late-style `require` in the `toResult` function. This PR removes the `require` completely and replaces it with a register dependency pattern.

#### `terminal-highlight.es6`

In `terminal-highlight.js` there is 1 late-style `require` in the `terminalHighlight` function. This PR removes the `require` completely and replaces it with a register dependency pattern.

#### Improvement

- All code is included at the top level of the file or by registering a dependency
- 6 `require` statements removed
- No public API is changed
- PostCSS can be bundled with Rollup or other bundlers

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -4,18 +4,6 @@ let Node = require('./node')
 
 let parse, Rule, AtRule
 
-function registerParse (dependant) {
-  parse = dependant
-}
-
-function registerRule (dependant) {
-  Rule = dependant
-}
-
-function registerAtRule (dependant) {
-  AtRule = dependant
-}
-
 function cleanSource (nodes) {
   return nodes.map(i => {
     if (i.nodes) i.nodes = cleanSource(i.nodes)
@@ -703,12 +691,19 @@ class Container extends Node {
    */
 }
 
-module.exports = {
-  default: Container,
-  registerParse,
-  registerAtRule,
-  registerRule
+Container.registerParse = dependant => {
+  parse = dependant
 }
+
+Container.registerRule = dependant => {
+  Rule = dependant
+}
+
+Container.registerAtRule = dependant => {
+  AtRule = dependant
+}
+
+module.exports = Container
 
 /**
  * @callback childCondition
--- a/lib/container.js
+++ b/lib/container.js
@@ -2,6 +2,20 @@ let Declaration = require('./declaration')
 let Comment = require('./comment')
 let Node = require('./node')
 
+let parse, Rule, AtRule
+
+function registerParse (dependant) {
+  parse = dependant
+}
+
+function registerRule (dependant) {
+  Rule = dependant
+}
+
+function registerAtRule (dependant) {
+  AtRule = dependant
+}
+
 function cleanSource (nodes) {
   return nodes.map(i => {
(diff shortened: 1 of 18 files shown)


lib/container.js
```diff
@@ -692,7 +703,12 @@ class Container extends Node {
    */
 }
 
-module.exports = Container
+module.exports = {
```
ai: It will break `require('postcss/lib/Container')`, but you can add `registerParse`/etc as static methods like `Container.registerParse = dependant => { … }`

test/root.test.js
```diff
@@ -1,5 +1,6 @@
 let Result = require('../lib/result')
 let parse = require('../lib/parse')
+require('../lib/processor')
```
ai: What test was broken without this call?

lib/container.js
mapgrid: Much nicer. Fixed.

test/root.test.js
mapgrid: The tests don't import `lib/postcss.js` so the register side-effects aren't run. In this case, `root` requires `registerProcessor` to be run.

ai: Yeap, I got it. My question is what will be broken without this import.

mapgrid: Another (possibly better) option is adding this line to the top of all tests:

```js
require('../lib/postcss')
```

To avoid this confusion as to why `require('../lib/processor')` is needed occasionally.

lib/css-syntax-error.js
```diff
@@ -1,7 +1,7 @@
 let supportsColor = require('supports-color')
 let chalk = require('chalk')
 
-let terminalHighlight = require('./terminal-highlight')
+let { default: terminalHighlight } = require('./terminal-highlight')
```
ai: We can’t change exports since it will break API

lib/terminal-highlight.js
```diff
@@ -56,3 +61,8 @@ module.exports = function terminalHighlight (css) {
   }
   return result
 }
+
+module.exports = {
+  default: terminalHighlight,
+  registerInput
```
ai: You can do the same `terminalHighlight.registerInput = registerInput` to not chang exports

ai:
Great work! Do you have Twitter handle to mention you on our Twitter?

mapgrid:
Thanks! Yes. Brand new account actually, getting ready to launch this new initiative: [@mapgridorg](https://twitter.com/mapgridorg)

PR #1323, opened 2020-02-05
Added postcss-viewport-height-correction

Faisal-Manzer:
postcss-viewport-height-correction solves the popular problem when 100vh doesn’t fit the mobile browser screen.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -450,7 +450,8 @@ See also plugins in modular minifier [`cssnano`].
 * [`stylefmt`] modern CSS formatter that works well with `stylelint`.
 * [`postcss-autocorrect`] corrects typos and notifies in the console.
 * [`postcss-px-to-viewport`] generates viewport units (`vw`, `vh`, `vmin`, `vmax`) from `px` units.
-* [`postcss-viewport-height-correction`] "corrects" `100vh` which doesn’t fit the mobile browser screen.
+* [`postcss-viewport-height-correction`] solves the popular problem when 100vh
+  doesn’t fit the mobile browser screen.
 
 [flexbox bugs]: https://github.com/philipwalton/flexbugs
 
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -450,7 +450,7 @@ See also plugins in modular minifier [`cssnano`].
 * [`stylefmt`] modern CSS formatter that works well with `stylelint`.
 * [`postcss-autocorrect`] corrects typos and notifies in the console.
 * [`postcss-px-to-viewport`] generates viewport units (`vw`, `vh`, `vmin`, `vmax`) from `px` units.
-
+* [`postcss-viewport-height-correction`] solves the popular problem when 100vh doesn’t fit the mobile browser screen.
```
ai: 1. `100vh` should be as a code:

  ```md
  `100vh`
  ```
2. You need to keep lines in 80 symbols long.

Faisal-Manzer: Corrected docs according to recommendation

docs/plugins.md
```diff
@@ -450,7 +450,7 @@ See also plugins in modular minifier [`cssnano`].
 * [`stylefmt`] modern CSS formatter that works well with `stylelint`.
 * [`postcss-autocorrect`] corrects typos and notifies in the console.
 * [`postcss-px-to-viewport`] generates viewport units (`vw`, `vh`, `vmin`, `vmax`) from `px` units.
-
+* [`postcss-viewport-height-correction`] "corrects" `100vh` which doesn’t fit the mobile browser screen.
```
ai: Let’s avoid quoted `"corrects"`. User will not understand why it is not normal correct. You can use old description, just split it into two lines.

Faisal-Manzer: Does splitting mean? In markdown file or in markdown preview?

ai:
```
A B
```

```
A
B
```

PR #1327, opened 2020-02-17
Update the type annotation for Input.origin

nex3:
This can return false if no origin is available.


ai:
Thanks

I created a [special tool to check `.d.ts` content](https://github.com/ai/check-dts) in my future projects. Hope types will be better in the future.

ai:
Released at 7.0.27

PR #1338, opened 2020-03-13
Update packs in plugins.md

denizdogan:
Remove abandoned packs and update Rucksack URL

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -34,32 +34,21 @@ Or enable plugins directly in CSS using [`postcss-use`]:
 
 * [`postcss-utilities`] includes the most commonly used mixins, shortcuts
   and helpers to use as `@util` rules.
-* [`atcss`] contains plugins that transform your CSS according
-  to special annotation comments.
 * [`cssnano`] contains plugins that optimize CSS size for use in production.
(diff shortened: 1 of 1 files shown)


ai:
What is criteria for abandonment?

denizdogan:
@ai Good question - I don't know! :)

The reason why I made this PR was because I tried using level4 but then got an error message telling me it was written for PostCSS 5.x but I am using PostCSS 7.x. Then I saw on GitHub that the last release of level4 was 4+ years old and [the latest issue discussion seems to indicate](https://github.com/stephenway/level4/issues/3) that it's been superseded by postcss-next.

About Hamster, it hasn't been updated for about 2.5 years, the website seems to not work anymore and it targets PostCSS 6.x. The GitHub user that owns the project looks like they haven't been active at all since 2018.

Oldie hasn't seen an update in 5 years and targets PostCSS 5.x. The same is true for AtCSS.

So my reasoning behind creating this PR is that I'd like for it to be easier for people to find maintained PostCSS packs which work with the latest stable version. I could see a point in keeping them in the list, but maybe they'd fit better into a separate category for packs that are not compatible with the current PostCSS?

ai:
We can remove all plugins which do not updated at least to PostCSS 6.0

PR #1356, opened 2020-05-07
Further improve type declarations

nex3:


ai:
Thanks. I will release it only on weekend. Too much work :(.

nex3:
Totally fine, thanks Andrey!

ai:
In the new version, I want to use TypeScript in tests to be sure that all types are correct

ai:
Released at 7.0.30

PR #1358, opened 2020-05-09
Use @babel/plugin-proposal-export-default-from

pravi:
Babel 7 has official @babel/plugin-proposal-export-default-from to replace third party add-module-exports plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,10 +38,10 @@
   },
   "devDependencies": {
     "@babel/core": "^7.4.3",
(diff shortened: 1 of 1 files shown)


ai:
1. CI falls after the plugin changes
2. PostCSS will not use Babel in the next 8.0 version. I do not want to change Babel plugins for PostCSS 7.0 branch, because it could bring bugs. It will be better to spend time to release PostCSS 8.0 faster.

PR #1294, opened 2019-09-16
Fix: Reads source maps only when it should

brandonkal:
Fixes: https://github.com/stylelint/stylelint/issues/4282
Removes false positives

PR #1360, opened 2020-05-24
Fix: Reads only the last sourceMappingURL in the file

emzoumpo:
Inspired by #1294, another attempt since that one is closed.

### DIFF
--- a/lib/previous-map.es6
+++ b/lib/previous-map.es6
@@ -72,8 +72,30 @@ class PreviousMap {
   }
 
   loadAnnotation (css) {
-    let match = css.match(/\/\*\s*# sourceMappingURL=(.*)\s*\*\//)
-    if (match) this.annotation = match[1].trim()
+    let sourceMapRegExp = '/\\*\\s*# sourceMappingURL=(.*)\\s*\\*/'
+
+    /* eslint-disable security/detect-non-literal-regexp */
+    let lastSourceMapRegex = new RegExp(sourceMapRegExp, 'mg')
+    let sourceMapSourceRegex = new RegExp(sourceMapRegExp)
+    /* eslint-enable security/detect-non-literal-regexp */
+
+    let lastSourceMapMatch
+    let sourceMapSourceMatch
+    let lastSourceMap
+    let sourceMapSource
+
+    lastSourceMapMatch = css.match(lastSourceMapRegex)
+
+    if (lastSourceMapMatch) {
+      // Locate the last sourceMappingURL to avoid picking up
+      // sourceMappingURLs from comments, strings, etc.
+      lastSourceMap = lastSourceMapMatch[lastSourceMapMatch.length - 1]
+      sourceMapSourceMatch = lastSourceMap.match(sourceMapSourceRegex)
+      if (sourceMapSourceMatch) {
+        sourceMapSource = sourceMapSourceMatch[1]
+        this.annotation = sourceMapSource.trim()
+      }
+    }
   }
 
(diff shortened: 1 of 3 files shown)


lib/previous-map.es6
```diff
@@ -72,8 +72,30 @@ class PreviousMap {
   }
 
   loadAnnotation (css) {
-    let match = css.match(/\/\*\s*# sourceMappingURL=(.*)\s*\*\//)
-    if (match) this.annotation = match[1].trim()
+    let sourceMapRegExp = '/\\*\\s*# sourceMappingURL=(.*)\\s*\\*/'
+
+    /* eslint-disable security/detect-non-literal-regexp */
+    let lastSourceMapRegex = new RegExp(sourceMapRegExp, 'mg')
+    let sourceMapSourceRegex = new RegExp(sourceMapRegExp)
```
ai, line 79: We can make the code much readable if we explicitly split it into “find” and “parse” steps.

What if we will add an additional `getAnnotationURL()` method and we will use it in `loadAnnotation()` after founding the last annotation? Do not afraid `'/\\*\\s*# sourceMappingURL=(.*)\\s*\\*/'` dublication in this case.

emzoumpo: @ai I agree, it could be more readable with splitting the functionality.
I just pushed the related changes, let me know if it looks good to you.
Thanks!

lib/previous-map.es6
```diff
@@ -71,9 +71,23 @@ class PreviousMap {
...
+  getAnnotationURL (sourceMapString) {
+    return sourceMapString
+      .match(/\/\*\s*# sourceMappingURL=(.*)\s*\*\//)[1]
+      .trim()
+  }
+
   loadAnnotation (css) {
-    let match = css.match(/\/\*\s*# sourceMappingURL=(.*)\s*\*\//)
-    if (match) this.annotation = match[1].trim()
+    let sourceMapMatch = css.match(/\/\*\s*# sourceMappingURL=(.*)\s*\*\//mg)
```
ai, line 81: Much better! What do you think of adding plural form here `let annotations =`, to mark that it is an array?

lib/previous-map.es6
```diff
@@ -71,9 +71,23 @@ class PreviousMap {
...
+
+    if (sourceMapMatch) {
```
ai, line 83: You can simplify these checks by:

```suggestion
    if (annotations && annotation.length > 0) {
```

lib/previous-map.es6
emzoumpo: Sure, sounds better, just pushed again!

lib/previous-map.es6
emzoumpo: I pushed this changed as well

ai:
Looks great 😸.

Can we add tests to avoid regressions in the future? Or it is hard to find a good test for it?

emzoumpo:
Thanks!
Yes, I think a test here would be nice to have, in order to avoid regressions.
I will check whether I can find a good test when I'm off work and will update you .

emzoumpo:
@ai I wrote a very small test that will fail on current `master` and will pass on my PR, so it would be able to catch future regressions.

ai:
Do you have Twitter account to mention you in `@postcss` tweet about this release?

ai:
Released in 7.0.31

emzoumpo:
Hi @ai, thanks for merging my PR! It was pretty late here and I was sleeping. I don't have twitter so it's fine 😄

PR #1368, opened 2020-06-16
Allow replaceWith to be called with input that includes itself.

josephkaptur:
Consider an AST where a node `root` has child nodes `[a, b, c]` and there are other nodes `d` and `e`. `a.replaceWith(a, d, e)` now results in `root` having the children `[a, d, e, b, c]`.


ai:
Thanks

PR #1369, opened 2020-06-17
chore: simplify colorizer

lukeed:
PostCSS has very little _actual_ use of ANSI color codes, so it didn't really make sense to use the largest color library – both in terms of size and [dependency chains](https://npm.anvaka.com/#/view/2d/chalk). 

You were also pulling in `supports-color`, which is mostly determining _levels_ of color support beyond the basics. The colors you're using are the lowest level and don't need extensive detection. 

Here I'm suggesting using `kleur` instead, which is a fast, no-dependency, single-file module. It also includes basic color detection, so no additional library is needed. Also, even though PostCSS isn't using ESM (yet?), `kleur` is compatible with native ESM and `kleur/colors` is fully treeshakeable 🌴 

There is **one difference** made in this PR:
PostCSS will now print colors in browser environments~! I can revert this, but I thought it would be a good addition. Only _colors_ are supported – which, for your use case, means that only `bold()` won't have an effect.

Let me know your thoughts & if I should make any changes.
If you don't want to make these changes, I totally understand :)

---

By the way, the Jest suites, `check-dts`, and `size-limit` are passing, but `test/integration.js` is failing on the Twitter CSS step. This is also the case with the base `ose` and `master` branches, so you're probably already aware of it.


lib/css-syntax-error.js
```diff
@@ -175,8 +174,8 @@ class CssSyntaxError extends Error {
 
     let css = this.source
     if (terminalHighlight) {
-      if (typeof color === 'undefined') color = supportsColor.stdout
-      if (color) css = terminalHighlight(css)
+      if (color != null) kleur.$.enabled = color
```
ai, line 177: Does `kleur` have color support autodetect build-in?

lib/css-syntax-error.js
```diff
@@ -175,8 +174,8 @@ class CssSyntaxError extends Error {
...
+      if (kleur.$.enabled) css = terminalHighlight(css)
```
lukeed, line 178: kleur reads the environment and always tries to be enabled by default. Here if `color: boolean` is passed, it modifies the kleur context and takes immediate effect.

lib/css-syntax-error.js
```diff
@@ -186,16 +185,10 @@ class CssSyntaxError extends Error {
     let maxWidth = String(end).length
 
     function mark (text) {
-      if (color && chalk.red) {
```
lukeed, line 189: Don't have to use `color` check anymore, since `kleur` will [internally no-op](https://github.com/lukeed/kleur/blob/master/colors.mjs#L12) if `kleur.$.enabled` is falsey.

And I removed the `chalk.red` check since kleur has basic browser support (mentioned above).

lib/css-syntax-error.js
lukeed: Yup 👍

ai:
Thanks!

lukeed:
Thank you :D

ai:
I also replaced `chalk` to `kleur` in all my open source

lukeed:
Oh wow, that's awesome! I'm glad you like it :)

PR #1379, opened 2020-07-25
Fix TypeError: Cannot set property 'parent' of undefined on removeChild

ranisalt:
I don't know how or why does it happen, but this error has been reported in vkalinichev/postcss-rtl#57 and this fix generates the same CSS files while preventing this error using said plugin.

I didn't have enough time to thoroughly debug either postcss or postcss-rtl for this.

### DIFF
--- a/lib/container.es6
+++ b/lib/container.es6
@@ -437,7 +437,7 @@ class Container extends Node {
    */
   removeChild (child) {
     child = this.index(child)
-    this.nodes[child].parent = undefined
+    (this.nodes[child] || {}).parent = undefined
     this.nodes.splice(child, 1)
 
     let index


ai:
As we told in [our Twitter](https://twitter.com/postcss) we do not accept PR to `master` because we made big refactoring for next PostCSS 8. Send PR to `ose`. This is why I am closing this PR.

ai:
Talking about the fix. I think we are fixing symptoms instead of fixing the real problem.

Seems like `postcss-rtl` breaks the AST. By fixing the symptom we do not fix the source of broken AST and you will have problems in other places.

ai:
Very likely that plugin try to remove the node, which already was removed

PR #1417, opened 2020-09-19
Support Class Plugins

jonathantneal:
**WIP**: This is just an idea. I’m unable to see it through right now.

Supporting class functions would allow PostCSS plugins to more easily setup and share data.

Currently, users need to create one or two functions to setup data. New authors will need to familiarize themselves with these patterns specific to PostCSS:

```js
module.exports = opts => ({
  postcssPlugin: "PLUGIN NAME",
  prepare(result) {
    const variables = {}
    return {
      Declaration(node) {
        if (node.variable) variables[node.prop] = node.value
      },
      RootExit() {
        console.log(variables)
      },
    }
  },
})
```

Using classes, the pattern will be more familiar to JavaScripts authors.

```js
module.exports = class PostCSSPlugin {
  constructor(opts, result) {
    this.variables = {}
  }
  get postcssPlugin() {
    return 'PLUGIN NAME'
  }
  Declaration(node) {
    if (node.variable) this.variables[node.prop] = node.value
  }
  RootExit() {
    console.log(this.variables)
  }
}
```

Furthermore, a class pattern would allow plugins to _extend_ one another. For instance, someone could create a plugin that extends visitors to selectors or values.

This PR prevents class functions from throwing an error, and supports them as a kind of visitor, only where the `constructor()` takes the place of an object `prepare()`.

Currently, using a class function throws an unexpected error to do with `i = i()`. This PR happens to resolve that, but it could also be resolved separately.

### DIFF
--- a/lib/lazy-result.js
+++ b/lib/lazy-result.js
@@ -109,15 +109,7 @@ class LazyResult {
     this.result = new Result(processor, root, opts)
     this.helpers = { ...postcss, result: this.result, postcss }
     this.plugins = this.processor.plugins.map(plugin => {
-      if (
-        typeof plugin === 'function' &&
-        plugin.prototype &&
-        plugin.prototype.constructor === plugin &&
-        plugin.postcss === true
-      ) {
-        // eslint-disable-next-line new-cap
-        return new plugin(this.result)
-      } else if (typeof plugin === 'object' && plugin.prepare) {
+      if (typeof plugin === 'object' && plugin.prepare) {
         return { ...plugin, ...plugin.prepare(this.result) }
       } else {
         return plugin
--- a/lib/lazy-result.js
+++ b/lib/lazy-result.js
@@ -109,7 +109,15 @@ class LazyResult {
     this.result = new Result(processor, root, opts)
     this.helpers = { ...postcss, result: this.result, postcss }
     this.plugins = this.processor.plugins.map(plugin => {
-      if (typeof plugin === 'object' && plugin.prepare) {
+      if (
+        typeof plugin === 'function' &&
+        plugin.prototype &&
+        plugin.prototype.constructor === plugin &&
+        plugin.postcss === true
+      ) {
+        // eslint-disable-next-line new-cap
+        return new plugin(this.result)
+      } else if (typeof plugin === 'object' && plugin.prepare) {
         return { ...plugin, ...plugin.prepare(this.result) }
       } else {
         return plugin
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -93,8 +93,8 @@ export interface PluginClass extends Processors {
   postcssPlugin: string
 }
 
-export interface PluginConstructor {
-  new (opts: any): PluginClass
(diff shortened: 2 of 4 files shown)


lib/lazy-result.js
```diff
@@ -109,7 +109,15 @@ class LazyResult {
...
-      if (typeof plugin === 'object' && plugin.prepare) {
+      if (
+        typeof plugin === 'function' &&
+        plugin.prototype &&
+        plugin.prototype.constructor === plugin &&
+        plugin.postcss === true
+      ) {
+        // eslint-disable-next-line new-cap
+        return new plugin(this.result)
+      } else if (typeof plugin === 'object' && plugin.prepare) {
```
jonathantneal, line 120: I left this line intact, but I think it would not hurt to check the type of `plugin.prepare`.

```suggestion
      } else if (typeof plugin === 'object' && typeof plugin.prepare === 'function') {
```

lib/processor.js
```diff
@@ -36,8 +36,8 @@ class Processor {
   normalize (plugins) {
     let normalized = []
     for (let i of plugins) {
-      if (i.postcss === true) {
-        i = i()
+      if (typeof i === 'function' && i.postcss === true) {
```
jonathantneal, line 39: This line protects PostCSS from attempting to call non-functions with a `postcss` property of `true`.

lib/processor.js
```diff
@@ -36,8 +36,8 @@ class Processor {
...
+        if (!i.prototype || i.prototype.constructor !== i) i = i()
```
jonathantneal, line 40: Following the previous check, this line protects PostCSS from calling classes, which would throw an error about the missing `new`.

lib/postcss.d.ts
```diff
@@ -89,6 +89,16 @@ export interface Plugin extends Processors {
   prepare?: (result: Result) => Processors
 }
 
+export interface PluginClass extends Processors {
+  postcssPlugin: string
+}
+
+export interface PluginConstructor {
+  new (opts: any): PluginClass
```
Semigradsky, line 97: These are not options as I can see
```ts
new (result: Result): PluginClass
```

lib/postcss.d.ts
```diff
@@ -89,6 +89,16 @@ export interface Plugin extends Processors {
...
+  prototype: PluginClass
+  postcss: boolean
```
Semigradsky, line 99:
```ts
postcss: true
```
This is more accurate

ai:
How to use it? `new require('plugin')(opts)`?

evilebottnawi:
Sorry for off top, but I like this idea

ai:
New class API will lead to minor release. I prefer to finish the rough period of PostCSS 7 → 8 transition and only then think about extending API.

I understand that you want a new shiny API for your plugins, but I just will not have enough time to fix all issues and think about the next API.

Let’s move thinking a little later,

matin417567-LGTM (approved):

ai:
Closing for now

Issue #1430, opened 2020-09-26
Deno support


oscarotero:
I've tried to install postcss in Deno using JSPM (https://dev.jspm.io/postcss) that converts the node code to standard javascript modules and seems that the `nanoid` dependency is failing, because I get this error:

```
error: Uncaught TypeError: Import 'https://dev.jspm.io/npm:nanoid@3.1.12/non-secure/index.dew.js' failed: 500 Internal Server Error
Imported from "https://dev.jspm.io/npm:nanoid@3/non-secure?dew:1"
```

This only happens with postcss 8, version 7.x can be installed correctly.

Due the source code of postcss is typescript, I guess it shouldn't require much effort to have a Deno version. Or, at least, fix these errors generated by jspm converter.

FYI, there's an attempt to port postcss to deno here: https://github.com/denolib/deno-postcss

ai:
> I've tried to install postcss in Deno

I like Deno and would like to add Deno support to PostCSS.

> Due the source code of postcss is typescript

Nope. PostCSS was written on pure JS (ES6+ with CommonJS modules). For TS we have only `.d.ts` files.

> This only happens with postcss 8, version 7.x can be installed correctly.

In PostCSS 8 we added `nanoid` dependency. `nanoid` has good ES modules support.

> FYI, there's an attempt to port postcss to deno here: https://github.com/denolib/deno-postcss

The fork is the last option. Very soon it will stop maintaince. We need to find another way.

> Or, at least, fix these errors generated by jspm converter.

What error do we have on JSPM? I see only 500 HTTP error without details, why JSPM wasn’t able to compile Nano ID.

oscarotero:
Hi. It's great you want to add Deno support!

> The fork is the last option. Very soon it will stop maintaince. We need to find another way.

Completely agree. There should be a way to create a code that can be exported for both platforms. Some libraries like [lodash](https://github.com/lodash/lodash) or [date-fns](https://github.com/date-fns/date-fns) are doing that.

> What error do we have on JSPM? I see only 500 HTTP error without details, why JSPM wasn’t able to compile Nano ID.

No idea. Seems an error in the v3 version: https://dev.jspm.io/npm:nanoid@3.1.12/non-secure/index.dew.js 
If you try v2, it seems to work fine: https://dev.jspm.io/npm:nanoid@2.0.0/non-secure/index.dew.js

The problem is in this file: https://github.com/ai/nanoid/blob/master/non-secure/index.js, but I don't see anything strange.
[...]

PR #1432, opened 2020-09-30
Code style changes to Deno port

oscarotero:
Related with https://github.com/postcss/postcss/issues/1430#issuecomment-701043034

I've added some changes in the way that some modules are imported, to ease the automatic transformation to Deno that I'm doing here: https://github.com/oscarotero/postcss-deno


ai:
Sorry for long delay. Forgot to merge.

PR #1442, opened 2020-10-18
Fix CommentRaws interface right type

Arthie:
I believe `right` of the CommentRaws interface should be of type `string` instead of `boolean`


ai:
Thanks. I will release it a few hours later.

ai:
Thanks. Released in 8.1.2.

PR #1448, opened 2020-10-23
Add missing "comment" export in types + reorder

Arthie:
Hello, I noticed "comment" was missing from the types when trying to `import { comment } from "postcss";`.
This PR adds the missing type and reorders the types to match the javascript file.


ai:
Thanks. Released in 8.1.4.

PR #1449, opened 2020-10-23
Update input.js: changed assignment of "this.from" in "Input" constructor

florianmarkusse:
Hello,

I changed the following line:
`if (this.map) this.map.file = this.from`
I did this because `this.from` is undefined at the moment of setting `this.map.file` if the conditional is met. Therefore one can changed the line to:
`if (this.map) this.map.file = undefined`

This changed is not related to any issues but reduces the complexity of that line and improved the readability.

I hope this is clear enough and look forward to your comments.

Kind regards,
Florian

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -46,7 +46,7 @@ class Input {
     if (!this.file) {
       this.id = '<input css ' + nanoid(6) + '>'
     }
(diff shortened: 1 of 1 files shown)


ai:
Tests are falling. Seems like `file` is not `undefined` for some cases.

PR #1459, opened 2020-11-10
Fix usage of default exports with TypeScript

### LINKED ISSUE #1424: Invalid TypeScript Type (Default Export)
PostCSS exports default type as seen in the type file:

https://github.com/postcss/postcss/blob/master/lib/postcss.d.ts

However that is certainly not the case because a code calling default exported type will not run:

```ts

        const postcssResult = await postcss([autoprefixer])
            .process(sassResult.css, postcssOptions);
```

```
TypeError: postcss_1.default is not a function
```

Instead, shouldn't the TypeScript declaration use `export =` ?

https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

**Not to mention, calling `import postcss = require('postcss');` from TypeScript will not compile anymore due to this type declaration change!**

----

Workaround for those affected by this issue: revert back to PostCSS 7

remcohaszing:
This is done by assigning the property `default` for all properties that are exported using `export default` according to their type definitions.

Closes #1424

### DIFF
--- a/lib/at-rule.js
+++ b/lib/at-rule.js
@@ -20,5 +20,6 @@ class AtRule extends Container {
 }
 
 module.exports = AtRule
+AtRule.default = AtRule
 
 Container.registerAtRule(AtRule)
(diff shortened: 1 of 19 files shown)


ai:
Released in 8.1.7

PR #1463, opened 2020-11-18
fix(typescript): ensure LazyResult complies with Promise interface.

ludofischer:
Hi, following our Gitter chat yesterday, I've learnt `LazyResult` was meant to implement the `Promise` contract.  This pull request adds the [[Symbol.toStringTag] property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) to `LazyResult` which Typescript requires to type something as a `Promise`. 

* Explicitly implement Promise<Result> in the type declaration
* Add mising [Symbol.toStringTag] property
* Add some type tests to ensure we can type `postcss(plugins).process` results

### DIFF
--- a/lib/lazy-result.d.ts
+++ b/lib/lazy-result.d.ts
@@ -13,7 +13,7 @@ import Root from './root.js'
  * const lazy = postcss([autoprefixer]).process(css)
  * ```
  */
-export default class LazyResult {
+export default class LazyResult implements Promise<Result> {
   /**
    * Processes input CSS through synchronous and asynchronous plugins
    * and calls `onFulfilled` with a Result instance. If a plugin throws
@@ -65,6 +65,12 @@ export default class LazyResult {
    * @param opts      Options from the `Processor#process` or `Root#toResult`.
    */
   constructor (processor: Processor, css: string, opts: ResultOptions)
+
+  /**
+   * Returns the default string description of an object.
+   * Required to implement the Promise interface.
+   */
+  get [Symbol.toStringTag] (): string
 
   /**
    * Returns a `Processor` instance, which will be used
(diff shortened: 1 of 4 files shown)


lib/lazy-result.d.ts
```diff
@@ -66,6 +66,12 @@ export default class LazyResult {
    */
   constructor (processor: Processor, css: string, opts: ResultOptions)
 
+  /**
+   * Returns the default string description of an object.
+   * Required to implement the Promise interface.
+   */
+  get [Symbol.toStringTag] (): string
```
ai, line 73: Do we need to implement it in the code?

test/types.ts
```diff
@@ -14,4 +14,19 @@ const plugin: PluginCreator<string> = prop => {
...
 
+interface StyleCompileResults {
+  code: string
+  map: SourceMap | undefined
+}
+
+const processResult: Promise<Result> | Result = postcss([
+  plugin
+]).process('h1{color: black;}', { from: undefined })
+const processed:
```
ai, line 25: Why we can use `processed: Result = ` here?

lib/lazy-result.d.ts
ludofischer: Yes, sorry I have added the `lazy-result.js` changes now. Forgot to save the file :-)

test/types.ts
ludofischer: Sorry I am not sure I  understand the question. I don't think `processed ` can be of type result since I am mapping it to a different object in `then()`. If you want I can simplify this example, it actually comes from a real-life sample in  the [Vue.js 3 repo](https://github.com/vuejs/vue-next/blob/master/packages/compiler-sfc/src/compileStyle.ts)).

ai:
Thanks. Released in 8.1.8.

PR #1465, opened 2020-11-20
Mark getters as readonly

toinehartman:
As I understand, all these properties should be treated as `readonly`. This stricter modifier ensures proper typechecking.

Furthermore, this seems to be a required change to fix [my PR on @types/autoprefixer](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49166). See [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49166#issuecomment-718236687); those errors can be fixed by incorporating these changes in `postcss`.

### DIFF
--- a/lib/container.d.ts
+++ b/lib/container.d.ts
@@ -47,7 +47,7 @@ export default abstract class Container extends Node {
    * rule.first === rules.nodes[0]
    * ```
    */
-  get first (): ChildNode | undefined
+  readonly first: ChildNode | undefined
 
   /**
    * The container’s last child.
@@ -56,7 +56,7 @@ export default abstract class Container extends Node {
    * rule.last === rule.nodes[rule.nodes.length - 1]
    * ```
    */
-  get last (): ChildNode | undefined
+  readonly last: ChildNode | undefined
 
   /**
    * Iterates through the container’s immediate children,
--- a/lib/input.d.ts
+++ b/lib/input.d.ts
@@ -103,7 +103,7 @@ export default class Input {
    * root.source.input.from //=> "<input css 1>"
    * ```
    */
-  get from (): string
+  readonly from: string
 
   /**
    * Reads the input source map and returns a symbol position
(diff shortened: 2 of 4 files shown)


ai:
What is the different between `get prop (): X` and `readonly prop: X`?

I am [getting](https://www.typescriptlang.org/play?#code/DYUwLgBA9gRgVhAvBA3gKAhA5uCAHAJyjwgAoBKVDTCA8AVwIDsIByARleoF81e1YcAHSFiSNgCZWQA) the same `Cannot assign to 'prop' because it is a read-only property` error on writting to `get prop ()`.

ai:
Am I right that we need this change to support old TypeScript 3.2 since it doesn’t support getters in some context?

toinehartman:
> Am I right that we need this change to support old TypeScript 3.2 since it doesn’t support getters in some context?

To be honest is wasn't sure **why** this fixes my problem, but only **that** it does that... But your explanation seems to be correct, indeed.

I can imagine that postcss does not target TS 3.2, but as it is used in `@types/autoprefixer`, which needs to support TS 3.2, I stumbled upon this.

ai:
@toinehartman maybe you want to send types directly to `autoprefixer`. In this case we will be free from supporting TS 3.2 😺

toinehartman:
@ai Didn't even think about that, thanks! I'll see if I can get that to work (TS declarations are a bit new to me) and will report here.

toinehartman:
@ai Would it be ok (as in allowed) to copy all relevant `*.d.ts` files to the DefinitelyTyped project?

ai:
> Would it be ok (as in allowed) to copy all relevant *.d.ts files

From `postcss`? Nope. DefinitelyTyped project is a hack for projects, which do not support TS. PostCSS supports TS.

I also think that adding Autoprefixer’s `.d.ts` files to DefinitelyTyped is a hack. It is much better to add `.d.ts` files to `postcss/autoprefixer` repo.

toinehartman:
I mean as part of the @types/autoprefixer definition (because autoprefixer types obviously depend on types postcss defined by postcss).

ai:
I am not sure that the whole `@types/autoprefixer` is a good idea. Why not add it to `postcss/autoprefuxer` official package?

PR #1466, opened 2020-11-21
Replace line-column with vfile-location

TrySound:
vfile-location is dependency free and quite fresh package.
https://github.com/vfile/vfile-location

line-column has outdated dependencies and was not touched in 5 years.
https://github.com/io-monad/line-column

### DIFF
--- a/lib/input.d.ts
+++ b/lib/input.d.ts
@@ -126,7 +126,5 @@ export default class Input {
    *
    * @param offset Source offset.
    */
-  fromOffset (
-    offset: number
-  ): { offset?: number; line?: number; column?: number }
+  fromOffset (offset: number): { line: number; col: number } | null
 }
--- a/lib/input.d.ts
+++ b/lib/input.d.ts
@@ -126,5 +126,7 @@ export default class Input {
    *
    * @param offset Source offset.
    */
-  fromOffset (offset: number): { line: number; col: number } | null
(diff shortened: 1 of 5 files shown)


lib/input.d.ts
```diff
@@ -126,5 +126,7 @@ export default class Input {
    *
    * @param offset Source offset.
    */
-  fromOffset (offset: number): { line: number; col: number } | null
+  fromOffset (
+    offset: number
+  ): { offset?: number; line?: number; column?: number }
```
ai, line 131: We can’t change public API. Let’s convert object.

lib/input.js
```diff
@@ -50,8 +50,14 @@ class Input {
...
 
   fromOffset (offset) {
-    let finder = lineColumn(this.css)
-    this.fromOffset = i => finder.fromIndex(i)
+    let finder = vfileLocation(this.css)
+    this.fromOffset = i => {
+      let position = finder.toPoint(i)
+      return {
+        line: position.line,
+        col: position.column
```
TrySound, line 58: Maybe add TODO here to make api consistent in v9?

ai: I am using issue for that

ai:
LGTM. Will release it a few hours later.

ai:
Released in 8.1.9. Thanks.

PR #1471, opened 2020-11-22
Expose LazyResult type

### LINKED ISSUE #1470: LazyResult is missing from type exports
It was present in 7.x but somehow not exported in the main `d.ts` file in 8.x.

yyx990803:
close #1470

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -13,6 +13,7 @@ import Root, { RootProps } from './root.js'
 import Comment, { CommentProps } from './comment.js'
 import AtRule, { AtRuleProps } from './at-rule.js'
 import Result, { Message } from './result.js'
+import LazyResult from './lazy-result.js'
 import Rule, { RuleProps } from './rule.js'
 import Container, { ContainerProps } from './container.js'
 import Warning, { WarningOptions } from './warning.js'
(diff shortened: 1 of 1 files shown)


ai:
Why do you need the `LazyResult` type? If you need to save `postcss(plugins).process(…)` results, it is better to use `Promise<Result>` (`LazyResult` is a hack to combine `Promise` with sync APIs).

We do not expose this class since it is a semi-private API. For instance, there are no types for `LazyResult#constructor` because users should not create `LazyResult` instances manually.

yyx990803:
The case is that we have an API that allows the user to decide whether it is forced to be sync or not, and we need to `try/catch` the process call:

```js
let result: LazyResult | undefined

try {
  result = postcss(plugins).process(source, options)
  if (isSyncC all) {
    // force sync
    return { code: result.css, map: result.map }
  } else {
    return result.then(r => ({ code: r.css, map: r.map })
  }
} catch (e) {

}
```

There could also be cases where the lazy result needs to be passed into another function where the `LazyResult` type would be needed in its type signature.

We can probably work around this, but it just feels unnecessary when the type can be exported to keep the same behavior as in v7. Plus the `LazyResult` type is publicly documented in the docs: https://postcss.org/api/#lazyresult

ai:
Yeap seems like this semi-private status confuses too many people.

ai:
I will release it after a few hours (too many works calls today)

ai:
Thanks. Released in 8.1.10.

PR #1474, opened 2020-11-24
Fixed Deno conversion workflow

oscarotero:
- Updated the Deno environment setup to the latest version
- Removed the copy commands because those files are already copied by the js script.


ai:
Since we have mostly TS fixes in recent releases, I think it will be OK to make Deno release with next PostCSS patch release

PR #1484, opened 2020-12-04
fromJSON and toJSON

mischnic:
There are rather ugly things in here, but let me know what you think...


lib/input.js
```diff
@@ -49,42 +49,41 @@ class Input {
...
+      for (let i = 0, l = lines.length; i < l; i++) {
+        this.lineToIndex[i] = prevIndex
+        prevIndex += lines[i].length + 1
+      }
 
-    for (let i = 0, l = lines.length; i < l; i++) {
-      lineToIndex[i] = prevIndex
-      prevIndex += lines[i].length + 1
+      this.lastLine = this.lineToIndex[this.lineToIndex.length - 1]
     }
```
mischnic, line 63: This change ensures that it doesn't try to store state in a function, which can't be serialized into JSON. This seems cleaner to me anyway (though an alternative would be to exclude this function from being returned in `Input#toJSON()`)

ai: Yeap. Removing that method replacing hack can be nice.

But I afraid that `this.lineToIndex` will bloat up `console.log(node)`. What if we will use `let lineToIndecCache = new Symbol()`?

test/fromJSON.test.ts
```diff
@@ -0,0 +1,28 @@
...
+      stringifier: postcss.stringify
+    }
+  ).css
+
+  let root = postcss.parse(cssWithMap)
+
+  let json = root.toJSON()
+  let serialized = v8.serialize(json)
+  let deserialized = v8.deserialize(serialized)
+  // @ts-ignore
```
ai, line 22: Am I right that you are planning to add types later?

test/fromJSON.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+it('can rehydrate a JSON AST', () => {
+  let cssWithMap = postcss().process(
+    '.foo { color: red; font-size: 12pt; } /* abc */ @media (width: 60em) { }',
+    {
+      from: undefined,
+      map: {
+        inline: true
+      },
+      stringifier: postcss.stringify
```
ai, line 13: `postcss.stringify` is a default stringifier. I think you can remove it.

test/fromJSON.test.ts
```diff
@@ -0,0 +1,28 @@
+import v8 from 'v8'
+
+import postcss, { Root } from '../lib/postcss.js'
+
+it('can rehydrate a JSON AST', () => {
```
ai, line 5:
```suggestion
it('rehydrates a JSON AST', () => {
```

lib/postcss.js
```diff
@@ -12,6 +12,7 @@ let AtRule = require('./at-rule')
 let Result = require('./result.js')
 let Input = require('./input')
 let parse = require('./parse')
+let fromJSON = require('./fromJSON')
```
ai, line 15: Can you keep imports order based on variable length? (I know that it is a strange system 😅)

lib/input.js
```diff
@@ -168,6 +167,21 @@ class Input {
   get from () {
     return this.file || this.id
   }
+
+  toJSON () {
+    let json = {}
+    if (this.hasBOM != null) json.hasBOM = this.hasBOM
```
ai, line 173: It will be nice to iterate throw through `['hasBOM', 'css', 'file', 'id']`

lib/fromJSON.js
```diff
@@ -0,0 +1,46 @@
...
+
+let Declaration = require('./declaration')
+let Comment = require('./comment')
+let AtRule = require('./at-rule')
+let Root = require('./root')
+let Rule = require('./rule')
+let Input = require('./input')
+let PreviousMap = require('./previous-map')
+
+function fromJSON (json) {
```
ai, line 11: We will need `fromJSON.d.ts` file

lib/fromJSON.js
```diff
@@ -0,0 +1,46 @@
...
+  } else if (json.type === 'decl') {
+    return new Declaration(defaults)
+  } else if (json.type === 'rule') {
+    return new Rule(defaults)
+  } else if (json.type === 'comment') {
+    return new Comment(defaults)
+  } else if (json.type === 'atrule') {
+    return new AtRule(defaults)
+  } else {
+    throw new Error('Unknown node type ' + json.type)
```
ai, line 42: We will need a test for this case too

ai:
Looks good! 👍

Let’s finish small changes and release under 8.2

test/fromJSON.test.ts
mischnic: "You did not set any plugins, parser, or stringifier. Right now, PostCSS does nothing. Pick plugins for your case on https://www.postcss.parts/ and use them in postcss.config.js."

test/fromJSON.test.ts
mischnic: The return value of `fromJSON` depends on the parameter's node type. I've instead casted the type with `as Root`, is that acceptable?

test/fromJSON.test.ts
ai: Try `postcss([])`

test/fromJSON.test.ts
ai: `as Root` :+1:

lib/fromJSON.js
mischnic: Should they be in that file or in `postcss.d.ts`? parse, stringify, ... are also documented in the `postcss.d.ts`

ai: Yeap, we can do it in `postcss.d.ts` too.

test/fromJSON.test.ts
mischnic: This also prints that warning:
```js
  let cssWithMap = postcss([]).process(
    '.foo { color: red; font-size: 12pt; } /* abc */ @media (width: 60em) { }',
    {
      map: {
        inline: true
      },
    }
  ).css
```

ai: OK, let’s use the current code

ai:
I will release 8.2 on the next week (I need to promote previous 8.1.14 release first since it solved some painful 8.x migration problems)

PR #1485, opened 2020-12-06
Proposal: change plugin creator type to allow returning a processor.

ludofischer:
Returning a a `Processor` instance from a plugin creator (function with `postcss = true`) already works in practice, and was supported with PostCSS 7 plugins, but TypeScript types do not allow it. This updates the types and adds a test. 

### Use case
Support 'plugins of plugins' like cssnano (https://github.com/cssnano/cssnano/pull/961) or postcss-modules. These 'meta-plugins' take some options, build a list of plugins, create a `Processor` and run `process()` over the AST. The issue is that in PostCSS 8, this does not notify visitor plugins, forcing these 'meta-plugins' to run as `OnceExit`. This pull request would allow to return a `Processor` instance instead of calling `process()` inside the 'meta-plugin'.

### Alternatives
1. Keep using `OnceExit()` with a separate `process()`. This means 'meta-plugins' and all plugins inside of them could never share the same tree as other plugins.
2. Instead of returning a `Processor`, allow returning any object with a `plugins` field. This would require to recursively call `normalize()` on the plugins arrays in the `Processor` constructor to initialize the plugins. I think it's better to let the meta-plugins take care of all initialization of their own plugins.
2. Do not expose meta-plugins as postcss plugins at all. Instead, force users to call a function that returns a plugin array or a Processor. In a sense I feel it would be better as it would remove a layer of abstaction, but on the other hand it makes configuring the PostCSS instance more complicated, and introduces a breaking change.

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -170,7 +170,7 @@ export interface Plugin extends Processors {
 }
 
 export interface PluginCreator<PluginOptions> {
-  (opts?: PluginOptions): Plugin
+  (opts?: PluginOptions): Plugin | Processor
   postcss: true
 }
 
--- a/test/postcss.test.ts
+++ b/test/postcss.test.ts
@@ -1,4 +1,4 @@
-import postcss, { Root } from '../lib/postcss.js'
+import postcss, { Root, PluginCreator } from '../lib/postcss.js'
 import Processor from '../lib/processor.js'
 
 afterEach(() => {
@@ -28,6 +28,16 @@ it('takes plugin from other processor', () => {
   let b = () => {}
   let c = () => {}
   let other = postcss([a, b])
+  expect(postcss([other, c]).plugins).toEqual([a, b, c])
+})
+
+it('takes plugins from a a plugin returning a processor', () => {
+  let a = () => {}
+  let b = () => {}
+  let c = () => {}
+  let other = postcss([a, b])
+  let meta = (() => other) as PluginCreator<void>
+  meta.postcss = true
   expect(postcss([other, c]).plugins).toEqual([a, b, c])
 })
 
--- a/test/processor.test.ts
+++ b/test/processor.test.ts
@@ -1,6 +1,13 @@
 import { resolve as pathResolve } from 'path'
 
-import postcss, { Plugin, Result, Node, Root, parse } from '../lib/postcss.js'
+import postcss, {
+  Plugin,
+  Result,
+  Node,
+  Root,
+  parse,
+  PluginCreator
+} from '../lib/postcss.js'
 import LazyResult from '../lib/lazy-result.js'
 import Processor from '../lib/processor.js'
 
@@ -525,6 +532,17 @@ it('supports plugins returning processors', () => {
   let other: any = (postcss as any).plugin('test', () => {
     return new Processor([a])
   })
+  processor.use(other)
+  expect(processor.plugins).toEqual([a])
+})
+
+it('supports plugin creators returning processors', () => {
+  let a = () => {}
+  let processor = new Processor()
+  let other = (() => {
+    return new Processor([a])
+  }) as PluginCreator<void>
+  other.postcss = true
   processor.use(other)
   expect(processor.plugins).toEqual([a])
 })


ai:
> The issue is that in PostCSS 8, this does not notify visitor plugins, forcing these 'meta-plugins' to run as OnceExit

Am I right that PostCSS 8 do not support `meta-plugins` right now because of this bug?

ludofischer:
> Am I right that PostCSS 8 does not support `meta-plugins` right now because of this bug?

It kinds of support them if you always run the `meta-plugin` last.
Visitor-style plugins that are supposed to run after the `meta-plugin` don't get called. I think it's quite natural as starting a new traversal just ignores the listeners that have been registered on the original tree. In the cssnano case you would probably run every cssnano plugin on exit anyway, but I feel that in general you should not be forced to.

ai:
> Visitor-style plugins that are supposed to run after the meta-plugin don't get called.

It is a critical bug, which blocks us from promoting this API via types.

Do you want to fix this issue in this PR? We can announce `Processor`’s plugins support in 8.2 release planned for next week.

ludofischer:
> Do you want to fix this issue in this PR? We can announce `Processor`’s plugins support in 8.2 release planned for next week.

This PR won't help someone if they keep caling `process()` inside their plugin. It just changes the TypeScript type to let them export a Processor() with all their plugins preconfigured. But if you think think it's a good solution, I am happy if it gets released as 8.2.

ai:
Adding `Processor` returning to types means that we add it to the official API.

But we can’t do it if we know that there are critical bugs with this API.

ludofischer:
> But we can’t do it if we know that there are critical bugs with this API.

I don't think the API can introduce new critical bugs, as you can already pass a `Processor` as an `AcceptedPlugin`. postcss just takes the processor's plugin list and append it to the main plugin list. https://github.com/postcss/postcss/blob/1203260b757f145698829c7dd6977d79ffe41df2/lib/processor.js. So it boils down to calling `concat()` on the top plugins array.
 
But right now since you have to pass a `Processor` directly, you are forced to hard code the extra plugins. This pull request adds the ability use a creator function to add different plugins to the processor by reading some options or config.

A bug will appear if someone uses the new API but it turns out their plugins need a separate traversal to work, in that case they can still hack it with `OnceExit`.

ai:
> you can already pass a Processor as an AcceptedPlugin

Yeap. Old API (with `plugin.postcss = fn`) supported `Processor`. But new API was created without `Processor` returning support.

> A bug will appear if someone uses the new API but it turns out their plugins need a separate traversal to work, in that case they can still hack it with OnceExit.

Yeap, this is a problem. We can allow to return `Processor` from `plugin.postcss = true` plugins only if we will add visitor API plugins support for this use case.

ludofischer:
> Yeap. Old API (with plugin.postcss = fn) supported Processor. But new API was created without Processor returning support.

I thought even in PostCSS 8, you could officially pass a directly Processor as a plugin https://postcss.org/api/#acceptedplugin

> Yeap, this is a problem. We can allow to return Processor from plugin.postcss = true plugins only if we will add visitor API plugins support for this use case.

Visitor plugins should already support that in general, as passing a processor boils down to adding plugins to the main processor instance. It's calling `process()` a second time inside a plugin that blocks all visitor plugins running after. I've tested some possible plugin combinations here: https://github.com/ludofischer/postcss-strange-cases/blob/main/tests/metaPlugin.js 

The problem I can see with the API change in this pull request is that it might not be enough to stop calling `process()` inside postcss-modules, as it does more complicated plugin manipulation that I don't understand at the moment: https://github.com/css-modules/postcss-modules/blob/477486a3dc2ace2c10b8c574380303b06be0fec1/src/index.js#L80

ludofischer:
Here's how I see this pull request:
### Problem
You cannot build a meta-plugin that runs _before_ visitor plugins 
### Point in favor of this pull request
* this pull request would allow creating meta-plugins that run before other visitor plugins

### Points against this pull request:
* the only two meta-plugins I know are cssnano and postcss-modules and  it's fine to run them last, this pull request isn't strictly needed
* this pull request might not help refactoring postcss-modules anyway

ai:
Why do you want to avoid fixing visitor API in `Processor` returning?

For it is the best way.

ludofischer:
> Why do you want to avoid fixing visitor API in `Processor` returning?

I am not sure I understand. Could you make an example of a case that would require fixing?

ai:
> I am not sure I understand. Could you make an example of a case that would require fixing?

```js
let plugin = () => {
  return {
    Declaration: {
      color: decl => {
        decl.remove()
      }
    }
  }
}
plugin.postcss = true

meta = () => {
  return postcss([plugin])
}
meta.postcss = true
```

I understood you that visitor API plugins do not work in meta plugins with `Processor`

ludofischer:
> I understood you that visitor API plugins do not work in meta plugins with Processor

There's been a misunderstanding. Adding visitor plugins with a `Processor` works.
The issue is plugins calling `process()`:

```javascript
postcss([
{
  postcssPlugin: 'stop-visitors'
  Once(root, helpers) {
     helpers.postcss([...]).process(root) // This prevents visitor plugins coming after from running
 },
{
  postcssPlugin: 'visitor',
  Declaration(decl) { 
  /* this will never run */
  }
}
])
```
 I've even tested your example and it works: https://github.com/ludofischer/postcss-strange-cases/blob/f9e373eaaa071ce48f023c2834a117c46a38c387/tests/processors.js#L67

ai:
Oops. Got it.

Why does `postcss(extraPlugins).process()` not work? It should be use separated `LazyResult` and separated `Processor`.

ludofischer:
> Why does `postcss(extraPlugins).process()` not work? It should be use separated `LazyResult` and separated `Processor`

* don't work: _visitor_  plugins outside `extraPlugins`
* work: `extraPlugins` _and_ `Once*` plugins outside `extraPlugins` 

I am not sure why. I thought that since `postcss(extraPlugins).process()`  performs a separate traversal, observers registered on the original AST tree never get notified, so Visitor plugins behave as if they had disappeared. 
That would be why the workaround is:

```javascript
postcss([
{
  postcssPlugin: 'stop-visitors'
  OnceExit(root, helpers) { // changed from Once to OnceExit
     helpers.postcss([...]).process(root) // This prevents visitor plugins coming after from running, but now it runs last
 },                                                        
{
  postcssPlugin: 'visitor',
  Declaration(decl) { 
  /* this now runs because it's coming before the meta-plugin */
  }
}
])
```

ai:
Yeap. This is a rare bug we will fix it later.

Do you need types changes now or I can release them with the next changes?

ludofischer:
You can release them with the next changes.

ai:
Released in 8.2.2

PR #1486, opened 2020-12-08
stop adding newline before sourceMappingURL

milahu:
solve issue https://github.com/sveltejs/svelte-preprocess/issues/286

adding the newline breaks line numers, when the style is injected into a html document

since its only eye-candy, with zero function, i simply removed it

### DIFF
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -145,10 +145,7 @@ class MapGenerator {
       content = this.outputFile() + '.map'
     }
 
-    let eol = '\n'
-    if (this.css.includes('\r\n')) eol = '\r\n'
-
-    this.css += eol + '/*# sourceMappingURL=' + content + ' */'
+    this.css += '/*# sourceMappingURL=' + content + ' */'
   }
 
   outputFile () {


milahu:
tests are expecting the newline ..

ai:
I do not like that we reduce readability and afraid of breaking changes.

Can you explain why this `\n` is breaking Svelte?

P.S. I am a big fan of Svelte and will be glad to help, but I need to be very careful with any changes in PostCSS

milahu:
> Can you explain why this `\n` is breaking Svelte?

as i said ..

> adding the newline breaks line numers, when the style is injected into a html document

usually the style is injected at the start of a html file
so all following line numers are off by one

> readability

overrated

ai:
> adding the newline breaks line numers, when the style is injected into a html document

I am still missing the connection between the extra line inside `<style>` and the broken source map.

Are we talking about Svelte component source maps? PostCSS is adding many new lines in CSS. Why adding comment break source maps?

milahu:
> PostCSS is adding many new lines in CSS. Why adding comment break source maps?

cos that last change is not tracked in the sourcemap

milahu:
> afraid of breaking changes.

do other tools expect the `/*#` at start of line?

when i fix the tests, this one is still failing

```js
it('reads only the last map from annotation', () => {
  let file = join(dir, 'c.map')
  outputFileSync(file, map)
  let root = parse(
    'a{}' +
      '/*# sourceMappingURL=a.map */' +
      '/*# sourceMappingURL=b.map */' +
      '/*# sourceMappingURL=c.map */',
    { from: file }
  )

  expect(root.source?.input.map.text).toEqual(map)
  // TypeError: Cannot read property 'text' of undefined

  expect(root.source?.input.map.root).toEqual(dir)
})
```

ai:
> do other tools expect the /*# at start of line?

I am not sure. But with 125M downloads, everything could happen. This is why I need a very good reason or some research to change the current behavior.

milahu:
> if you really want to keep the newline
> you could try adding that change to the sourcemap

this could work

<details>

```js
  addAnnotation () {
    let content

    this.map.mappings.push([], []);
    // add two unmapped lines to sourcemap
    // for the /*# */ special comment at end-of-file

    if (this.isInline()) {
      content =
        'data:application/json;base64,' + this.toBase64(this.map.toString())
    } else if (typeof this.mapOpts.annotation === 'string') {
      content = this.mapOpts.annotation
    } else if (typeof this.mapOpts.annotation === 'function') {
      content = this.mapOpts.annotation(this.opts.to, this.root)
    } else {
      content = this.outputFile() + '.map'
    }

    let eol = '\n'
    if (this.css.includes('\r\n')) eol = '\r\n'

    //this.css += eol + '/*# sourceMappingURL=' + content + ' */';
    this.css += eol + '/*# sourceMappingURL=' + content + ' */' + eol;
    // add extra end-of-line to avoid column offsets
  }
```

</code>

ai:
Yeap. Extra newline at the end of the file is much safer.

BTW, maybe we should add it in `svelte-preprocess`?

milahu:
> maybe we should add it in `svelte-preprocess`?

what exactly?

ai:
> what exactly?

Add extra `\n` after PostCSS output

milahu:
solved in svelte

"blame the consumer" is right in this case:
seems to be a de-facto-standard for code-transformers
to append 'annotation' comments, which are not mapped in the sourcemap

then its the consumer's job to

* either keep the annotation comments and add unmapped lines to the sourcemap
* or keep the sourcemap and remove annoation comments

in both cases, source and sourcemap should have the same number of lines

in a perfect world, we would have a better interface
to pass code and sourcemap in separate variables
and also allow to pass [decoded sourcemaps](https://github.com/Rich-Harris/sourcemap-codec/issues/82)
to avoid the unnecessary encode + decode steps

but sadly, the sourcemap spec writers did not cover this scenario
so the common denominator is the "simple but slow" encoded-sourcemap interface

PR #1487, opened 2020-12-08
map all lines in sourcemap, also annotation-comment lines

milahu:
replace #1486 

* add newline after annotation comment `/*# sourceMappingURL=.... */`
* add two unmapped lines to sourcemap

### DIFF
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -134,6 +134,20 @@ class MapGenerator {
   addAnnotation () {
     let content
 
+    // append two unmapped lines to sourcemap
+    // for the /*# */ annotation comment at end-of-file
+    let numLines = (this.css.match(/\n/g) || []).length + 1
+    this.map.addMapping({
+      source: null,
+      original: null,
+      generated: { line: numLines + 1, column: 0 }
+    })
+    this.map.addMapping({
+      source: null,
+      original: null,
+      generated: { line: numLines + 2, column: 0 }
+    })
+
     if (this.isInline()) {
       content =
         'data:application/json;base64,' + this.toBase64(this.map.toString())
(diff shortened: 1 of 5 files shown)


milahu:
as expected, one test is failing

this is a bug in source-map@0.6.1 and its lib/source-map-consumer.js

```js
SourceMapConsumer.prototype.eachMapping =
// ....
source = util.computeSourceURL(sourceRoot, source, this._sourceMapURL);
```
here, source changes from `null` to `'.'`
and further downstream, that mapping is handled as an 'original mapping', but

> original.line and original.column are not numbers -- you probably meant to omit the original mapping entirely and only map the generated position. If so, pass null for the original mapping instead of an object with empty or null values.

i tried to 'simply upgrade' source-map, but that breaks even more ..

ai:
Yeap. New `source-map` requires async API. I thought how we can use it and didn't find a way without big breaking changes for PostCSS API.

milahu:
> New `source-map` requires async API. I thought how we can use it

ideally, not at all

source-map and mozilla in general are abandoned, since google stopped funding ..

the problems are known [here](https://github.com/mozilla/source-map/issues/331) and [here](https://github.com/mozilla/source-map/issues/370)
two related npm packages are dead [here](https://www.npmjs.com/package/@snek/source-map-support) and [here](https://www.npmjs.com/package/@contrast/synchronous-source-maps)

one patched fork is [here](https://github.com/mozilla/source-map/compare/0.6.1...benthemonkey:patch-0.6.1) (via issue 370)

looks like 'we' need to ....

make our own npm package, without mozilla

(the crowd is shocked)

ai:
Closing since test was not fixed

PR #1488, opened 2020-12-08
Fixup JSON AST conversion

mischnic:
I've found two problems related to `Input`:

1std and 2nd commit: I missed that all nodes have a `source` property, not just the root (this caused stringification to fail)

3rd commit: More of an optimization: because of 1), the source contents & map would be stored once for every node (!). This has something like a quadratic memory usage because larger CSS files have more files, so with 10x more rules, the CSS source is about 10x larger and is stored 10x times.
I've instead deduplicated these inputs objects by replacing `node.source.input` with a number in the `toJSON` output and including a `_inputs` array which maps number to input object (because most of these are `===`-equal). This is then reversed in fromJSON. This breaks a buch of `postcss-parser-tests` tests at the moment though. What do you think of this approach? (Before I try to adjust the parser test runner to accept this version...)

### DIFF
--- a/lib/fromJSON.js
+++ b/lib/fromJSON.js
@@ -9,23 +9,30 @@ let Root = require('./root')
 let Rule = require('./rule')
 
 function fromJSON (json, inputs) {
-  let { _inputs, ...defaults } = json
-  inputs = _inputs || inputs
+  let { inputs: ownInputs, ...defaults } = json
+  if (ownInputs) {
+    inputs = []
+    for (let input of ownInputs) {
+      let inputHydrated = { ...input }
+      if (inputHydrated.map) {
+        inputHydrated.map = {
+          ...inputHydrated.map,
+          __proto__: PreviousMap.prototype
+        }
+      }
+      inputs.push(inputHydrated)
+    }
+  }
   if (defaults.nodes) {
     defaults.nodes = json.nodes.map(n => fromJSON(n, inputs))
   }
   if (defaults.source) {
-    defaults.source = { ...defaults.source }
-    if (defaults.source.input != null) {
+    let { inputId, ...source } = defaults.source
+    defaults.source = source
+    if (inputId != null) {
       defaults.source.input = {
-        ...inputs[defaults.source.input],
+        ...inputs[inputId],
         __proto__: Input.prototype
-      }
-      if (defaults.source.input.map) {
-        defaults.source.input.map = {
-          ...defaults.source.input.map,
-          __proto__: PreviousMap.prototype
-        }
       }
     }
   }
--- a/lib/fromJSON.js
+++ b/lib/fromJSON.js
@@ -13,7 +13,7 @@ function fromJSON (json, inputs) {
   if (ownInputs) {
     inputs = []
     for (let input of ownInputs) {
-      let inputHydrated = { ...input }
+      let inputHydrated = { ...input, __proto__: Input.prototype }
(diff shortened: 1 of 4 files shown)


lib/node.js
```diff
@@ -166,8 +166,11 @@ class Node {
     if (!keepBetween) delete this.raws.between
   }
 
-  toJSON () {
+  toJSON (inputs) {
     let fixed = {}
+    let emitInputs = inputs == null
+    inputs = inputs || new Map()
```
ai, line 172: Why do we use `Map` if we plan to stringify it to JSON?

lib/fromJSON.js
```diff
@@ -8,16 +8,17 @@ let Input = require('./input')
 let Root = require('./root')
 let Rule = require('./rule')
 
-function fromJSON (json) {
-  let defaults = { ...json }
-  if (json.nodes) {
-    defaults.nodes = json.nodes.map(i => fromJSON(i))
+function fromJSON (json, inputs) {
+  let { _inputs, ...defaults } = json
```
ai, line 12: Let’s rename `_inputs` to `inputs`. We do not use `_*` in PostCSS.

lib/node.js
```diff
@@ -180,23 +183,33 @@ class Node {
...
       } else if (name === 'source') {
+        let input = inputs.get(value.input)
+        if (input == null) {
+          input = inputsNextIndex
+          inputs.set(value.input, inputsNextIndex)
+          inputsNextIndex++
+        }
         fixed[name] = {
-          input: value.input.toJSON(),
+          input,
```
ai, line 201: Let’s change the property name if we changed the content. `inputId` for instance.

ai:
[`postcss-parser-tests/jsonify.js`](https://github.com/postcss/postcss-parser-tests/blob/main/jsonify.js) is the best place to fix tests

lib/node.js
mischnic: To be able to easily deduplicate them. The alternative (making this an array and using `indexOf` instead of `get`) is probably slower.
At the end, the keys of the Map are converted into an array and added to the JSON result object.

ai: You can use `{}` object.

```js
if (!root.inputs[id]) {
  root.inputs[id] = input
}
```

mischnic: The Map contains a mapping `node.input object -> id in the inputs array`. That wouldn't work with an object.
(`inputs.get(node.input)` returns the id if this `Input` object was already encountered, otherwise it's added)

ai: Got it. Yeap, you are right.

mischnic:
> postcss-parser-tests/jsonify.js is the best place to fix tests

Should that function also delete the `inputId` and `inputs` properties?

ai:
> Should that function also delete the inputId and inputs properties?

Yeap. Send PR. I will release it.

ai:
Are we ready to release it?

mischnic:
I've just tested this in my project and it seems to work now

ai:
BTW, why do you need JSON AST rehydration?

mischnic:
For Parcel, I want to send ASTs across worker threads/cache them to disk.
(In this case, for CSS modules tree shaking: https://github.com/parcel-bundler/parcel/pull/5363. Previously, we just stringified the AST before that).

Thanks you so much for the fast process!

ai:
Released in 8.2.1

PR #1495, opened 2020-12-29
remove root assignment in stringifier

### LINKED ISSUE #1494: stringifer has problems with tailwind roots
I'm currently working my way through postcss to try find a solution to transforming CSS literals in JS.

@ai helpfully pointed me in the right direction i think, to use postcss-jsx. so here's what im using:

* postcss
* postcss-jsx
* tailwind
* postcss-cli

My config:

```ts
module.exports = {
  syntax: require('postcss-syntax'),
  plugins: {
    tailwindcss: {}
  }
};
```

My sources:

```ts
const tpl = css`@tailwind base;`;
```

I now get this error:

```
TypeError: this[node.type] is not a function
```

I tracked this down and it looks to be a genuine bug:

https://github.com/postcss/postcss/blob/b6c2f95fa817db962118a8117d77d41c5f20c814/lib/stringifier.js#L40-L44

We overwrite `this.root` where `this` is a `Stringifer`. meaning the `root` method won't exist anymore.

i think this is probably an edge case and has only been noticed now because tailwind seems to produce a `Root` inside a `Document`... both of which have `type: "root"`.

Tailwind seems to produce:

```
Document {
  type: "root",
  nodes: [
    Root {
      type: "root"
    }
  ]
}
```

no clue if this is also questionable, two node types having the same type. and im not sure if postcss-jsx is responsible for making it or tailwind is...

EDIT:

looks like `Document` belongs to postcss-syntax, and `Root` to postcss.

43081j:
Fixes #1494 

I had a search around your codebase and it doesn't look like anything used `Stringifier#root` as a property. so i just removed the line. if you'd rather have it renamed, let me know

it was overwriting the `root(node)` method previously, which was clearly wrong and caused issues with nested roots.

cc @ai

### DIFF
--- a/lib/stringifier.js
+++ b/lib/stringifier.js
@@ -38,7 +38,6 @@ class Stringifier {
   }
 
   root (node) {
-    this.root = node
     this.body(node)
     if (node.raws.after) this.builder(node.raws.after)
   }
--- a/test/stringifier.test.js
+++ b/test/stringifier.test.js
@@ -163,3 +163,12 @@ it('uses optional raws.indent', () => {
   rule.append({ prop: 'color', value: 'black' })
   expect(rule.toString()).toEqual('a {\n color: black\n}')
 })
+
+it('handles nested roots', () => {
+  let root = new Root()
+  let subRoot = new Root()
+  subRoot.append(new AtRule({ name: 'foo' }))
+  root.append(subRoot)
+
+  expect(root.toString()).toEqual('@foo')
+})


nesara-cmd:
@43081j

ai:
Thanks for the deep investigation.

Yeap, it was just a typo, which causes a bug in a rare case.

I will release it a few hours later after lunch.

43081j:
Thanks a lot for the quick turnaround. This'll allow us to use postcss inside web components a lot easier, so I'm sure it'll make plenty of people happy.

ai:

43081j:
Sure, I've been tweeting about this stuff for a few days now myself haha. You can find me at @43081j the same as here

ai:
Released in 8.2.2

PR #1507, opened 2021-01-08
fix: add plugin name to deprecation message

Alphy11:
Adding the plugin name to the deprecation message to help people debug which plugin needs to be upgraded.  
Translated string is from my friend who can read/write Chinese but only conversation, doesn't use it on a technical basis, so a proofread would be appreciated.


ai:
Nice idea. Thanks. Released in 8.2.4.

PR #1523, opened 2021-02-06
Don't split on escaped characters in list.split()

nex3:
I ran into this in practice when cssnano converted a quoted attribute
selector (which contained a comma) to use backslash escapes instead.


ai:
Thanks. Released in 8.2.5.

Issue #1455, opened 2020-11-05
DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at G:\SNWS\sn-micro-front-web-project-template\node_modules\postcss\package.json.


LP1994:
This is not a error!!!

operating system: Windows 10,
node: 15.1.0,
npm: 6.14.8,
webpack: 5.4.0,
postcss: 8.1.4

This is only a warning issued by "nodejs". It is recommended to make corrections according to the warning, because some writing methods are not recommended.

![image](https://user-images.githubusercontent.com/12012671/98194446-68c21680-1f5a-11eb-8624-2025f06536c6.png)

ShenHongFei:
fixed in this PR
https://github.com/postcss/postcss/pull/1456

ai:
The fix was released in 8.1.5.

ai:
@ShenHongFei just replacing `./` → `./*` cause an issue in Node.js 12.

Can we fix this warning by adding `./` and `./*` together in `package.json→exports`?

ShenHongFei:
> @ShenHongFei just replacing `./` → `./*` cause an issue in Node.js 12.
> 
> Can we fix this warning by adding `./` and `./*` together in `package.json→exports`?

Yes, that works.

ai:
How I can see this warning?

ai:
@LP1994 what do I need to call to see this warning?

LP1994:
> @LP1994 what do I need to call to see this warning?

These warnings appear only after upgrading to the following versions:
"postcss": "8.1.6",
"webpack": "5.4.0",
"npm": "6.14.8",
"node": "15.1.0",
operating system: Windows 10

ai:
Another way to fix the problem https://github.com/postcss/postcss/commit/92155c73e105a36af5196a75bcc34ada377a711f

ai:
Nope, this way again kills Node.js 15 in CommonJS.

Can you create a small test repo?

LP1994:
> Nope, this way again kills Node.js 15 in CommonJS.
> 
> Can you create a small test repo?
Later, I'll try.

borodadada:
Hello. I have DeprecationWarning

(node:29624) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at D:\js\project\web\node_modules\css-loader\node_modules\postcss\package.json.
[...]

PR #1534, opened 2021-02-23
Update deprecated subpath pattern to "./*"

CarlVitasa:
Fixes the following NPM error:

```
ERROR  (node:14100) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at 
... node_modules\postcss\package.json.
Update this package.json to use a subpath pattern like "./*".
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
       "import": "./lib/postcss.mjs",
       "types": "./lib/postcss.d.ts"
     },
-    "./": "./"
+    "./*": "./*"
   },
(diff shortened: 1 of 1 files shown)


Semigradsky:
@CarlVitasa just update your NodeJS. See #1455

ai:
These changes will break PostCSS in the older Node.js version.

This warning is just a bug from old Node.js 15.x versions.

PR #1536, opened 2021-03-03
Ignore path module for browser builds

### LINKED ISSUE #1509: Can't resolve 'path'
I got an error with Angular 11 project
```
Error: ./node_modules/sanitize-html/node_modules/postcss/lib/input.js
Module not found: Error: Can't resolve 'path' in '/Users/mkrajcir/Devel/BMB/eshop/node_modules/sanitize-html/node_modules/postcss/lib'

Error: ./node_modules/sanitize-html/node_modules/postcss/lib/map-generator.js
Module not found: Error: Can't resolve 'path' in '/Users/mkrajcir/Devel/BMB/eshop/node_modules/sanitize-html/node_modules/postcss/lib'

Error: ./node_modules/sanitize-html/node_modules/postcss/lib/previous-map.js
Module not found: Error: Can't resolve 'path' in '/Users/mkrajcir/Devel/BMB/eshop/node_modules/sanitize-html/node_modules/postcss/lib'
```

mattcompiles:
Compiling `postcss` in webpack 5 will cause an error due to the removal of automatic node built-in polyfills. 

This change tells webpack to ignore the `path` module when compiling, the same treatment `fs` already receives. I have successfully run `postcss` in webpack 5 with this change. 

This issue was raised in #1509
Resolves #1509

Example error:
```
ERROR in ./node_modules/postcss/lib/previous-map.js 4:24-39
Module not found: Error: Can't resolve 'path' in '/Users/person/some-project/node_modules/postcss/lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,8 @@
   "browser": {
     "./lib/terminal-highlight": false,
     "colorette": false,
-    "fs": false
+    "fs": false,
+    "path": false
   },
   "size-limit": [
     {


ai:
Do you test this patch in webpack 5?

mattcompiles:
I haven't written an automated test for `postcss`. But I have manually verified that it works. 

I have this code running in browser with webpack 5.
```
const result = postcss().process(flattenedCss, {
    parser: postcssJs,
    from: undefined,
  });
```

ai:
> I have manually verified that it works.

Thanks enough. Thanks. I will release it today.

Do you have a Twitter account to mention you in the release tweet?

mattcompiles:
Thanks mate. Twitter is `@mattcompiles` 👍

ai:
Released in 8.2.7.

charlessuh:
I'm not sure this is a complete fix -- and it breaks the compile in webpack 4 where it worked previously because of the automatically included polyfill.

When `path` is set to false when compiling for the web / browser (see https://github.com/defunctzombie/package-browser-field-spec#ignore-a-module), the `path` object is null. Multiple places in the `postcss` call does not actually check that `path` exists before calling methods on it -- for example, the `isAbsolute` function at https://github.com/postcss/postcss/blob/8.2.7/lib/input.js#L4

I think the complete fix is either:

1) make sure your webpack 5 config polyfills `path` so code like postcss assumes it exists.
2) update postcss to polyfill `path` itself.

ai:
Can somebody send a PR that will test `path` as we do with `fs` (to support webpack 4)

ai:
@mattcompiles can you fix your changes for webpack 4?

mattcompiles:
Had a quick look but I'm not seeing anywhere in the code where `fs` is checked for reference. The only file I see using `fs` is `previous-map.js` and it just uses `fs` functions freely, just like `path`.

@ai How would you like to handle the case where `path` isn't available? I see that it's mostly source map related code. Does that make sense in a browser anyway? 🤔

ai:
@mattcompiles you need to add `if (path)` before using it and fix tests to use real `from`

charlessuh:
I'll let @mattcompiles respond, but my 2 cents is the easiest fix would be to just revert this PR for now.

For webpack 5, if you look at the full message, you can have `path` be polyfilled to `path-browserify` using one or two extra lines of config:

```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add an fallback 'resolve.fallback: { "path": "path-browserify" }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }
```

The ultimate result of this is (whether you're using webpack 4 with the automatic config for this, or webpack 5 where you have to opt-in to this) is that when webpack is compiling to a node target, code that imports `path` is just left alone (since node has `path`) but if webpack is compiling for the browser,  an import to `path` is replaced by an import to `path-browserify`.

I guess the longer-term question for you as the `postcss` project is whether you want to more explicitly support browsers and handle cases like this yourself (which is what you are saying you will do by adding `browsers: { path: false }` to your package.json) or rely on upstream bundlers to handle this for you.

mattcompiles:
@charlessuh My issue with that approach is you're assuming the package using `postcss` has control over the webpack config. Happy if you want to revert for now, but I'd prefer to work on a proper fix to make working in the browser a first class option. 

@ai Happy to make that change but it'll mean all code relying on `path` will potentially be broken. Not sure if that's a good result for all consumers?

ai:
> it'll mean all code relying on path will potentially be broken. Not sure if that's a good result for all consumers?

Can you explain why? Maybe I am missing something, but for my side we are talking just about adding extra `if` condition

mattcompiles:
Well it seems the `path` module is mostly used for handling source maps in the code. Do we expect source maps to be generated when running in the browser? Take the following code for example. 

```js
// Example from "input.js"
if (opts.from) {
  if (/^\w+:\/\//.test(opts.from) || isAbsolute(opts.from)) {
    this.file = opts.from
  } else {
   	this.file = resolve(opts.from)
  }
}
```

What would you set `this.file` to if `isAbsolute` and `resolve` aren't available? This is just one example. 

An alternative approach here would be to change the `"browser": { "path": false }` to `"browser": { "path": "path-browserify" }` inside `postcss` itself. Saving consumers from the hassle. Looking at [bundlephobia](https://bundlephobia.com/result?p=path-browserify@1.0.1) the package isn't that large and wouldn't add a huge amount to the `postcss` build.

ai:
> Do we expect source maps to be generated when running in the browser?

Nope

> What would you set this.file to if isAbsolute and resolve aren't available?

```
this.file = opts.from
```

> An alternative approach here would be to change the "browser": { "path": false } to "browser": { "path": "path-browserify" } inside postcss itself. Saving consumers from the hassle.

I would like to have a webpack support without a need to install anything additional for the user and without adding a new dependency to `postcss` package.

mattcompiles:
OK, I'll have a crack at this later today.

mattcompiles:
Just opened #1538 which hopefully resolves this.

PR #1538, opened 2021-03-09
Handle empty path object in browser builds

mattcompiles:
I've also fixed the browser tests as they were not mocking correctly. Calling `doMock` after importing `postcss` had no effect as the module had already been processed before the mock takes effect.


ai:
Looks good. Can you just call `npx eslint --fix .` to fix code format?

mattcompiles:
> Looks good. Can you just call `npx eslint --fix .` to fix code format?

👍

ai:
Thanks. Released in 8.2.8.

PR #1547, opened 2021-03-26
Add postcss-multi-value-display to plugins.md

jake-low:
I just published a new plugin, [`postcss-multi-value-display`](https://github.com/jake-low/postcss-multi-value-display), which transforms [CSS Display Module Level 3](https://www.w3.org/TR/css-display-3/) multi-value syntax into legacy single-value equivalents when such equivalents exist. The plugin can be installed [from NPM](https://www.npmjs.com/package/postcss-multi-value-display).

This PR adds a link and short description of that plugin to postcss/docs/plugins.md. I also fixed a couple of alphabetization issues in the "Future CSS Syntax" section (though looking again I see that I didn't catch them all).


ai:
Looks cool!

@jonathantneal should we add it to `postcss-preset-env`?

PR #1549, opened 2021-03-30
Export types for NodeErrorOptions

realityking:
This will simplify the types import for `postcss-values-parser`: https://github.com/shellscape/postcss-values-parser/pull/132


ai:
Released in 8.2.9.

PR #1552, opened 2021-04-06
Fix import url for Deno example

roelandmoors:
The current Deno example did't work. I created [this issue](https://github.com/postcss/autoprefixer/issues/1402).

Here is an announcement about the url change: https://jspm.org/jspm-dev-release


ai:
Thanks

PR #1553, opened 2021-04-08
Fix browser bundling with webpack 5

barak007:
I think that the browser field is missing `url: false`.
I'm not completely sure that this is the right fix and also the npm package `url` does not implement the APIs used by postcss. 

steps to reproduce

```
mkdir postcss-browser 
cd postcss-browser 
mkdir src
echo 'import "postcss";' > src/index.js
npm init -y
npm install webpack webpack-cli postcss
npx webpack
```
This will emit error: 
```
ERROR in ./node_modules/postcss/lib/input.js 3:39-53
Module not found: Error: Can't resolve 'url' in 'C:\projects\postcss-web-bundle\postcss-browser\node_modules\postcss\lib'
```

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -130,7 +130,9 @@ class Input {
 
     result.input = { line, column, source: this.css }
     if (this.file) {
-      result.input.url = pathToFileURL(this.file).toString()
+      if(pathToFileURL) {
+        result.input.url = pathToFileURL(this.file).toString()
+      }
       result.input.file = this.file
     }
 
@@ -162,7 +164,11 @@ class Input {
     }
 
     if (fromUrl.protocol === 'file:') {
-      result.file = fileURLToPath(fromUrl)
+      if(fileURLToPath){
+        result.file = fileURLToPath(fromUrl)
+      } else {
+        throw new Error(`file: protocol is not available in this postcss build`);
+      }
     }
 
     let source = consumer.sourceContentFor(from.source)
--- a/lib/input.js
+++ b/lib/input.js
@@ -130,7 +130,7 @@ class Input {
 
     result.input = { line, column, source: this.css }
     if (this.file) {
-      if(pathToFileURL) {
+      if (pathToFileURL) {
         result.input.url = pathToFileURL(this.file).toString()
       }
       result.input.file = this.file
@@ -164,10 +164,10 @@ class Input {
     }
 
     if (fromUrl.protocol === 'file:') {
-      if(fileURLToPath){
+      if (fileURLToPath) {
         result.file = fileURLToPath(fromUrl)
       } else {
-        throw new Error(`file: protocol is not available in this postcss build`);
+        throw new Error(`file: protocol is not available in this PostCSS build`);
       }
     }
 
--- a/lib/input.js
+++ b/lib/input.js
@@ -167,6 +167,7 @@ class Input {
       if (fileURLToPath) {
         result.file = fileURLToPath(fromUrl)
       } else {
+        // istanbul ignore next
         throw new Error(`file: protocol is not available in this PostCSS build`);
       }
     }
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -4,7 +4,7 @@ let { dirname, resolve, relative, sep } = require('path')
 let { pathToFileURL } = require('url')
 let mozilla = require('source-map')
 
-let pathAvailable = Boolean(dirname, resolve, relative, sep)
+let pathAvailable = Boolean(dirname && resolve && relative && sep)
 
 class MapGenerator {
   constructor(stringify, root, opts) {
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -202,10 +202,10 @@ class MapGenerator {
     if (this.mapOpts.from) {
       return this.toUrl(this.mapOpts.from)
     } else if (this.mapOpts.absolute) {
-      if(pathToFileURL) {
+      if (pathToFileURL) {
         return pathToFileURL(node.source.input.from).toString()
       } else {
-        throw new Error('map option "absolute" is not available in this postcss build')
+        throw new Error('`map.absolute` option is not available in this PostCSS build')
       }
     } else {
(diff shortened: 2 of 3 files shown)


ai:
Can I ask you to test this branch in webpack 4 and webpack 5?

barak007:
I didn't. I will do later today. But I guess that any way the webpack 4 built-in polyfill is also wrong since it's based on the same url polyfill.

ai:
`"url": false` will force `require('url')` to return empty object `{}`. It could break source map generation.

ai:
It will be breaking changes

barak007:
why?

ai:
Because we will break webpack 4 build. We need a solution to support both bundlers without config changes.

barak007:
At the current state of things:
- webpack@4 users get a partial polyfill that does not include the used functions. I suspect the only reason it didn't throw at runtime is because that branch in the flow was never called by a consumer using it inside a browser.
- webpack@5 users either get an error of missing "url" or implicitly get the same partial polyfill if it's included in node_modules by another dependency.

I propose one of two possible solutions:

1. Include a small (posix only) polyfill for the two used functions when running in the browser. webpack@4 and webpack@5 users will both be able to bundle postcss, with the code getting the correct polyfilled functions at runtime. 
2. Map `url` to false and protect the flow that uses those functions.

Assuming the above is correct, either approach will _probably_ not be a breaking change, but a bugfix.

ai:
> Include a small (posix only) polyfill for the two used functions

I like the idea

barak007:
After deep investigation I notice few things:

Source map generation is disabled when bundling for browser since path is mapped to false (this is why it didn't break) https://github.com/postcss/postcss/blob/8682b1e4e328432ba692bed52326e84439cec9e4/lib/map-generator.js#L280   
And the other two locations that uses `url` are not really that important 
1. When the `from` is file protocol
https://github.com/postcss/postcss/blob/8682b1e4e328432ba692bed52326e84439cec9e4/lib/input.js#L164
2. When throwing an error and annotating the results with the url
https://github.com/postcss/postcss/blob/8682b1e4e328432ba692bed52326e84439cec9e4/lib/input.js#L133

I did managed to create the polyfill but I'm now thinking maybe disable these features will be good also.

If not maybe we should consider to also to implement the path behavior required for the entire code to work. (path.resolve is already in-lined inside my url polyfill)

Anyhow I will probably update the PR tomorrow.

ai:
Yeap, disabling this features will be the better option.

lib/map-generator.js
```diff
@@ -4,7 +4,7 @@ let { dirname, resolve, relative, sep } = require('path')
 let { pathToFileURL } = require('url')
 let mozilla = require('source-map')
 
-let pathAvailable = Boolean(dirname, resolve, relative, sep)
+let pathAvailable = Boolean(dirname && resolve && relative && sep)
```
barak007, line 7: Boolean only accept one argument

lib/input.js
```diff
@@ -162,7 +164,11 @@ class Input {
     }
 
     if (fromUrl.protocol === 'file:') {
-      result.file = fileURLToPath(fromUrl)
+      if(fileURLToPath){
+        result.file = fileURLToPath(fromUrl)
+      } else {
+        throw new Error(`file: protocol is not available in this postcss build`);
```
barak007, line 170: Wasn't sure about the error

lib/map-generator.js
```diff
@@ -202,7 +202,11 @@ class MapGenerator {
     if (this.mapOpts.from) {
       return this.toUrl(this.mapOpts.from)
     } else if (this.mapOpts.absolute) {
-      return pathToFileURL(node.source.input.from).toString()
+      if(pathToFileURL) {
+        return pathToFileURL(node.source.input.from).toString()
+      } else {
+        throw new Error('map option "absolute" is not available in this postcss build')
```
barak007, line 208: Wasn't sure about the error

lib/input.js
```diff
@@ -130,7 +130,9 @@ class Input {
 
     result.input = { line, column, source: this.css }
     if (this.file) {
-      result.input.url = pathToFileURL(this.file).toString()
+      if(pathToFileURL) {
```
barak007, line 133: Wasn't sure about the behavior

barak007:
It turned out more "complex" for me to remove these features because I wasn't sure how to handle the error cases.

ai:
Can you put `/* istanbul ignore next */` to pass `npx jest --coverage`?

lib/map-generator.js
ai: Looks good. Let’s just clean the text:

```suggestion
        throw new Error('`map.absolute` option is not available in this PostCSS build')
```

lib/input.js
ai:
```suggestion
        throw new Error(`file: protocol is not available in this PostCSS build`);
```

lib/input.js
ai:
```suggestion
      if (pathToFileURL) {
```

lib/input.js
```diff
@@ -162,7 +164,11 @@ class Input {
     }
 
     if (fromUrl.protocol === 'file:') {
-      result.file = fileURLToPath(fromUrl)
+      if(fileURLToPath){
```
ai, line 167:
```suggestion
      if (fileURLToPath){
```

lib/map-generator.js
```diff
@@ -202,7 +202,11 @@ class MapGenerator {
     if (this.mapOpts.from) {
       return this.toUrl(this.mapOpts.from)
     } else if (this.mapOpts.absolute) {
-      return pathToFileURL(node.source.input.from).toString()
+      if(pathToFileURL) {
```
ai, line 205:
```suggestion
      if (pathToFileURL) {
```

ai:
Looks good. Thanks.

I will release it next week since I need to take deal with another issue.

barak007:
Thanks.

ai:
Do you have Twitter to mention you in the release tweet?

barak007:
No Twitter 10x.

PR #1567, opened 2021-04-26
Fix ReDoS in previous-map

yetingli:
Fix Strategy: Replace `(.*)` with `(?:(?!sourceMappingURL=).)*`

### DIFF
--- a/lib/previous-map.js
+++ b/lib/previous-map.js
@@ -48,11 +48,11 @@ class PreviousMap {
   }
 
   getAnnotationURL(sourceMapString) {
-    return sourceMapString.match(/\/\*\s*# sourceMappingURL=(.*)\*\//)[1].trim()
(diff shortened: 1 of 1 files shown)


ai:
Thanks. Released in 8.2.13.

AndrewRayCode:
I'm looking at [snyk](https://snyk.io/vuln/SNYK-JS-POSTCSS-1255640) and I don't understand how this would be exploited. Would you somehow need to construct a malicious source map (?) and inject it (?) into someone's build pipeline that uses postcss?

ai:
@AndrewRayCode yeap. It could be used on services like CodePen, when users use servers to compile user’s CSS.

And it is not a real vulnerability. But a way to increase used resources. Even without fixing PostCSS you can add extra timeout check and kill commands which takes too much time.

PR #1571, opened 2021-05-05
Remove source-map from browser build

barak007:
Since `path` is false in browsers build the `source-map` package is actually not in use*.  

This PR adds `"source-map": false` to the `package.json` also added explicit protection for the existence of the `source-map` API's.
Now if one will want to use postcss in browser and support source maps he will need to include `path` and `source-map` packages

This will decreased the bundle size of postcss by ~27% (from 29KB to 21KB)

(*) There is only one feature that I wasn't sure what to do with related to `fromJSON`. It is using the `PreviousMap` (no problem there) and `PreviousMap` uses the `source-map` package and I wasn't sure how to fallback. It might be a super edge case for browsers.

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -2,6 +2,7 @@
 
 let { fileURLToPath, pathToFileURL } = require('url')
 let { resolve, isAbsolute } = require('path')
+let { SourceMapConsumer, SourceMapGenerator } = require('source-map')
 let { nanoid } = require('nanoid/non-secure')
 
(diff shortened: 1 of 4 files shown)


ai:
`fromJSON` is a rare feature (I think only Parcel i using it).

ai:

barak007:
No Twitter. maybe I should open one

ai:
Thanks. Released in 8.2.14.

PR #1580, opened 2021-05-13
Add `dir-dependency` documentation

bradlc:
As discussed in #1537 this pull request adds documentation for a new `dir-dependency` message type that can be used to signify that a plugin depends on the contents of a directory. This would be used in the same way as the `dependency` message but with a `dir` property instead of a `file` property:

```js
result.messages.push({
  type: 'dir-dependency',
  plugin: 'postcss-import',
  dir: '/js',
  parent: result.opts.from
})
```

Runners should watch this directory for changes and rebuild the CSS when something changes.

This PR also adds some documentation for the existing `dependency` message to the plugin and runner guides, and clarifies the usage of dependency messages in the "Writing a PostCSS Plugin" documentation.

### DIFF
--- a/docs/guidelines/plugin.md
+++ b/docs/guidelines/plugin.md
@@ -217,7 +217,7 @@ a `dependency` message to the `result`:
 result.messages.push({
   type: 'dependency',
   plugin: 'postcss-import',
-  file: '/js/example.js',
+  file: '/imported/file.css',
   parent: result.opts.from
 })
 ```
--- a/docs/guidelines/plugin.md
+++ b/docs/guidelines/plugin.md
@@ -208,9 +208,9 @@ is described in [API docs].
 
 ## 3. Dependencies
 
-### 3.1. Use the `dependency` message type for file dependencies
+### 3.1. Use messages to specify dependencies
 
-If a plugin depends on another file, it should be declared by attaching
+If a plugin depends on another file, it should be specified by attaching
 a `dependency` message to the `result`:
 
 ```js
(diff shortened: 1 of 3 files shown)


mischnic:
Is a `dir-dependency` recursive (= are changes in subdirectories also watched)?

bradlc:
> Is a `dir-dependency` recursive (= are changes in subdirectories also watched)?

I would say so, yes. Do you think the documentation should mention that?

ai:
> Do you think the documentation should mention that?

It is better to mention.

docs/guidelines/plugin.md
```diff
@@ -206,9 +206,41 @@ is described in [API docs].
 [API docs]: https://postcss.org/api/
 
 
-## 3. Errors
+## 3. Dependencies
```
ai, line 209: We need to mention it in runner guidelines

https://github.com/postcss/postcss/blob/main/docs/guidelines/runner.md

docs/guidelines/plugin.md
```diff
@@ -206,9 +206,41 @@ is described in [API docs].
...
+### 3.1. Use the `dependency` message type for file dependencies
+
+If a plugin depends on another file, it should be declared by attaching
+a `dependency` message to the `result`:
+
+```js
+result.messages.push({
+  type: 'dependency',
+  plugin: 'postcss-import',
+  file: '/js/example.js',
```
ai, line 220: Let’s use image example as more popular one

```suggestion
  file: '/inline/image.png',
```

bradlc:
Thanks for the feedback @ai. I have made some changes 👍

bradlc:
@mischnic I have opened a [pull request](https://github.com/parcel-bundler/parcel/pull/6299) for this over on the Parcel repo.

PR #1582, opened 2021-05-16
Add Document node and related API (#1498)

### LINKED ISSUE #1498: Add new `Document` node type
It's a lot of text, I wanted to include enough information. It is important for a front-end tooling ecosystem. But maybe it's just a stylelint only problem with custom syntaxes ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯

## Custom syntaxes
PostCSS by default supports CSS in a form, where parsed file has only CSS. Most PostCSS syntaxes also for more classic CSS usage: SCSS, Sass, SugarSS, Less. However there are cases, when CSS is a part of other language. `style` tags and `style` attributes in HTML, Vue, Svelte. CSS-in-JS in JS template literals (Styled Components) or objects (JSS).

Currently `stylelint` supports both types of files: classic one file — one syntax, and the other one file — multiple syntaxes. It is done by huge work of @gucong3000. He made custom syntaxes to support multiple syntaxes in one file:

* [postcss-css-in-js](https://github.com/stylelint/postcss-css-in-js) — PostCSS syntax for parsing CSS in JS
* [postcss-html](https://github.com/stylelint/postcss-html) — PostCSS syntax for parsing HTML (and HTML-like)
* [postcss-markdown](https://github.com/stylelint/postcss-markdown) — PostCSS Syntax for parsing Markdown

All this syntaxes are powered by [postcss-syntax](https://github.com/stylelint/postcss-syntax). This package does two jobs: automatically switch PostCSS syntax based on file extensions, and provides common tools for other syntaxes (more on this later).

These packages were developed only by him. Unfortunately he is not active on Github for the past two years. These packages are complex and intertwined together. It's hard to understand how they work. No one else know how they work. I digged into them recently in an attempt to upgrade to PostCSS 8. Here's my surface understanding (they my not be accurate).

## `postcss-syntax`
`postcss-syntax`  does two jobs. First job is documented in package's readme: automatically switch PostCSS syntax based on file extensions. Second job is not documented, but very important for all other syntaxes: provide common tools for syntaxes. It introduces a new AST node type [`Document`](https://github.com/stylelint/postcss-syntax/blob/0bca5a408f12891a622fd6149d8f23d3bd98df86/document.js), but new AST node still has `type: 'root'`, because it's extended from PostCSS `Root` class.  It does it by monkey-patching PostCSS via Node.js [`require.cache`](https://github.com/stylelint/postcss-syntax/blob/0bca5a408f12891a622fd6149d8f23d3bd98df86/patch-postcss.js#L69) (crazy hack). For example for HTML document with two `style` tags we would have this kind of AST:

```
Document {
	nodes: [
		Root {
			nodes: [
				Rule {},
				AtRule {},
			]
		},
		Root {
			nodes: [
				Rule {},
				AtRule {},
			]
		},
	]
}
```

`postcss-syntax` makes sure that every PostCSS plugins are [run on each `Root`](https://github.com/stylelint/postcss-syntax/blob/0bca5a408f12891a622fd6149d8f23d3bd98df86/patch-postcss.js#L9-L23).

Also it adds extra information to `sources`.  Probably, for a successful stringifying.

Another issues with these syntaxes:

* Implicit dependencies and code execution. `postcss-syntax` also [calls](https://github.com/stylelint/postcss-syntax/blob/0bca5a408f12891a622fd6149d8f23d3bd98df86/load-syntax.js#L7-L14) custom syntax internal files, when `postcss/syntax/load-syntax.js` is [called](https://github.com/stylelint/postcss-html/blob/d17e25c1a50be9a0657fee453e49c7004b1d281a/extract.js#L4) from custom syntax.
* They use `postcss` internal files, which I believe is not allowed, because they are not part of official API.

### Why it works this way
I don't know exactly why `postcss-syntax` and dependent syntaxes architected this way.

However, I have an idea about some of behavior of these custom syntaxes. Each stylelint rule is a PostCSS plugin. These syntaxes were created with stylelint in mind. For many stylelint rules it makes more sense to have e. g. every styled component (<code>const Element = styled.div``</code>) as a separate root so a rule (PostCSS plugin) is run only on this piece of code.

For example we have following HTML:

```html
<p style="color: red"></p>

<p style="color: blue"></p>
```

If we don't distinguish them as separate Roots, rule like [`declaration-block-no-duplicate-properties`](https://stylelint.io/user-guide/rules/declaration-block-no-duplicate-properties) would show a lint violation.

We can't use `Rule` type for each `style` attribute, because we have many lint rules, which would check different aspects of `Rule` nodes (selectors, brackets, spacings), which would not be applicable for this fake `Rule`. In the mean time there is no stylelint rules, which has any specific behavior for a `Root` node.

Having `postcss-syntax` parse each entity as a separate Root, and they run plugins over each Root, allowed adoption of all this syntaxes in stylelint without changing hundreds of rules in stylelint codebase and in stylelint plugins.

## What it's all about
Custom syntaxes supported by stylelint and its plugins were in a rough state for the past two years. And usage of CSS in files, which are not CSS only, is increasing. It's important for stylelint and PostCSS to support such cases. I would like to find a solution to this problem. `postcss-css-in-js`, `postcss-html`, `postcss-markdown` and `postcss-syntax` in they current form should be changed. To support PostCSS 8 and unblock community to make changes to this syntaxes, by refactoring and making code architecture more clear.

I think it's important to find a vector which PostCSS and custom syntaxes should follow to support all types of CSS flavors modern development has.

hudochenkov:
New `Document` node and API.

* Added `Document` node.
	* `Document` node is optional. Default PostCSS parser doesn't create it. Because of this in test I had to create node manually every time.
	* Added `postcss.document()` to create node.
	* Allowed only `Root` nodes as children.
	* `Document` node doesn't have `raws`. Raws are handled in `Root` nodes. Stringifier concatenates whatever `Root` stringifier will give. No spacing or anything else is handled.
* Allow `raws.before` on `Root` node.
* Added `Document` and `DocumentExit` events to visitors API. They run similarly to `Root` and `RootExit`.
* `Once` and `OnceExit` run once per `Root` node. So if `Document` has two `Root` node children, then `Once` and `OnceExit` will run once per root. So in total twice per document.
*  Added `Node#document()` method. It works similar to `Node#root()` — return `Document` node in a tree, or node itself.
* Changed `Node#root()` to not go higher than the top most parent, which is not `Document` node.

Closes #1498.

Imaging that we have HTML syntax. Here's input:

```html
<html><style>a{color:black}</style><style>b{z-index:2}</style>
```

AST will look like:

```js
new Document({
     nodes: [
          new Root({
               raws: {
                    before: "<html><style>",
               },
               nodes: [
                    /* ... */
               ],
          }),
          new Root({
               raws: {
                    before: "</style><style>",
                    after: "</style>",
               },
               nodes: [
                    /* ... */
               ],
          }),
     ],
});
```


lib/document.d.ts
```diff
@@ -0,0 +1,264 @@
...
+   *
+   * ```js
+   * const root = postcss.parse('a { color: black }')
+   * root.nodes.length           //=> 1
+   * root.nodes[0].selector      //=> 'a'
+   * root.nodes[0].nodes[0].prop //=> 'color'
+   * ```
+   */
+  // @ts-expect-error
+  nodes: Root[]
```
hudochenkov, line 56: This and props/methods after I've copied from `container.d.ts` to limit children to `Root` only. However there are TypeScript errors, which I can't figure out how to solve.

```
Property 'nodes' in type 'Document' is not assignable to the same property in base type 'Container'.
  Type 'Root[]' is not assignable to type 'ChildNode[]'.
    Type 'Root' is not assignable to type 'ChildNode'.
      Type 'Root' is missing the following properties from type 'Rule': selector, selectors. ts(2416)
```

Descriptions I also didn't touch just yet.

lib/root.d.ts
```diff
@@ -29,7 +34,7 @@ export interface RootProps extends ContainerProps {
  */
 export default class Root extends Container {
   type: 'root'
-  parent: undefined
+  // parent: Document | undefined
```
hudochenkov, line 37: `Root` could have optional `Document` parent. However when I have this line uncommented (and `import Document from './document.js'`) there are type errors in `processor.test.ts`:

```
Type '{ parse: () => Root; stringify: (node: Node, builder: (s: string) => void) => void; }' is not assignable to type 'Parser | Syntax | undefined'.
  Type '{ parse: () => Root; stringify: (node: Node, builder: (s: string) => void) => void; }' is not assignable to type 'Syntax'.
    Types of property 'stringify' are incompatible.
      Type '(node: Node, builder: (s: string) => void) => void' is not assignable to type 'Stringifier'.
        Types of parameters 'node' and 'node' are incompatible.
          Type 'AnyNode' is not assignable to type 'Node'.
            Type 'Root' is not assignable to type 'Node'.
              Types of property 'parent' are incompatible.
                Type 'Document | undefined' is not assignable to type 'Container | undefined'.
                  Type 'Document' is not assignable to type 'Container'.
                    Types of property 'nodes' are incompatible.
                      Type 'Root[]' is not assignable to type 'ChildNode[]'. ts(2322)
```

test/visitor.test.ts
```diff
@@ -276,11 +283,39 @@ it('works visitor plugin add-prop', async () => {
...
+  })
+  expect(css).toEqual('.a{ color: red; }' + '.b{ will-change: transform; }')
+})
+
 it('works with at-rule params', () => {
-  let { css } = postcss([
-    replacePrintToMobile,
-    replaceScreenToPrint
-  ]).process('@media (screen) {}', { from: 'a.css' })
+  let { css } = postcss([replacePrintToMobile, replaceScreenToPrint]).process(
```
hudochenkov, line 315: Prettier changed.

test/visitor.test.ts
```diff
@@ -255,9 +261,10 @@ it('works classic plugin replace-color', async () => {
 })
 
 it('works visitor plugin will-change', async () => {
-  let { css } = postcss([
-    willChangeVisitor
-  ]).process('.foo { will-change: transform; }', { from: 'a.css' })
+  let { css } = postcss([willChangeVisitor]).process(
+    '.foo { will-change: transform; }',
+    { from: 'a.css' }
+  )
```
hudochenkov, line 267: Prettier changed.

test/visitor.test.ts
```diff
@@ -918,9 +1205,7 @@ for (let type of ['sync', 'async']) {
       }
     }
 
-    let result = postcss([
-      scanner
-    ]).process(
+    let result = postcss([scanner]).process(
```
hudochenkov, line 1208: Prettier changed.

lib/document.d.ts
ai: `Container` doesn’t allow `Root` nodes.

Possible solution: use `nodes: Node[]` in `Container` and `nodes: ChildNode[]` in `AtrRule`, `Rule`, `Root` (and `nodes: Root[]` in `Document`).

lib/root.d.ts
ai: Seems like it is related to https://github.com/postcss/postcss/pull/1582#discussion_r633122472 

Right now `Document#nodes` is incompatible with `Container#nodes`. But if we will refactor `Container#nodes` it could fix this issue too.

ai:
It is really great that you thought about many edge cases. Well done PR.

lib/document.d.ts
hudochenkov: It worked!

lib/root.d.ts
hudochenkov: Approach similar to fixing `nodes` worked in  this case too.

lib/document.d.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  /**
+   * The container’s first child.
+   *
+   * ```js
+   * rule.first === rules.nodes[0]
+   * ```
+   */
+  // @ts-expect-error
+  get first(): ChildNode | undefined
```
hudochenkov, line 54: There is 10 methods, which I had to silence type errors.

```
Property 'first' in type 'Document' is not assignable to the same property in base type 'Container'.
  Type 'Root | undefined' is not assignable to type 'ChildNode | undefined'.
    Type 'Root' is not assignable to type 'ChildNode | undefined'.
      Type 'Root' is missing the following properties from type 'Rule': selector, selectors. ts(2416)
```

In type errors `Root | undefined` is resolved `get first()` type, and `ChildNode | undefined` is a type, which comes from `container.d.ts`.

From what I google it's impossible to change method signature for a subclass. Currently I see only two ways:

1. Ignore error and keep `@ts-expect-error`.
2. Do the same trick as with `nodes`: in `container.d.ts` all these methods make wider by adding `Root`, and then in `atrule`, `rule`, `root` types make methods narrower. But it's a lot of copy-paste, which is error prone in future.

ai: There is a better way. You can use generic for `Container`:

```ts
class Container<Child extends Node = ChildNode> extends Node {
  nodes: Child[]
  get first (): Child | undefined
}
```

and use `extends Container<Root>` in `Document`.

hudochenkov:
When this PR is ready to merge, maybe `Document` node should be released as “experimental” or “unstable”? I have plans to create new syntax to parse styled components-like CSS-in-JS. More narrower version of [@stylelint/postcss-css-in-js](https://www.npmjs.com/package/@stylelint/postcss-css-in-js), which will be parsing styled components flavor of CSS-in-JS and works with PostCSS 8. During this field test some problems with `Document` could arise hypothetically. When everything is good, `Document` node could be “stable”.

lib/document.d.ts
hudochenkov: Interesting. I'll take a look tomorrow.

ai:
> When this PR is ready to merge, maybe Document node should be released as “experimental” or “unstable”? 

Hm. Looks like a good idea, especially since we do not use it in core parser.

Do you know the proper way to do it? Just put a comment in class description?

lib/document.d.ts
hudochenkov: Worked like a charm!

I'm curious why `extends Node` in `Child extends Node` is needed? Without this extend everything seems working.

hudochenkov:
I don't know a proper way. In stylelint we communicated such thing in documentation and in a changelog.

I've added a note to `document.d.ts`: https://github.com/postcss/postcss/pull/1582/files#diff-e78e125886d3e557b014e52ab18d4041858a7dd675b44989977891cac1253decR16

lib/document.d.ts
ai: Without `extends Node` `arg: Container` will have `any[]` in `arg.nodes`

lib/document.d.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export interface DocumentProps extends ContainerProps {
+  nodes?: Root[]
+}
+
+type ChildNode = Root
+type ChildProps = RootProps
+
+/**
+ * Represents a file and contains all its parsed nodes.
```
ai, line 14: Let’s add CSS-in-JS or HTML example to description.

lib/document.d.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * **Experimental!** Some aspects of this node could change within minor or patch version releases.
```
ai, line 16:
```suggestion
 * **Experimental:** some aspects of this node could change within minor or patch version releases.
```

lib/document.d.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * ```js
+ * const document = postcss.parse('<html><style>a{color:black}</style><style>b{z-index:2}</style>')
```
ai, line 19:
```suggestion
 * const document = htmlParser('<html><style>a{color:black}</style><style>b{z-index:2}</style>')
```

`postcss.parse` doesn’t have HTML support.

lib/document.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+let Container = require('./container')
+
+let LazyResult, Processor
+
+class Document extends Container {
```
ai, line 7:
```suggestion
class Document extends Container<Root> {
```

lib/postcss.d.ts
```diff
@@ -74,6 +76,20 @@ type CommentProcessor = (
 ) => Promise<void> | void
 
 interface Processors {
+  /**
+   * Will be called on `Document` node.
+   *
+   * Will be called again on children changes.
+   */
+  Document?: RootProcessor
```
ai, line 84: Do we need to use `DocumentProcessor` here?

lib/document.js
hudochenkov: It's a JavaScript file. I've added this to `document.d.ts`.

lib/postcss.d.ts
hudochenkov: Good catch!

hudochenkov:
I've addressed all comments.

lib/at-rule.d.ts
```diff
@@ -69,7 +70,9 @@ export interface AtRuleProps extends ContainerProps {
  */
 export default class AtRule extends Container {
   type: 'atrule'
+  parent: Container | undefined
   raws: AtRuleRaws
+  nodes: ChildNode[]
```
ai, line 75: Do we need this lines if we use generic in `Container`?

lib/declaration.d.ts
```diff
@@ -51,6 +52,7 @@ export interface DeclarationProps {
  */
 export default class Declaration extends Node {
   type: 'decl'
+  parent: Container | undefined
```
ai, line 55: Should we move it to `Node` generic too? And use `Node<Parent extends Node = Container>`, `Container<Child …, Parent …> extends Node<Parent>`

lib/document.d.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * **Experimental:** some aspects of this node could change within minor or patch version releases.
+ *
+ * ```js
+ * const document = htmlParser('<html><style>a{color:black}</style><style>b{z-index:2}</style>')
+ * document.type         //=> 'document'
+ * document.nodes.length //=> 2
+ * ```
+ */
+export default class Document extends Container<Root> {
```
ai, line 24: If `Document` has no `raws`, how we will track symbols after the last style?

Should we add `Document.raws.after`? It will fit `AtRule` and `Rule` way to keep last symbols.

lib/node.js
```diff
@@ -151,7 +151,18 @@ class Node {
...
   root() {
     let result = this
-    while (result.parent) result = result.parent
+    while (result.parent && result.parent.type !== 'document') {
+      result = result.parent
+    }
+    return result
+  }
+
+  document() {
```
ai, line 160: It could return `Root` as well. `document` could be misleading.

Suggestions:

1. Add `getTop()` or some common name for both `Root` and `Document`.
2. Do not extend API. It could be better to add helper only if we really need it somewhere.

lib/document.d.ts
hudochenkov: I have in mind that `Root.raws.before` and `Root.raws.after` are used:

https://github.com/postcss/postcss/blob/a5ac723f1f63c59db0a6e4aa7747c400419d755d/test/stringifier.test.js#L266-L286

ai: Hm. I am worried that it is a different way to take deal with spaces compared to `Rule`/`AtRule`.

But on another hand, we already have `Root.raws.after`.

Do you think that the `Root.raws.after` is better?

lib/node.js
hudochenkov: Indeed it could be confusing.

`root()` method also could return any top node, even not `Root`. Here test currently in `main` branch:

https://github.com/postcss/postcss/blob/a5ac723f1f63c59db0a6e4aa7747c400419d755d/test/node.test.ts#L320-L323

I've added `.document()` to match current `root()` behavior. I think the best way is to remove `.document()`, and only add it if there is a need.

ai: Yeap, let’s remove it. I like to more too.

lib/declaration.d.ts
hudochenkov: I have no luck with this due lack of experience working with generics and classes. When added `<Parent extends Node = Container>` to `Node` class, then visitor test shows type errors. For `Container` I couldn't get past error in `Container` type.

lib/node.js
hudochenkov: Removed.

lib/at-rule.d.ts
hudochenkov: Fixed.

lib/declaration.d.ts
ai: Don’t worry. I will try to fix it after merging PR.

lib/document.d.ts
hudochenkov: ## Current state

`Rule`/`AtRule` treat `before` as string before the rule. `after` is a string after the last child node.

`Root` doesn't used `before` in default parser. `after` is a string after the last child node.

Name `after` is confusing, because it doesn't do opposite of `before`. It should have been `afterLastChild` :)

## First idea

I think if we want to keep the same approach to `before` (string before the node) and `after` (string after the last child of the node), we would need to add `Root.raws.before`, and `Document.raws.after`.

`Root.raws.before` is already added in this PR, `Document.raws.after` is missing.

For the following HTML:

```html
<html>
<head>
<style>
a{color:black}

</style>
<style>

b{z-index:2}
</style>
</head>
```

We would have following AST:

```js
new Document({
    raws: {
        after: "</style>\n</head>",
    },
    nodes: [
        new Root({
            raws: {
                before: "<html>\n<head>\n<style>",
                after: "\n\n",
            },
            nodes: [
                /* ... */
            ],
        }),
        new Root({
            raws: {
                before: "</style>\n<style>",
                after: "\n",
            },
            nodes: [
                /* ... */
            ],
        }),
    ],
});

```

In this examples Root boundaries would be content between `style` tags. As if we took this content and put it in a separate CSS file and parsed. 

## Second idea (a better one, in my opinion)

After writing above, looking at the AST, and some thinking... I think this is all wrong :)

https://github.com/postcss/postcss/blob/ab60e0d2963aa0e3fa879a83ecdb5bd607375c6c/lib/node.d.ts#L144-L150

Adding `Root.raws.before` and using it not for spacing could potentially be a breaking change. E. g. some plugin looks into every node's `raws.before` and remove spaces. Then if `before` for parsed HTML would contain some HTML markup stringified output would be broken.

This is what I think is what we need to have. Due to differences what parsed document could be (HTML, JS, TS, Markdown, vue, etc) we should not change any `raws`. So revert `Root.raws.before` in this PR.

`Document.raws` should be in types, but no specifics (`any`).

Every syntax will put whatever they want in `Document.raws`, and in `Root` could use any new key.

Idea is to keep Root the same as if PostCSS default parser would create it, but we could add new `raws` keys with names, that PostCSS doesn't use.

Here how I would make HTML parser:


```js
new Document({
    nodes: [
        new Root({
            raws: {
                markupBefore: "<html>\n<head>\n<style>",
                after: "\n\n",
            },
            nodes: [
                /* ... */
            ],
        }),
        new Root({
            raws: {
                markupBefore: "</style>\n<style>",
                after: "\n",
                markupAfter: "</style>\n</head>",
            },
            nodes: [
                /* ... */
            ],
        }),
    ],
});
```

If we remove `markupBefore` and `markupAfter` it would be regular CSS and any plugin will work as expected.

ai: > After writing above, looking at the AST, and some thinking... I think this is all wrong :)

Is it about first or second idea?

> Second idea (a better one, in my opinion)

I like it more, but with `codeBefore` and `codeAfter` instead of `markup`, since we will not have markup in CSS-in-JS.

hudochenkov: I think first idea is a bad one.

In second idea raws names are up to developer of a syntax, because PostCSS wouldn't do anything with it. But I think your names for raws are better :)

ai: Good. Change the `raws` and we are ready to merge.

hudochenkov: I removed `Root.raws.before`

Added `raws` for `Document` and decided to test custom syntax with `Document`. This revealed that some types need to be adjusted. Also it helpful to have at least one test with syntax which works with `Document` :)

ai:
I renamed `markupBefore` to `codeBefore` and added them to `Root.raws` types 7b3c872

ai:
I will try to release 8.3 tomorrow

hudochenkov:
> I renamed markupBefore to codeBefore

Damn it, I wanted to use new names but was caught up in understanding how stringifier works :)

Thank you!

ai:
Released in [8.3](https://github.com/postcss/postcss/releases/tag/8.3.0).

hudochenkov:
@ai Thank you!

PR #1590, opened 2021-06-01
Add `glob` message property documentation

bradlc:
As discussed in #1585 this pull requests adds documentation for an optional `glob` property for `dir-dependency` messages. This property can be used to indicate which files within a directory a plugin depends on:

```js
result.messages.push({
  type: 'dir-dependency',
  plugin: 'postcss-import',
  dir: '/imported',
  glob: '**/*.css',
  parent: result.opts.from
})
```

### DIFF
--- a/docs/guidelines/plugin.md
+++ b/docs/guidelines/plugin.md
@@ -222,13 +222,17 @@ result.messages.push({
 })
 ```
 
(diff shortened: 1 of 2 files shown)


bradlc:
I totally forgot to add @DeMoorJasper as a co-author on the commit but this was 100% their idea! 🙏

PR #1605, opened 2021-06-18
Fix duplicate changelog

megos:
7.0.33 to 7.0.36 were duplicated.


ai:
Thanks. Seems like it is the result of merge.

Issue #1617, opened 2021-07-21
checkMissedSemicolon gives incorrect colomn/token


Gusted:
Hi!

Coming off from [a issue with stylelint](https://github.com/stylelint/stylelint/issues/5419), seems like that the postCSS parser is returning an incorrect token/column from the `checkMissedSemicolon` function.

A simple test-case:
```CSS
test.css
div {
        color: red !important
        background-color: #000
}
```
```bash
gusted@gusted-beast [15:38:26] [~] 
-> % node -pe 'require("postcss").parse(fs.readFileSync("test.css", "utf-8"))'

/usr/lib/node_modules/postcss/lib/parse.js:33
    throw e
    ^

CssSyntaxError: <css input>:2:20: Missed semicolon
    at Input.error (/usr/lib/node_modules/postcss/lib/input.js:123:16)
    at Parser.checkMissedSemicolon (/usr/lib/node_modules/postcss/lib/parser.js:555:22)
    at Parser.decl (/usr/lib/node_modules/postcss/lib/parser.js:253:12)
    at Parser.other (/usr/lib/node_modules/postcss/lib/parser.js:147:12)
    at Parser.parse (/usr/lib/node_modules/postcss/lib/parser.js:59:16)
    at Function.parse (/usr/lib/node_modules/postcss/lib/parse.js:11:12)
    at [eval]:1:20
    at Script.runInThisContext (node:vm:129:12)
    at Object.runInThisContext (node:vm:305:38)
    at node:internal/process/execution:81:19 {
  reason: 'Missed semicolon',
  source: 'div {\n        color: red !important\n        background-color: #000\n}\n',
  line: 2,
  column: 20,
  input: {
    line: 2,
    column: 20,
    source: 'div {\n        color: red !important\n        background-color: #000\n}\n'
  }
}
```

As you see it says column: 20 which is before the `!important`, however the error should shay it's an error at column: 30, to include the `!important`

ai:
Can you send PR? Sorry, have no resources for the project on this week.

Gusted:
> Can you send PR? Sorry, have no resources for the project on this week.

No problems, #1618, got a interesting look in how PostCSS works 😄.

PR #1618, opened 2021-07-21
Return correct column on `missed semicolon`

Gusted:
I've added some test-cases to ensure it works.
Also added a comment about `why` we it need to get token[3]+1

And ultimately it resolves https://github.com/postcss/postcss/issues/1617

Regards,
Gusted

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -552,7 +552,13 @@ class Parser {
         if (founded === 2) break
       }
     }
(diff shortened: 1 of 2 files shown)


ai:
Thanks. I will release it today.

Gusted:
Thank you for merging and your time.

ai:
Released in 8.3.6.

PR #1654, opened 2021-11-02
refactor: refactor buffer from array to string

CGQAQ:
When I tried to do some evil tricks in [this project](https://github.com/justjavac/postcss-rs/pull/28), I found that this change would make all tests pass.
If this change is proving to be valid, it will be a huge performance boost.
Please take a look at this PR, thank you.

### DIFF
--- a/lib/tokenize.js
+++ b/lib/tokenize.js
@@ -288,6 +288,7 @@ module.exports = function tokenizer(input, options = {}) {
         case '}':
           return i
         case '/':
+          /* istanbul ignore next */
           if (_input[i - 1] === '*') {
             return i
           } else {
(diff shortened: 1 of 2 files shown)


ai:
Looks very interesting.

Can you show https://github.com/postcss/benchmark/ result before and after this change?

CGQAQ:
Still don't know if this change is valid or not
image

I sad a huge performance boost is saying about the rust implementation version

ai:
Thanks for the trying to backport optimization.

Even if it didn’t help in JS environment, it is important work (who knows, maybe another idea will be revolution).

And as usual, feel free to ask promotion for your PostCSS on Rust project.

PR #1662, opened 2021-11-08
Add postcss-opacity-percentage to list of plugins

Dreamseer:
`postcss-opacity-percentage` is a plugin for transforming a CSS4 percentage-based `opacity` value to the more supported float-based value (e.g. `opacity: 75%;` ➡️ `opacity: 0.75;`).

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -130,6 +130,8 @@ by one line of code.
 * [`postcss-hash-classname`] append hash string to your css class name.
 * [`postcss-mqwidth-to-class`] converts min/max-width media queries to classes.
 * [`postcss-opacity`] adds opacity filter for IE8.
+* [`postcss-opacity-percentage`] transforms CSS4 percentage-based `opacity`
+  values to float values.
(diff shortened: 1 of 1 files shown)


ai:
Thanks. Looks super-cool.

1. Do you want to add it to `postcss-preset-env`?
2. Maybe you should send PR to Stylelint. They have a rule for percentage-based `opacity` enabled by default and will be happy to promote it on rule’s page.
3. Feel free to give me a tweet with plugin announce. I will rewteet in in `@postcss`.

Dreamseer:
1. Yeah, eventually I’ll create a PR for `postcss-preset-env`. But I guess it needs to be added to `cssdb` first.
2. That default rule is the reason I created this plugin in the first place. 😄
3. I’ll mention `@postcss` in the tweet so you get notified.

Semigradsky:
Was added to "Candidates for adding" https://github.com/csstools/postcss-preset-env/issues/191#issuecomment-693302956 🎉

Issue #1661, opened 2021-11-08
Move from Jest to uvu


ai:
[`uvu`](https://github.com/lukeed/uvu/) is faster, simpler (no global API), and have less foorprint to `node_modules` (which speed up CI too).

PR #1665, opened 2021-11-15
Move from Jest to uvu

kimoofey:
Move from Jest to uvu #1661


package.json
```diff
@@ -53,7 +53,8 @@
     "transpiler"
   ],
   "scripts": {
-    "test": "jest --coverage && eslint . && ./test/version.js && check-dts && ./test/integration.js && size-limit"
+    "test": "c8 yarn unit && eslint . && ./test/version.js && check-dts && ./test/integration.js && size-limit",
```
ai, line 56: We need to copy `c8` config with coverage limit

https://github.com/ai/nanodelay/blob/main/package.json#L51-L57

test/browser.test.ts
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
```
ai, line 1: We can remove this test. It tests rare case.

ai:
We also need to change CI task to `yarn unit` https://github.com/postcss/postcss/blob/main/.github/workflows/test.yml#L63 and https://github.com/postcss/postcss/blob/main/.github/workflows/test.yml#L44

ai:
We need to fix:

1. Coverage
2. Windows support (should we add `uvu` issue?)
3. Node.js 10 support, when there is not ESM. I am OK to create some `simple-test.js` script which will parse simple CSS if ESM support is not available (but maybe we can hack Node.js with custom loader?).

kimoofey:
1. Do we need to increase coverage with tests or reduce this number in c8 rule in `package.json` (I've seen uncovered code and all of them with precommented line `// istanbul ignore next`)?
2. As a alternative, to rewrite `yarn unit` command to `uvu -r tsm test` (it will works on both platforms, but need to replace `version.js` and `integration.js`, so they don't run twice on `yarn test`)

ai:
> I've seen uncovered code and all of them with precommented line `// istanbul ignore next`

Seems like we forget to convert `istanbul` ignore comments to `c8` ignore comments

> As a alternative, to rewrite yarn unit command to `uvu -r tsm test`

How will it help?

We already has `uvu -r tsm test` in `yarn unit` script. As I understand, the problem is with shell script.

kimoofey:
> How will it help?

As in uvu [docs](https://github.com/lukeed/uvu/blob/master/docs/cli.md#cli), we can specify only target folder (without matching pattern, that different on some platforms).

ai:
> As in uvu docs, we can specify only target folder (without matching pattern, that different on some platforms).

I afraid it will run all JS files. It is not our case. We need to find a way to run only specific files.

Solution:

1. Use a different command for Windows
2. Ask `uvu` team for some cross-platform API

.github/workflows/test.yml
```diff
@@ -60,6 +60,6 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
       - name: Run unit tests
-        run: npx jest
+        run: yarn unit:win
```
ai, line 63: If we need a script only in the single place, let’s put command here directly.

```suggestion
        run: npx uvu -r tsm test ^.*test.([tj]s)
```

ai:
@kimoofey sorry, you will need to update PR because I merged another PR with test changes

ai:
@kimoofey all other 8.3 PRs was merged. I hope we will not have more conflicts.

kimoofey:
Everything is fine, except Node 10 support
I was looking for additional loader (for example this [one ](https://github.com/lukeed/loadr)), but still have not success with it

ai:
> Everything is fine

We also have Windows issue.

> except Node 10 support

I asked `uvu` author https://github.com/lukeed/uvu/issues/162

If there is no way we can write a simple smoke test. A simple Node.js script (without any test framework) which will parse CSS and accept some basic plugin.

ai:
@kimoofey `uvu`’s author recommends `uvu -r esm tests` to run on Node.js 10

ai:
Great. Now only coverage is a blocker.

package.json
```diff
@@ -54,7 +54,8 @@
     "transpiler"
   ],
   "scripts": {
-    "test": "jest --coverage && eslint . && node ./test/version.js && check-dts && node ./test/integration.js && size-limit"
+    "test": "c8 yarn unit && eslint . && node ./test/version.js && check-dts && node ./test/integration.js && size-limit",
+    "unit": "ts-node -T node_modules/uvu/bin.js -r esm test '\\.test\\.(ts|js)$'"
```
43081j, line 58: you can probably:

```sh
uvu -r ts-node/regiser -r esm test '\\.test\\.(ts|js)$'
```

kimoofey: you right, `ts-node` will be enough here. will remove `esm`

PR #1668, opened 2021-11-16
fix: add `Document` to result sets and default parser type

### LINKED ISSUE #1667: `parser` option is typed differently to `syntax['parse']`, `root` is always a `Root` but can be a `Document`
Here we define the `parser` option as being a `Syntax` or a `Parser<Root>` (since `Root` is the default type parameter):

https://github.com/postcss/postcss/blob/6318030a62ceb1667cdd3fb9e49e8e12a910c003/lib/postcss.d.ts#L329

Here we define the `parse` of a `Syntax` as being a `Parser<Root | Document>`:

https://github.com/postcss/postcss/blob/6318030a62ceb1667cdd3fb9e49e8e12a910c003/lib/postcss.d.ts#L249

Maybe these should be aligned?

Similarly, we define the `root` of a `Result` as being a `Root`:

https://github.com/postcss/postcss/blob/6318030a62ceb1667cdd3fb9e49e8e12a910c003/lib/result.d.ts#L97

This isn't true anymore, though. It is possible a custom syntax can emit a `Document` as the root.

43081j:
Fixes #1667

A `Result` can now contain a root which is ` Document`. For example, a
custom syntax may produce a `Document` containing several nested `Root`
stylesheets.

Similarly, the default type parameter of `Parser` has been updated to
include `Document` since custom syntaxes must account for this anyway.

FYI the casts in the tests shouldn't reduce our test accuracy, since if the type was anything which doesn't cross over with `Root`, typescript would complain.

### DIFF
--- a/lib/postcss.d.ts
+++ b/lib/postcss.d.ts
@@ -222,7 +222,7 @@ export type AcceptedPlugin =
     }
   | Processor
 
-export interface Parser<RootNode = Root> {
+export interface Parser<RootNode = Root | Document> {
   (
     css: string | { toString(): string },
     opts?: Pick<ProcessOptions, 'map' | 'from'>
@@ -246,7 +246,7 @@ export interface Syntax {
   /**
    * Function to generate AST by string.
    */
-  parse?: Parser<Root | Document>
+  parse?: Parser
 
   /**
    * Class to generate string by AST.
@@ -379,7 +379,7 @@ export interface Postcss {
    * root1.append(root2).toResult().css
    * ```
    */
-  parse: Parser
+  parse: Parser<Root>
 
   /**
    * Rehydrate a JSON AST (from `Node#toJSON`) back into the AST classes.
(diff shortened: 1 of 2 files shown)


ai:
Can we return `Root` type from `parse()` if `parser` option was not changes? It will allow avoiding `as Root`.

43081j:
maybe this works?

so the static entrypoints will always use a `Parser<Root>`.

e.g.:

```ts
import {
  default as postcss,
  parse
} from 'postcss';

parse(...); // Root
postcss.parse(...); // Root
```

ai:
> so the static entrypoints will always use a Parser<Root>

Sure

43081j:
i pushed that btw if you don't mind re-reviewing

ai:
I will release it in 8.3 when https://github.com/postcss/postcss/pull/1665 will be finished

PR #1669, opened 2021-11-17
feat: Support ranges for errors and warnings

adalinesimonian:
Context: https://github.com/stylelint/vscode-stylelint/issues/315, https://github.com/stylelint/stylelint/issues/5694

Allows for better diagnostics in linters, language servers, and plugins.

Currently, since only single positions are supported, this means that downstream, when we display errors in different tools, in my particular case, a language server and editor integration, we can only display a single position as part of the error, even when the error pertains to more than just one character.

These changes allow warnings and errors to be emitted with an additional `endLine` and `endColumn` which pertains to an end position, exclusive, of the problem. It also adjusts CSS syntax errors to always provide an end position for consistency. With these additions, we would be able to better highlight problems in Stylelint and other tools and use that information in integrations for a better developer experience.

- End position is exclusive for consistency with Language Server Protocol ranges
- Tests added for ranges
- Default warning message text is unchanged and still shows a single position
- Additional properties added to warnings, warning options, errors, and file positions
- `rangeBy` added, which provides a range for a given word or start and end indices
- Changes to functions made in such a way that they can still be called with their former signatures to avoid breaking existing dependent code

### DIFF
--- a/lib/css-syntax-error.d.ts
+++ b/lib/css-syntax-error.d.ts
@@ -1,6 +1,21 @@
 import { FilePosition } from './input.js'
 
 /**
+ * A position that is part of a range.
+ */
+export interface RangePosition {
+  /**
+   * The line number in the input.
+   */
+  line: number
+
+  /**
+   * The column number in the input.
+   */
+  column: number
+}
+
+/**
  * The CSS parser throws this error for broken CSS.
  *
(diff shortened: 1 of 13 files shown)


ai:
Wow. Looks cool and very on time (I am planing 8.3 release next week or next month).

yoyo837:
Great work!

PR #1674, opened 2021-11-19
fix: make test script work on windows

adalinesimonian:
Running `./script.js` doesn't work on Windows, only `node ./script.js` does. This patch updates the test script so that it works on Windows as well as on other platforms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "transpiler"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


ai:
Strange. It works on CI.

adalinesimonian:
The CI workflow for Windows doesn't call the test script: https://github.com/postcss/postcss/blob/main/.github/workflows/test.yml#L53-L65

PR #1675, opened 2021-11-19
fix: add stringifier type definitions

### LINKED ISSUE #1673: Expose `Stringifier` class to make extending the default implementation easier
I'm writing a syntax for parsing CSS in JS (though tightly scoped to [lit](https://lit.dev)).

Since i'll be mostly stringifying regular CSS, ideally I want to use the default stringifier but override what happens with some things:

```ts
class LitStringifier extends Stringifier {
  comment(node: Comment): void {
    // currently i use comments as placeholders for some original source we stripped
    // so here i would build the original text if its a "special" comment
  }

  root(node: Root): void {
    // handle the codeBefore
    super.root(node);
    // handle the codeAfter
  }
}
```

`Stringifier` currently has no typescript definitions and no `default` export (i.e. your usual `Foo.default = Foo` workaround).

could we expose this? im happy to add the types and the default export if there's no reason not to.

this way i can new up my custom stringifier just like your internal stringifier function does, and rely on the default implementation for the rest.

43081j:
Fixes #1673 

Some of the methods have `undefined` in the return type because of this pattern:

```ts
let value; // undefined at this point

root.walk((n) => {
  if (n.something) {
    value = n.something;
  }
});

return value;
```

which would mean `value` can sometimes be undefined (if `something` was never found).

let me know if you want me to update the JS to have a default value

### DIFF
--- a/lib/stringifier.d.ts
+++ b/lib/stringifier.d.ts
@@ -1,0 +1,57 @@
+import {
+  Document,
+  Root,
+  Comment,
+  Declaration,
+  Builder,
+  AnyNode,
+  Rule,
+  AtRule,
+  Container
+} from './postcss.js';
+
+export class Stringifier {
+  public constructor(builder: Builder);
+
+  public stringify(node: AnyNode, semicolon?: boolean): void;
+
+  public document(node: Document): void;
+
+  public root(node: Root): void;
+
+  public comment(node: Comment): void;
+
+  public decl(node: Declaration, semicolon?: boolean): void;
+
+  public rule(node: Rule): void;
+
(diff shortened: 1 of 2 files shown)


lib/stringifier.d.ts
```diff
@@ -0,0 +1,57 @@
...
+  Declaration,
+  Builder,
+  AnyNode,
+  Rule,
+  AtRule,
+  Container
+} from './postcss.js';
+
+export class Stringifier {
+  public constructor(builder: Builder);
```
ai, line 14: Can you remove `public` prefix for better readability?

PR #1679, opened 2021-11-23
Update README.md

hasnath1:
I have used the extension you recommended but it is horrible . When I use it  in .css file , emmet sugestions don't work at all . So I changed the link of the extension . I am using this https://marketplace.visualstudio.com/items?itemName=mhmadhamster.postcss-language extension for postcss and It is working great . It also supports .css files and enables emmet .


ai:
Let put both for a while

PR #1681, opened 2021-11-24
fix: export stringifier as a default

43081j:
@ai is there any chance we can get some kind of hotfix release out for 8.4?

i was an idiot and exported the stringifier named, rather than as a default... it is actually a default export 🤦‍♂️


ai:
Sure. Will do it today evening

ai:
Released in 8.4.1

PR #1686, opened 2021-12-01
[docs] add link to Svelte preprocessor

benmccann:
To use PostCSS with Svelte, you must install the preprocessor which first converts PostCSS syntax into regular CSS so that the Svelte compiler can handle it


ai:
Can you put it to `Runners` section?

benmccann:
done!

PR #1697, opened 2021-12-23
Update README.md

vladsholokhov:
Updated information about JetBrains IDEs

### DIFF
--- a/README.md
+++ b/README.md
@@ -508,10 +508,9 @@ module.exports = {
 
 ### WebStorm
 
-WebStorm 2016.3 [has] built-in PostCSS support.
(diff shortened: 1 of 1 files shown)


ai:
Can you explain this changes and compare IDE with and without this plug-in?

vladsholokhov:
Hello!
1) I replaced the link to the very old(2016) blog-post with the link to the plugin. Information of "How to use it" and "What's supported" can be found in the plugin's description.
2) `... [has] built-in PostCSS support.` I updated this phrase because **built-in support** for JB IDEs usually means that you don't need to install anything.

PR #1722, opened 2022-02-24
Fix `Node#warn()` return type

ybiquitous:
I think `Node#warn()` should return `Warning`, not `void`:

https://github.com/postcss/postcss/blob/8718e0656e78f96d7812300bc372ff3a574d6fad/lib/node.js#L70-L74


ai:
LGTM.

I will try to merge it and release when I will be back from anti-war protest.

ybiquitous:
Thank you! 😊

ai:
Thanks for the fix. Released in 8.4.7.

Tuneabq (commented):
👍💚

PR #1740, opened 2022-04-29
Fix `append()` failing after mutating nodes through .parent

### LINKED ISSUE #1738: Traversing though `.parent` and then mutating `.nodes` causes `append` to fail
If you use postcss-nested to un-nest CSS, later traverse the tree, and mutate the parent's children it can cause calls to `append` to fail for any _original_ reference to the parent. If you go through `parent` to append children the call works fine.Additionally, if you mutate the children with the original reference the call to append will succeed.

I looked into this some and it appears that `append` calls `normalize` on the nodes passed to the children. This results in calls to `parent.removeChild(…)` which tries to find the node, fails silently, and uses the `-1` index to remove the child. This the removal of the parent from the node to throw an error because `this.proxyOf.nodes[child]` is undefined.

I'm guessing this is because the proxies postcss is using aren't being cleaned up properly when the parent's children are reassigned (at least I think that's what's happening) but I'm not sure. I wanted to PR the fix for this but I'm not entirely sure where to start on this one. Any help would be super appreciated, thanks!

Here's a short reproduction using `postcss-nested`:

```js
let root = postcss.parse(`
  @media (min-width: 640px) {
    .page {
      width: auto;
    }
  }
  .page {
    @media (min-width: 640px) {
      width: auto;
    }
  }
`)

let result = await postcss([
  postcssNested,
  
  // This is meant to emulate collapsing the declarations into one
  {
    postcssPlugin: 'some-plugin',
    OnceExit(root) {
      let firstRule = root.nodes[0]
      let secondRule = root.nodes[1]

      let node = firstRule.nodes[0].parent
      node.nodes = node.nodes
      firstRule.append(secondRule.nodes)
      secondRule.remove()
    },
  },
]).process(root, { from: 'whatever' })
```

And here's a reproduction using just postcss with more explanation and other scenarios that appear to work fine. Note that one of the classes was changed to simplify the `Rule` visitor:

```js
let root = postcss.parse(`
  @media (min-width: 640px) {
    .page {
      width: auto;
    }
  }
  .nested {
    @media (min-width: 640px) {
      width: auto;
    }
  }
`)

let result = await postcss([
  {
    postcssPlugin: 'some-plugin',
    
    // Make some changes to the tree
    // This appears to be required
    // This emulates having used postcss-nested to mutate the tree
    Rule (rule) {
      // This is obviously not the same as the postcss-nested version
      // but this simpler visitor still reproduces the bug
      if (rule.selector === '.nested' && rule.nodes.length === 1) {
        rule.after(rule.nodes[0])
      }
    },

    // This is meant to emulate collapsing the declarations into one
    OnceExit(root) {
      let firstRule = root.nodes[0]
      let secondRule = root.nodes[1]
      let node

      // If we:
      // 1. Traverse the tree
      // 2. Traverse back up the parent
      // 3. Mutate that parent's children (even by setting it to itself)
      // 4. Later append nodes to the parent using an original reference (not from traversal)
      // 5. The call to append will throw an error

      // This fails:
      node = firstRule.nodes[0].parent
      node.nodes = node.nodes
      firstRule.append(secondRule.nodes)

      // This works #1:
      // node = firstRule.nodes[0].parent
      // node.nodes = node.nodes
      // firstRule.nodes[0].parent.append(secondRule.nodes)

      // This works #2:
      // node = firstRule.nodes[0].parent
      // node.nodes = node.nodes
      // firstRule.append(secondRule.nodes)

      // This works #3:
      // node = firstRule
      // node.nodes = node.nodes
      // firstRule.append(secondRule.nodes)
      
      secondRule.remove()
    },
  },
]).process(root, { from: 'whatever' })
```

thecrypticace:
Fixes #1738

I wasn't quite sure where to put the test but the gist of the issue is that

1. `node.append` calls `node.normalize`
2. `node.normalize` sets the parent to `this`.

The problem arises when `node` is a proxy. This means that `this` ends up being a proxy instead of the raw node. Assigning the proxy to the "raw" child node's parent has a side-effect of causing `proxyOf.nodes` to contain proxies rather than raw postcss nodes.

This causes `indexOf` lookups to fail since they're sadly not transparent to Proxies b/c of object identity.
Later on `append` tries to use this index assuming that the nodes passed in are definitely children and have been found and crashes.

I'm not 100% sure this is the right fix because the interactions just to get this to happen are fairly specific. Change any number of things and it all magically works. But, it seems reasonable.

I noticed that the `.push` method sets the parent as well but I don't have a test case where it's affected so I've not updated it. I might be able to come up with a test case if needed. Also, if you have any suggestions for reducing the test case I've got in this PR I would love some because it seems like it's a bit much for such a "simple" fix.

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -326,7 +326,7 @@ class Container extends Node {
           i.raws.before = sample.raws.before.replace(/\S/g, '')
         }
       }
-      i.parent = this
+      i.parent = this.proxyOf
       return i
     })
 
--- a/test/visitor.test.ts
+++ b/test/visitor.test.ts
@@ -7,6 +7,7 @@ import postcss, {
   Container,
   Root,
   Rule,
+  AtRule,
   Declaration,
   Plugin,
   PluginCreator,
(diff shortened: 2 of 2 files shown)


thecrypticace:
Whoops my bad! Fixed that up.

ai:
Released in 8.4.13

thecrypticace:
Awesome — thank you very much!

PR #1744, opened 2022-05-13
fix: print deprecation warning only when plugin is used

zardoy:
I use some CJS package and I import from its index.js the functionality I need, however it also exposes an old postcss plugin. I don't use it, so I think it makes more sense to print a warning only when the plugin is actually used. I also added a check to preserve behavior that the warning is being printed only once.

### DIFF
--- a/test/postcss.test.ts
+++ b/test/postcss.test.ts
@@ -122,9 +122,12 @@ test('has deprecated method to creat plugins', () => {
     }
   })
 
+  equal(warn.callCount, 0)
+
   let func1: any = postcss(plugin).plugins[0]
   is(func1.postcssPlugin, 'test')
   match(func1.postcssVersion, /\d+.\d+.\d+/)
+  equal(warn.callCount, 1)
 
   let func2: any = postcss(plugin()).plugins[0]
   equal(func2.postcssPlugin, func1.postcssPlugin)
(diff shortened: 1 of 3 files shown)


test/postcss.test.ts
```diff
@@ -134,9 +132,13 @@ test('has deprecated method to creat plugins', () => {
 
   let result1 = postcss(plugin('one')).process('a{ one: 1; two: 2 }')
   is(result1.css, 'a{ two: 2 }')
+  equal(warn.callCount, 1)
```
zardoy, line 135: Just added another check to ensure we don't duplicate warnings into console

zardoy:
Should be good now

test/postcss.test.ts
```diff
@@ -122,17 +122,19 @@ test('has deprecated method to creat plugins', () => {
     }
   })
 
+  equal(warn.callCount, 0)
+
   let func1: any = postcss(plugin).plugins[0]
   is(func1.postcssPlugin, 'test')
   match(func1.postcssVersion, /\d+.\d+.\d+/)
+  equal(warn.callCount, 1)
```
zardoy, line 130: Also moved it to ensure it is printed here

ai:
I fixed CI in `main`. I will look at PR on next week.

ai:
Released in 8.4.14.

PR #1757, opened 2022-07-03
update plugins

romainmenke:
With so many plugins being absorbed into the csstool mono repo the plugins list here had a few entries that linked to archived repos for plugins that are actively maintained.

I did not add each new plugin we created that is intended for `postcss-preset-env`.
These have a generated list which is easier to maintain : https://github.com/csstools/postcss-plugins/blob/main/plugin-packs/postcss-preset-env/FEATURES.md#list-of-features-in-postcss-preset-env


ai:
Yes, it is enough. Maybe later I will clean up this list. For now, there is better DB with plugins.

Issue #1765, opened 2022-08-09
Strict Parser should throw error or Safe Parser should reject the declaration, if CSS variables are not accompanied by proper unicode character


phoenisx:
## Issue

Parser successfully returns for CSS custom properties accompanied by improper Unicode character.

## Current Behaviour

```js
await postcss().process(":root { -- : #333; }");

// Returns
{
  ...
   nodes: [
    Declaration {
      raws: [Object],
      type: 'decl',
      parent: [Circular *1],
      source: [Object],
      prop: '--', // This should be ignored
      value: '#333',
      [Symbol(isClean)]: false,
      [Symbol(my)]: true
    }
  ],
  ...
}
```

## Expected Behaviour

The parser for the above content should throw an error (or ignore it if used with safe-parser).

### References:

<div align="center">
<a href="https://drafts.csswg.org/css-syntax/#ident-token-diagram" target="_blank">
<img width="66%" alt="image" src="https://user-images.githubusercontent.com/11786283/183624927-8be38d7e-eb28-4b46-ae82-7ba854467bb8.png">
</a>

<img width="32%" alt="image" src="https://user-images.githubusercontent.com/11786283/183625700-7ead79ad-4194-43d1-923b-5ce8e40866ea.png">

ai:
Can I ask you to send PR? I agree that we could improve it, but I personally does not see a use case and prefer focusing on another task. But I will help you.

phoenisx:
Sure @ai
Let me take a look. I would be able to work on it next weekend.

<br>

> P.S: Forgot to mention: _Thanks for building such a great tool_ 🙇🏽

Issue #333, opened 2015-05-11
PostCSS plugins website?


ben-eb:
At the current pace of development, containing every plugin in the README is going to be unsustainable. I think we need a site to collect all of the PostCSS plugins, which should offer a search functionality, and example recipes for different use cases for plugins. For example, quick benefits to using modules like `postcss-plugin-context` with `postcss-simple-vars`, or using `postcss-simple-vars` with `postcss-map`, or "building your own preprocessor" with the relevant plugins.

Some of the nice similar resources are http://perf-tooling.today/tools and http://gulpjs.com/plugins/ - but I feel like something that's not just a localised npm search would be really beneficial.

What features would you like to see in an official PostCSS plugin website?

ai:
My current plans to start building a PostCSS after Autoprefixer 5.2 release.

ai:
Also we need a PostCSS playground with ability to use any set of plugins.

MoOx:
a PostCSS what ?
website ? lang ? :)

ben-eb:
@MoOx :grinning:

whdigger:
Great idea =) and in the futures to do something like automatic travis ;)

ai:
@whdigger what do you mean?

simonsmith:
You could certainly start with something like the gulpjs list, would be better than the current README setup.

> Also we need a PostCSS playground with ability to use any set of plugins.

+1

ai:
@simonsmith BTW, why you think that Gulp list is better, that current categorized list (in current size)?

whdigger:
something like this
The user downloads the src PostCss syntax and the output gets :sparkles: css

tunnckoCore:
@whdigger erm? again? Maybe you mean online demo for converting postcss to css?

simonsmith:
@ai Mainly for the ability to filter but also being able to send someone to a proper plugins page, rather than hunting around in a large README.

ai:
@simonsmith search is very good for some case. But not for cases, when you didn't know what to search (it is a popular case in PostCSS, because it is a new tool class).

I am thinking right now, how to join benefits of categorized list with autogenerated pages like Gulp plugin list.

BTW, do you know, that you can send a anchor link to any section on GitHub's READMEs?

ben-eb:
> But not for cases, when you didn't know what to search (it is a popular case in PostCSS, because it is a new tool class)

[...]

PR #1772, opened 2022-08-22
[Feat] Fail on invalid custom prop names

phoenisx:
## Changes:

Closes: #1765 

- Changes do not use logical operators. Instead `regex` is being used for simplicity.
  - I tried benchmarking the logical operators and `regex` using `process.hrtime.bigint()`
    and there's a few 100 nanoseconds of difference between the two approaches (using 
    `regex` on a single character compared to logical operations on a single char code).
  - Please do suggest if you have any better approach for the solution.
- Minimal test cases added
  - Following manual tests were done:
    - `pcss.parse(':root { --var: #333 }')`
    - `pcss.parse(':root { --0ar: #333 }')`
    - `pcss.parse(':root { --_ar: #333 }')` and `pcss.parse(':root { ---ar: #333 }')`
    - `pcss.parse(':root { --~ar: #333 }')`, which should fail.

## Screenshots:
<p align="center">
Screenshot 2022-08-21 at 11 48 10 PM
Screenshot 2022-08-21 at 11 48 19 PM
</p>

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -12,6 +12,8 @@ const SAFE_COMMENT_NEIGHBOR = {
   space: true
 }
 
+const CUSTOM_PROP_ACCEPTED_IDENT = /[-_0-9a-z]/i;
+
 function findLastWithPosition(tokens) {
   for (let i = tokens.length - 1; i >= 0; i--) {
     let token = tokens[i]
@@ -201,6 +203,9 @@ class Parser {
       let type = tokens[0][0]
       if (type === ':' || type === 'space' || type === 'comment') {
         break
+      }
+      if (customProperty) {
+        this.checkValidCustomPropName(tokens[0]);
       }
       node.prop += tokens.shift()[1]
     }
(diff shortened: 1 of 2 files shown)


Semigradsky:
Fails on valid
```css
:root {
  --\.valid-name: green;
}
```

phoenisx:
@Semigradsky If I understand it correctly from the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/custom-ident), this change is missing support for any escaped custom property name, right?

**_Please help me understand, does this mean any name that has invalid ASCII, but is escaped with a back-slash (`\`), is an accepted property name?_**

I tried the following on Firefox:
- `--\.name: blue` (valid)
- `--\\name: blue` (valid)
- `--\~tild: blue` (valid)
- `--\😅boo: blue` (valid)

lib/parser.js
```diff
@@ -544,6 +549,20 @@ class Parser {
     )
   }
 
+  checkValidCustomPropName(token) {
```
ai, line 552: PostCSS doesn’t do validity check. You need to add this code to Stylelint.

phoenisx: Understood

No worries!
Will close this PR and the respective Issue as well.
Thanks for your time 💯

Issue #1771, opened 2022-08-18
Document how to declare a PostCSS plugin in an ES6/TypeScript module


nex3:
It's currently unclear how to properly declare a PostCSS module using ES6 module syntax (which TypeScript also uses idiomatically). The suggested `module.exports` isn't idiomatic in these systems.

ai:
```js
let plugin = () => { … }
plugin.postcss = true

export default postcss
```

Where do you think we should change docs?

nex3:
That was the first thing I tried, but it doesn't work if you're loading it from TypeScript:

```
 FAIL  test/index.test.ts
  ● Test suite failed to run

    test/index.test.ts:22:25 - error TS2349: This expression is not callable.
      Type 'typeof import("/home/nweiz/goog/pkg/js/postcss-src/build/index")' has no call signatures.

    22   return await postcss([plugin(options)]).process(input, {from: undefined});
                               ~~~~~~
```

I also tried

```ts
const plugin: PluginCreator = (opts: Options = {}): Plugin => { … };
plugin.postcss = true;

export = plugin;
```

but this produced

```
src/index.ts:80:1 - error TS2309: An export assignment cannot be used in a module with other exported elements.                       

80 export = plugin;
   ~~~~~~~~~~~~~~~~
```

So far, this is the best I've been able to manage:

```ts
const plugin: PluginCreator = (opts: Options = {}): Plugin => { … };

export = plugin;
```

...but this doesn't include the `postcss` flag, which I assume is desirable.

nex3:
It's worth noting that the `export default` syntax also doesn't work with `postcss-cli`:

```shellsession
$ npx postcss-cli -u ../../build/index.js < test.css

> postcss-src@1.0.0 compile
> tsc -p .

Plugin Error (../../build/index.js): TypeError: (intermediate value).default is not a function'
```

nex3:
Okay, I've found an incantation that works:

```ts
// eslint-disable-next-line @typescript-eslint/no-namespace
namespace plugin {
  export interface Options { … }
}

const plugin: PluginCreator<olugin.Options> = (opts = {}) => { … };
plugin.postcss = true;

export = plugin;
```

It would be nice if there were an alternative way to declare these plugins that was more module-system friendly—allowing the `PluginCreator` to be defined *as* the top-level `postcss` field, for example.

ai:
These problems are coming from ESM support in TS.

[...]

PR #1773, opened 2022-08-31
Add ESM plugins support

ai:
Closes https://github.com/postcss/postcss/issues/1771

### DIFF
--- a/docs/guidelines/plugin.md
+++ b/docs/guidelines/plugin.md
@@ -72,14 +72,15 @@ It is better even not to import `postcss`.
 
 ```diff
 - const { list, decl } = require('postcss')
-  module.exports = opts => {
+  const plugin = opts => {
     postcssPlugin: 'postcss-name',
 -   Once (root) {
 +   Once (root, { list, decl }) {
       // Plugin code
     }
   }
-  module.exports.postcss = true
+  plugin.postcss = true
+  module.exports = plugin
 ```
 
 
@@ -88,7 +89,7 @@ It is better even not to import `postcss`.
 Plugin name will be used in error messages and warnings.
 
 ```js
-module.exports = opts => {
+const plugin = opts => {
   return {
     postcssPlugin: 'postcss-name',
     Once (root) {
@@ -96,7 +97,7 @@ module.exports = opts => {
     }
   }
 }
-module.exports.postcss = true
+plugin.postcss = true
 ```
 
 
@@ -117,7 +118,7 @@ For example, use `fs.writeFile` instead of `fs.writeFileSync`:
 ```js
 let { readFile } = require('fs').promises
 
-module.exports = opts => {
+const plugin = opts => {
   return {
     postcssPlugin: 'plugin-inline',
     async Decl (decl) {
@@ -129,7 +130,7 @@ module.exports = opts => {
     }
   }
 }
-module.exports.postcss = true
+plugin.postcss = true
 ```
 
 ### 2.3. Use fast node’s scanning
@@ -137,7 +138,7 @@ module.exports.postcss = true
 Subscribing for specific node type is much faster, than calling `walk*` method:
 
 ```diff
-  module.exports = {
+  const plugin = () => ({
     postcssPlugin: 'postcss-example',
 -   Once (root) {
 -     root.walkDecls(decl => {
@@ -147,15 +148,15 @@ Subscribing for specific node type is much faster, than calling `walk*` method:
 +   Declaration (decl) {
 +     // Faster
 +   }
-  }
-  module.exports.postcss = true
+  })
+  plugin.postcss = true
 ```
 
 But you can make scanning even faster, if you know, what declaration’s property
 or at-rule’s name do you need:
 
 ```diff
-  module.exports = {
+  const plugin = () => ({
     postcssPlugin: 'postcss-example',
 -   Declaration (decl) {
 -     if (decl.prop === 'color') {
@@ -167,8 +168,8 @@ or at-rule’s name do you need:
 +       // The fastest
 +     }
 +   }
-  }
-  module.exports.postcss = true
+  })
+  plugin.postcss = true
 ```
 
 
--- a/docs/writing-a-plugin.md
+++ b/docs/writing-a-plugin.md
@@ -93,6 +93,12 @@ plugin.postcss = true
 module.exports = plugin
 ```
 
+For ESM project replace last line with:
+
+```js
+export default plugin
+```
+
 [PostCSS plugin boilerplate]: https://github.com/postcss/postcss-plugin-boilerplate/
 [plugin template]: https://github.com/postcss/postcss-plugin-boilerplate/blob/main/template/index.t.js
 
--- a/docs/writing-a-plugin.md
+++ b/docs/writing-a-plugin.md
@@ -80,10 +80,6 @@ For public plugins:
 2. Create a repository on GitHub or GitLab.
 3. Publish your code there.
 
-You can also use [our Sharec config] to keep the best practices up to date.
-Every time when you will update the config, it will update development configs
-and development tools.
-
 ```js
 const plugin = (opts = {}) => {
   // Plugin creator to check options or prepare caches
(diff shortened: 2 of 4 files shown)


nex3:
If a plugin is defined like this:

```js
export default (opts) => {
  return {
    postcssPlugin: 'postcss-name',
    Once(root) { /* ... */ }.
  };
};

export const postcss = true;
```

This doesn't quite work, because setting `i = i.default` ignores the fact that `postcss = true` was set on the *original* value of `i`. I think this is the logic you want in `normalize()`:

```js
if (i.postcss === true) {
  if (i.__esModule && i.default) {
    i = i.default()
  } else {
    i = i()
  }
} else if (i.postcss) {
  i = i.postcss
}
```

ai:
The origin idea of ESM plugins is:

```js
const plugin = () => {
  …
}
plugin.postcss = true

export default plugin
```

`postcss = true` is a mark on top of plugin.

nex3:
I think given that the recommended CJS plugin declaration is

```js
module.exports = (opts = {}) => {
  // Plugin creator to check options or prepare caches
  return {
    postcssPlugin: 'PLUGIN NAME'
    // Plugin listeners
  }
}
module.exports.postcss = true
```

people will expect the ES6 version to be

```js
export default (opts = {}) => {
  // Plugin creator to check options or prepare caches
  return {
    postcssPlugin: 'PLUGIN NAME'
    // Plugin listeners
  }
}
export const postcss = true
```

ai:
I replaced examples: https://github.com/postcss/postcss/commit/0b5728372bf57d39335e18eec09583ff34d55c74

And add explicit ESM example: https://github.com/postcss/postcss/commit/e3a67482b1a5a7d42e8739343be8c540f5b75b9e

GrapevineLin:
hi when will it be merged

ai:
@GrapevineLin when somebody will check that it is worked

nex3:
Sorry for letting this fall of my radar, @ai! Here are my test results:

## CJS

### No options

This case works great!

```js
const postcss = require('postcss');
const plugin = require(...);
postcss([plugin]).process(...);
```

### Options

Passing options to the plugin is pretty annoying. The object loaded by `require()` isn't a function, so in that case you have to write either

```js
const postcss = require('postcss');
const plugin = require(...);
postcss([plugin.default(...)]).process(...);
```

or

```js
const postcss = require('postcss');
const plugin = require(...).default;
postcss([plugin.default(...)]).process(...);
```

both of which expose the implementation of the plugin as an ESM module to the user. 

## Native ESM

### No arguments

Works great!

```js
import plugin from ...;
postcss([plugin]).process(...);
```

### Arguments

Same problem as for CJS: you can't pass arguments directly to `plugin()` because it's not a function, it's an object with a `default` field that is itself a function. It is surprising and dismaying to me that Node's native ESM support doesn't recognize the `__esModule` convention. I found https://github.com/nodejs/node/issues/40891 which covers this in more detail.

## TypeScript

### No Arguments

I get a type error from this code:

```ts
import plugin from ...;
postcss([plugin]).process(...);
```

```
test.ts:4:1 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(plugins?: AcceptedPlugin[]): Processor', gave the following error.
    Type '{ (opts?: Options): Plugin; postcss: boolean; }' is not assignable to type 'AcceptedPlugin'.
      Type '{ (opts?: Options): Plugin; postcss: boolean; }' is not assignable to type 'PluginCreator<any>'.
        Types of property 'postcss' are incompatible.
          Type 'boolean' is not assignable to type 'true'.
  Overload 2 of 2, '(...plugins: AcceptedPlugin[]): Processor', gave the following error.
    Argument of type '{ (opts?: Options): Plugin; postcss: boolean; }[]' is not assignable to parameter of type 'AcceptedPlugin'.
      Type '{ (opts?: Options): Plugin; postcss: boolean; }[]' is missing the following properties from type 'Processor': version, plugins, use, process

4 postcss([plugin]).process('a {b: src("c")}').then(result => {
  ~~~~~~~~~~~~~~~~~

  test.ts:4:10
    4 postcss([plugin]).process('a {b: src("c")}').then(result => {
               ~~~~~~
    Did you mean to call this expression?
```

This could be pretty easily solved though by adding an explicit TypeScript example that suggests writing

```ts
import type {Plugin, PluginCreator} from 'postcss';

export interface Options {
  // ...
}

const plugin: PluginCreator<Options> = opts => {
  return {
    postcssPlugin: 'postcss-name',
    Once (root) {
      // Plugin code
    }
  }
}
plugin.postcss = true
export default plugin;
```

### Arguments

This one actually works!

```ts
import plugin from ...;
postcss([plugin(...)]).process(...);
```

## `postcss-cli`

This doesn't work at all. It'll probably need to be updated separately to match this new schema.

# Conclusion

There's an irreducible problem here where the ESM polyfill simply cannot play nicely with passing arguments to default 
exports. This will _eventually_ go away in a beautiful future world when we can start distributing native ESM packages, but that's a long way away. In the meantime, I think there are three potential routes for mitigation:

1. Push the pain onto users by requiring them to _maybe_ write `.default` or maybe not, depending on whether they're using CJS or emulated/native ESM. I don't like this option.

2. Push the pain onto package authors by asking to provide two separate CJS and ESM exports. I believe Node.js supports this, but I need to investigate further to figure out how complex/painful the optimal setup would be, particularly for plugins written in typescript.

3. Support a new calling convention in PostCSS that doesn't rely on default exports. For example, you could define:

   ```ts
   export type PluginCreator<PluginOptions> = (opts?: PluginOptions): Plugin | Processor;
   
   export type PluginModule<PluginOptions> =
   | {postcssPlugin: PluginCreator<PluginOptions>}
   | PluginCreator<PluginOptions> & {postcss: true}
   ```

   ...but this does mean there are two somewhat-different ways to declare plugin modules, and it means _all_ users have to write a more verbose import (although at least it would be consistent across different module systems).

I'll investigate option 2 more and post the results.

nex3:
I've poked around some more with [conditional exports](https://nodejs.org/api/packages.html#conditional-exports). Unfortunately, there doesn't seem to be a way to export a given file _only for ESM-polyfill users_.

So unless PostCSS goes with Option 3 above, I think from a package author's perspective the best option is to accept that the old CommonJS syntax is the only thing that's universally usable and add a shim:

```js
module.exports = require('./build/index').default;
```

along with `"main": "./cjs.js"` in `package.json`. This will essentially undo the EJS polyfill and make the package act like it exports a normal CommonJS module with a `module.exports =`. You could go a step further and use conditional exports to also export a native ESM module, but there's really not a strong reason to do so since ESM can consume CJS just fine.

@ai My recommendation to you is:

* Keep the code as-is, since it works about as well as it can given the constraints.
* Add explicit TypeScript documentation like I described in my earlier comment.
* Add a note to the ESM/TS section saying "If your plugin takes arguments, you'll also need to create a `cjs.js` shim and add `"main": "./cjs.js"` to your `package.json`."
* Update `postcss-cli` to detect `__esModule` like `processor.js` does.

nex3:
Update: It turns out TypeScript users can avoid a lot of this pain by using `export =` instead of `export default`, like so:

```ts
import type {PluginCreator} from 'postcss';

namespace plugin {
  export interface Options {
    // ...
  }
}

const plugin: PluginCreator<plugin.Options> = opts => {
  return {
    postcssPlugin: 'postcss-name',
    Once (root) {
      // Plugin code
    }
  }
}
plugin.postcss = true;
export = plugin;
```

This generates a CommonJS module just like the one you'd write in raw JS, without the need for the wrapper shims I described above. Unfortunately, it's not even vaguely ESM—that `export =` construct is TypeScript-only. That said, I think this is probably what you should suggest for TypeScript users rather than `export default` since, in the end, its compiled output matches what you'd write in plain JS.

ai:
> Update postcss-cli to detect __esModule like processor.js does.

Unfortunately, `postcss-cli` is not an only tool with custom config loader. For instance, webpack and some other builders do it too.

Seems like a critical blocker for me.

GrimeyPickle:
Check

GrimeyPickle (approved):
feature/esm-plugins

ai:
I am closing this PR since we need a different approach

PR #1784, opened 2022-10-05
Move title

muddv:
Moved title to allow documentation pages to generate properly.


docs/writing-a-plugin.md
```diff
@@ -1,3 +1,4 @@
+# Writing a PostCSS Plugin
```
ai, line 1: Should we remove title on line 13?

muddv: Yes, definitely. Done now.

PR #1788, opened 2022-10-12
fix: `absolute: true` source maps will preserve source content

KingSora:
Good day!

I noticed a bug within `postcss` where the `sourceContent` field is `[null]` if I'm using the `absolute: true` option. This happens, because if I'm using `absolute: true` all the `source` [paths are converted to file urls](https://github.com/postcss/postcss/blob/main/lib/map-generator.js#L236-L245) With this change the internal mapping is not correct anymore because we compare file paths with file urls. 

With this change this behavior should be fixed, I've also added a test to reflect the changes.


ai:
Looks good. I will try to merge it today evening.

KingSora:
@ai thanks! :)

ai:
Released in 8.4.18.

PR #1811, opened 2023-01-05
Add types for an Input.error()

hudochenkov:
Something I found when was developing [postcss-styled-syntax](https://github.com/hudochenkov/postcss-styled-syntax).

I didn't find a good way to test if types are correct.

### DIFF
--- a/lib/input.d.ts
+++ b/lib/input.d.ts
@@ -1,4 +1,4 @@
-import { ProcessOptions } from './postcss.js'
+import { CssSyntaxError, ProcessOptions } from './postcss.js'
 import PreviousMap from './previous-map.js'
 
 export interface FilePosition {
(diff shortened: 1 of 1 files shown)


ai:
Will look and take tomorrow

ai:
I fixed types a little c9bd4452

ai:
Thanks. Released in 8.4.21.

hudochenkov:
Thanks!

How do you test all types in PostCSS? `check-dts` doesn't catch all of them. And there is nothing like `tsconfig` with `allowJs`.

ai:
> How do you test all types in PostCSS?

By writing tests in TS (not case for this method, since it is not really “public”, most users should use `Node#error()`)

Issue #1145, opened 2018-05-01
Native selector and value parsing in PostCSS 7


jonathantneal:
I would like to make an ambitious proposal. In PostCSS 7, could we parse selectors and values, or provide that ability out-of-the-box? If so, how could I help?

At the lowest effort, could we integrate [postcss-selector-parser](https://github.com/postcss/postcss-selector-parser) and [postcss-values-parser](https://github.com/shellscape/postcss-values-parser)?

At a higher effort, could we integrate the tokenizers? And have one tokenization _to rule them all_?

ai:
In PostCSS 8 ;).

PostCSS 7 will be a quick release without Node.js 4.

The only thing blocking us from PostCSS 7 is cssnano 4 release by @evilebottnawi

ai:
Technically we don’t need a major release for it. We can release it in 7.1.

I suggest starting with selectors. And add values parser only in 7.2.

ai:
How you can help:

* I need a list of AST nodes for selectors.
* I need a token list to parse those AST nodes.
* I need ideas of AST API to change these nodes. You had a great experience with plugins. What API will be compact and easy?

chriseppstein:
The postcss-selector-parser api is really weird when it comes to handling raws (escapes, comments,  and intervening whitespace). Sometimes comments are their own node, sometimes comments are part of raws, etc. I have a design in my head that is much better and more uniform that I would like to build out that I think should be the foundation of a first-class postcss api for selectors.

There's a lossy parse that trims whitespace but preserves comments but it can only be done at parse time (I think there should be methods to normalize/minify whitespace, prune comments, etc instead or as an option during stringification). Nodes didn't maintain internal consistency across mutation, and there's no sourcemap support to speak of.

I'd like to write up a design proposal soon.

ai:
@chriseppstein thanks, This is very useful feedback.

> Sometimes comments are their own node, sometimes comments are part of raws, etc.

Let’s make the rule for new selector/value API. Comments must be a separated node.

But whitespaces we must put inside nodes to be compatible with other PostCSS API.

> Nodes didn't maintain internal consistency across mutation

Can you show an example?

> There's a lossy parse that trims whitespace but preserves comments but it can only be done at parse time
> there's no sourcemap support to speak of.

Yeap, new parser will continue PostCSS best practices. Keep all bytes of origin input. Support source map.

chriseppstein:
> Comments must be a separated node.

[...]

PR #1812, opened 2023-01-10
switch tokenizer

romainmenke:
see : https://github.com/postcss/postcss/issues/1145

Intro : 

This change is not intended to be merged as it exists today.
It is importing the csstools tokenizer package but ideally the tokenizer is "owned" by PostCSS.

The goals is to surface possible issues early.

--------

There are 3 test failures

2 are related to `!important`.
In the existing tokenizer `!important` is one word.
A declaration will have different contents in `raws` depending on the `!important` vs `! /* a comment */ important`.

In the new tokenizer this is a delim and ident token.
Rewriting the algorithm to match the old output is possible, but it wasn't trivial.

1 is related to `@`.
The existing tokenizer has specific handling for a lone `@` and throws.
In the new tokenizer doesn't throw any error as it is just a delim token which is valid CSS.

-------

The existing tokenizer also has a mechanic to "return" tokens.
This is missing from the new tokenizer and requires slow and hacky code to patch back in.

These hacks would not be needed if the tokenizer was written for PostCSS within this repository.


lib/parser.js
```diff
@@ -35,36 +72,69 @@ class Parser {
...
   parse() {
     let token
-    while (!this.tokenizer.endOfFile()) {
-      token = this.tokenizer.nextToken()
+    while (!this.tokenizer.endOfFileWithBack()) {
+      token = this.tokenizer.nextTokenWithBack()
 
       switch (token[0]) {
-        case 'space':
+        case 'whitespace-token':
```
ai, line 117: What do you think of using numbers like `token[0] === Tokens.whitespace` for a better performance and memory consumption?

test/parse.test.ts
```diff
@@ -141,6 +141,8 @@ test('throws on property without value', () => {
   }, /:1:5: Unknown word/)
 })
 
+// TODO : this is just a delim token and not an error
+// Do we make it an error?
 test('throws on nameless at-rule', () => {
```
ai, line 146: Not really, it is just a 100% coverage test

lib/parser.js
```diff
@@ -35,36 +72,69 @@ class Parser {
...
+
+          default:
+            throw this.input.error(err.message, { offset: err.sourceStart }, { offset: err.sourceEnd + 1 })
+        }
+      }
+    })
+
+    let returned = [];
+
+    this.tokenizer.nextTokenWithBack = () => {
```
ai, line 92: Feel free to change parser and avoid `nextToken` call.

We used it because our current tokenizer is stream less (it has no array of tokens and just pass next token to parser). Stream tokenizer is better for memory consumption but anyway will not work for our future plans to put tokens to AST.

test/parse.test.ts
romainmenke: Good to know!

After spending more timing thinking about it I think it might need to remain an error.
I absolutely want to avoid changing how PostCSS parses CSS.

If this is not an error `@ {}` would become a rule with `@` as a selector.
Throwing on `@` ensures that we reserve this syntax space for any future expansion in the CSS specification.

romainmenke:
There is only one discrepancy left between PostCSS with the old tokenizer and this new version : https://github.com/postcss/postcss-parser-tests/issues/23

My next step will be to add the tokenizer in this repo as source, not as an external dependency. That will allow us to more easily apply the suggestions for better performance.

romainmenke:
tokenizer progress : https://github.com/romainmenke/postcss/pull/1

test/parse.test.ts
romainmenke: This still throws and that is ok

test/tokenize.test.js
```diff
@@ -275,47 +372,45 @@ test('ignores unclosing function on request', () => {
...
+//     ['space', ' '],
+//     ['word', 'it', 6, 7],
+//     ['space', ' '],
+//     ['word', 'going', 9, 13],
+//     ['space', ' '],
+//     ['(', '(', 15]
+//   ]
+
+//   equal(tokens, expected)
+// })
```
romainmenke, line 413: @ai Before I dive into this I wanted to know if this feature is currently in use?
Outside of this test I couldn't find anything that uses the per token ignoring of error.s

romainmenke:
Current status :

- the tokenizer code has been added to this repository
- some final tweaks to the parser were needed
- highlighting of CSS syntax errors needed to be updated
- test coverage is back up to 100 (lines)

At this time it would be good to have a review of the changes in the tokenizer tests.
The intention remains that this change to the tokenizer must not be noticeable to anyone using PostCSS. I want to make sure I haven't overlooked anything.

test/tokenize.test.js
ai: It was needed for `postcss-less` https://github.com/postcss/postcss/pull/1191

test/tokenize.test.js
```diff
@@ -202,32 +249,40 @@ test('tokenizes CSS', () => {
...
+    ['whitespace-token', ' ', 14, 14, undefined],
+    ['string-token', '"a"', 15, 17, { value: 'a' }],
+    ['semicolon-token', ';', 18, 18, undefined],
+    ['whitespace-token', '\n  ', 19, 21, undefined],
+    ['ident-token', 'width', 22, 26, { value: 'width' }],
+    ['colon-token', ':', 27, 27, undefined],
+    ['whitespace-token', ' ', 28, 28, undefined],
+    ['function-token', 'calc(', 29, 33, { value: 'calc' }],
+    [
+      'dimension-token',
```
ai, line 267: Soo long. Do we really need `-token` prefix for all tokens? :)

ai:
Can you also test the performance by https://github.com/postcss/benchmark

It could take current and dev version of `postcss` (Just put `postcss/` next to `postcss-benchmark/`).

romainmenke:
benchmarks :

### preprocessors

```
PostCSS:           26 ms
PostCSS sync:      26 ms (1.0 times slower)
Next PostCSS:      29 ms (1.1 times slower)
Next PostCSS sync: 29 ms (1.1 times slower)
LibSass sync:      40 ms (1.5 times slower)
Less:              40 ms (1.5 times slower)
LibSass:           41 ms (1.6 times slower)
Dart Sass sync:    53 ms (2.0 times slower)
Dart Sass:         97 ms (3.7 times slower)
```

### parsers

```
Stylis:       3 ms  (3.2 times faster)
CSSOM:        10 ms (1.0 times faster)
PostCSS:      11 ms
CSSTree:      11 ms (1.0 times slower)
Mensch:       11 ms (1.1 times slower)
Next PostCSS: 13 ms (1.2 times slower)
Rework:       15 ms (1.4 times slower)
Stylecow:     24 ms (2.3 times slower)
PostCSS Full: 32 ms (3.0 times slower)
Gonzales:     51 ms (4.8 times slower)
ParserLib:    53 ms (5.0 times slower)
```

### prefixers

```
Lightning CSS:    4 ms   (8.3 times faster)
Stylis:           5 ms   (6.3 times faster)
Autoprefixer:     33 ms
Autoprefixer dev: 36 ms  (1.1 times slower)
Stylecow:         183 ms (5.6 times slower)
```

test/tokenize.test.js
romainmenke: No we do not :)

These names are used throughout the CSS specification.
Using them during initial development helps me implement things correctly.
Then I can just read the specification without having to mentally map the wording.

We should change this to what is best for performance and as an API surface later.

lib/parser.js
```diff
@@ -20,6 +21,42 @@ function findLastWithPosition(tokens) {
   }
 }
 
+function tokenIsWordLike(tokenType) {
```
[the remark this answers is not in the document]
romainmenke: I've tested multiple variations and their might be a small advantage but this is negligible overal.

lib/parser.js
romainmenke: Performance and memory consumption is the same with strings or numbers.
String literals are interned by V8 and other JS engines so these are pointer comparisons.

lib/parser.js
romainmenke: Currently the tokenizer allocates a large array to store all the code point values for each character from the source string.

This allocation takes a lot of time.

I've tried a variation where the tokenizer is fully streaming and only keeps 4 code points as state.

This is a little bit slower because that approach results in more value changes but it keeps memory consumption a bit slower.

I did not see a notable difference between small and very large CSS sources with either approach.

---------

I haven't yet tried the other way.
To completely remove the streaming aspect.

This would allocate a lot more memory, but it would reduce CPU cache misses.
At this time the parser is the driver. So we switch a lot between the functions and logic for the tokenizer and parser.

Doing everything separately would have benefits.
This is similar to "tiling" in image processing.

These kinds of optimizations are very difficult to do without benchmarking on a wide range of different hardware. CPU cache size, memory and IO latency greatly affect the outcome. One approach might be faster for me, but slower for most users overal.

PR #1815, opened 2023-02-16
Fix type definitions

### LINKED ISSUE #1814: "module": "node16" error: `This expression is not callable`
Hi there 👋 First of all, thanks for your continued effort on PostCSS and the surrounding community, really amazing!

Using the `"module": "node16"` configuration option in `tsconfig.json` within an ES Module causes the following TypeScript error:

```bash
$ yarn tsc
index.ts:3:1 - error TS2349: This expression is not callable.
  Type 'typeof import("/home/projects/node-vqx4zh/node_modules/postcss/lib/postcss")' has no call signatures.

3 postcss([]);
  ~~~~~~~


Found 1 error in index.ts:3

error Command failed with exit code 2.
```

Demo on StackBlitz - run `yarn tsc` in the Terminal at the bottom to reproduce:

https://stackblitz.com/edit/node-vqx4zh?file=tsconfig.json,package.json,index.ts

Screenshot 2023-01-29 at 18 30 24

## Proposed Solution

One way that seems to work to fix the problem is to use `export = postcss` instead of `export default postcss` here:

https://github.com/postcss/postcss/blob/8834314ff334ea8321bf2c83934f276d379512cd/lib/postcss.d.ts#L473

But maybe there is a more elegant way of doing this...

cc @andrewbranch in case I still have the proposed solution wrong over here

## Background

See more information about the problem of misconfigured declaration files in packages here:

- https://github.com/microsoft/TypeScript/issues/49160#issuecomment-1137482639
- https://github.com/microsoft/TypeScript/issues/49189#issuecomment-1137756847

remcohaszing:
Many files from postcss are exported. Each of this file is written using the syntax:

```js
module.exports = Thing
```

This should be typed as:

```ts
export = Thing
```

If such a file needs to export additional types, a namespace should be used.

However, it was typed as:

```ts
export default thing
```

This was solved at the time in #1459 (by me) by adding the following code to each module:

```js
Thing.default = Thing
```

Whilst affective at the time, this was not the correct solution. The introduction of the `"module": "node16"` option in TypeScript made this issue more apparent. I recommend to remove this in a next semver major release.

Usage has changed as follows:

**cts**

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "module": "commonjs",
    "moduleResolution": "node16"
  }
}
```

```ts
// Old
import postcss = require('postcss')

postcss.default(/* … */)
```

```ts
// New
import postcss = require('postcss')

postcss(/* … */)
```

**cjs**

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "checkJs": true,
    "module": "commonjs",
    "moduleResolution": "node16"
  }
}
```

```ts
// Old

postcss.default(/* … */)
```

```ts
// New

postcss(/* … */)
```

**mts** / **mjs**

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "module": "node16"
  }
}
```

```ts
// Old
import postcss from 'postcss'

postcss.default(/* … */)
```

```ts
// New
import postcss from 'postcss'

postcss(/* … */)
```

**bundler**

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "allowSyntheticDefaultImports": true,
    "module": "esnext",
    "moduleResolution": "node16"
  }
}
```

```ts
// Unchanged
import postcss from 'postcss'

postcss(/* … */)
```

**ES module interop** (Not recommended for library code!)

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "esModuleInterop": true,
    "module": "commonjs",
    "moduleResolution": "node16"
  }
}
```

```ts
// Unchanged
import postcss from 'postcss'

postcss(/* … */)
```

When compiling to CJS, it is still possible to write ESM-like syntax, for example:

```jsonc
// Same as import postcss = require('postcss')
import * as postcss from 'postcss'
// Named imports from namespaces are allowed
import { root, Root } from 'postcss'
```

The following changes were made to `tsconfig.json`:

- `lib` was specified explicitly. By default, TypeScript includes the `dom` types. This is not desirable, it even causes some conflicts with types used by Postcss.
- `allowJs` was removed. When this is enabled, TypeScript treats `.d.ts` files matching `.js` files as output. This caused some unexpected issues where TypeScript simply ignored type errors.
- `skipLibCheck` was removed. Ideally all types just work together. This can be added again later if necessary.
- `esModuleInterop` was removed. Enabling this option causes TypeScript to allow code that is otherwise invalid. This may be useful for CLIs or website, but is dangerous to use for libraries.

Correcting these options also highlighted some issues in the tests. These have been resolved. `nanodelay@1` has incorrect type definitions, so it has been marked using `@ts-expect-error` comments.

In `test/errors.ts` were some type errors. These were marked as `THROWS`, which signals `check-dts` a type error is expected. This was changed to `@ts-expect-error`, which works too, but it also works with other TypeScript tooling.

The `Postcss` type was missing some properties. As a side effect of the corrected types, these are now available. This also affects the `Helper` type.

Last, but not least, type definitions have been added for `postcss.mjs`.

Closes #1814

### DIFF
--- a/lib/container.d.ts
+++ b/lib/container.d.ts
@@ -1,8 +1,8 @@
-import Node, { ChildNode, NodeProps, ChildProps } from './node.js'
-import Declaration from './declaration.js'
-import Comment from './comment.js'
-import AtRule from './at-rule.js'
-import Rule from './rule.js'
+import Node = require('./node.js')
+import Declaration = require('./declaration.js')
+import Comment = require('./comment.js')
+import AtRule = require('./at-rule.js')
+import Rule = require('./rule.js')
 
 interface ValueOptions {
   /**
@@ -16,8 +16,10 @@ interface ValueOptions {
   fast?: string
 }
 
-export interface ContainerProps extends NodeProps {
-  nodes?: (ChildNode | ChildProps)[]
+declare namespace Container {
+  interface ContainerProps extends Node.NodeProps {
+    nodes?: (Node.ChildNode | Node.ChildProps)[]
+  }
 }
 
 /**
@@ -27,8 +29,8 @@ export interface ContainerProps extends NodeProps {
  * Note that all containers can store any content. If you write a rule inside
  * a rule, PostCSS will parse it.
  */
-export default abstract class Container<
-  Child extends Node = ChildNode
+declare abstract class Container<
+  Child extends Node = Node.ChildNode
 > extends Node {
   /**
    * An array containing the container’s children.
@@ -117,7 +119,7 @@ export default abstract class Container<
    * @return  Returns `false` if iteration was broke.
    */
   walk(
-    callback: (node: ChildNode, index: number) => false | void
+    callback: (node: Node.ChildNode, index: number) => false | void
   ): false | undefined
 
   /**
@@ -269,7 +271,7 @@ export default abstract class Container<
    * @return This node for methods chain.
    */
   append(
-    ...nodes: (Node | Node[] | ChildProps | ChildProps[] | string | string[])[]
+    ...nodes: (Node | Node[] | Node.ChildProps | Node.ChildProps[] | string | string[])[]
   ): this
 
   /**
@@ -293,7 +295,7 @@ export default abstract class Container<
    * @return This node for methods chain.
    */
   prepend(
-    ...nodes: (Node | Node[] | ChildProps | ChildProps[] | string | string[])[]
+    ...nodes: (Node | Node[] | Node.ChildProps | Node.ChildProps[] | string | string[])[]
   ): this
 
   /**
@@ -321,7 +323,7 @@ export default abstract class Container<
    */
   insertBefore(
     oldNode: Child | number,
-    newNode: Child | ChildProps | string | Child[] | ChildProps[] | string[]
+    newNode: Child | Node.ChildProps | string | Child[] | Node.ChildProps[] | string[]
   ): this
 
   /**
@@ -333,7 +335,7 @@ export default abstract class Container<
    */
   insertAfter(
     oldNode: Child | number,
-    newNode: Child | ChildProps | string | Child[] | ChildProps[] | string[]
+    newNode: Child | Node.ChildProps | string | Child[] | Node.ChildProps[] | string[]
   ): this
 
   /**
@@ -440,3 +442,5 @@ export default abstract class Container<
    */
   index(child: Child | number): number
 }
+
+export = Container
--- a/lib/container.d.ts
+++ b/lib/container.d.ts
@@ -17,9 +17,13 @@ interface ValueOptions {
 }
 
 declare namespace Container {
-  interface ContainerProps extends Node.NodeProps {
+  export interface ContainerProps extends Node.NodeProps {
     nodes?: (Node.ChildNode | Node.ChildProps)[]
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-shadow
+  class Container<Child extends Node = Node.ChildNode> extends Container_<Child> {}
+  export { Container as default }
 }
 
 /**
@@ -29,11 +33,9 @@ declare namespace Container {
  * Note that all containers can store any content. If you write a rule inside
  * a rule, PostCSS will parse it.
  */
-declare abstract class Container<
+declare abstract class Container_<
   Child extends Node = Node.ChildNode
 > extends Node {
-  static default: typeof Container
-
   /**
    * An array containing the container’s children.
    *
@@ -444,5 +446,7 @@ declare abstract class Container<
    */
   index(child: Child | number): number
 }
+
+declare class Container<Child extends Node = Node.ChildNode> extends Container_<Child> {}
 
 export = Container
--- a/lib/container.d.ts
+++ b/lib/container.d.ts
@@ -21,9 +21,8 @@ declare namespace Container {
     nodes?: (ChildNode | ChildProps)[]
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-shadow
-  class Container<Child extends Node = ChildNode> extends Container_<Child> {}
-  export { Container as default }
+  // eslint-disable-next-line @typescript-eslint/no-use-before-define
+  export { Container_ as default }
 }
 
 /**
--- a/lib/container.d.ts
+++ b/lib/container.d.ts
@@ -32,6 +32,8 @@ declare namespace Container {
 declare abstract class Container<
   Child extends Node = Node.ChildNode
 > extends Node {
+  static default: typeof Container
+
   /**
    * An array containing the container’s children.
    *
--- a/lib/container.d.ts
+++ b/lib/container.d.ts
@@ -1,8 +1,8 @@
-import Node = require('./node.js')
-import Declaration = require('./declaration.js')
-import Comment = require('./comment.js')
-import AtRule = require('./at-rule.js')
-import Rule = require('./rule.js')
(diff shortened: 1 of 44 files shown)


remcohaszing:
FWIW I uploaded the package to https://arethetypeswrong.github.io and it’s all green (except for the missing types). ✅

test/errors.ts
```diff
@@ -1,17 +1,17 @@
 import { PluginCreator } from '../lib/postcss.js'
 
 const plugin: PluginCreator<{ a: number }> = opts => {
-  // THROWS 'opts' is possibly 'undefined'
+  // @ts-expect-error 'opts' is possibly 'undefined'
```
ai, line 4: `THROWS` is a special command by `check-dts`

`@ts-expect-error` doesn’t check the type error text

lib/container.d.ts
```diff
@@ -1,8 +1,8 @@
-import Node, { ChildNode, NodeProps, ChildProps } from './node.js'
-import Declaration from './declaration.js'
-import Comment from './comment.js'
-import AtRule from './at-rule.js'
-import Rule from './rule.js'
+import Node = require('./node.js')
```
ai, line 1: Did you change tests because it is breaking change?

package.json
```diff
@@ -41,7 +40,6 @@
     "./package.json": "./package.json"
   },
   "main": "./lib/postcss.js",
-  "types": "./lib/postcss.d.ts",
```
ai, line 44: Some online tools tell that packages has type support because of these fields.

Please revert it. PostCSS's releases should be very small because could lead to breaking changes by a very small changes.

remcohaszing: I was afraid this might cause some issues, but I was wrong.

It used to be necessary to get the TS icon on https://npmjs.com. This is no longer the case though, so it might be unnecessary now.

Anyway, I reverted it.

lib/container.d.ts
remcohaszing: This changed because of a combination of two things:

1. `esModuleInterop` was disabled. Without it, this would be hidden.
2. The same changes that are mentioned in the PR description, apply to the nested imports. I consider this to be a bug fix, but it does change behaviour (of types only).

I ran `grep postcss/lib -r node_modules --include='*.ts'` in a repository that has quite a lot of postcss dependencies. The only match that came up was https://github.com/postcss/postcss-load-config/blob/main/src/index.d.ts#L3. This is quite easy to resolve.

ai: I can’t allow breaking changes in types too.

PostCSS has 76M downloads per week. I do not want to lose a few of my days fixing issue with custom PostCSS plugins of teams.

We need to find a solution without breaking changes.

remcohaszing: It took some more effort to make this work than I anticipated. I even got some help from a TypeScript team member and learnt some new things about TypeScript needed to fix this.

It now works as expected for all situations, while maintaining backwards compatibility with previous workarounds.

To test it, run `npm pack` to pack postcss, then Create a new project with the following files:

`package.json`

```js
{
  "type": "module",
  "dependencies": {
    "postcss": "file:../postcss/postcss-8.4.21.tgz",
    "typescript": "^4.9.5"
  }
}
```

`tsconfig.json`

```jsonc
{
  "compilerOptions": {
    "module": "node16",

    // "module": "commonjs",
    // "moduleResolution": "node",

    // "module": "commonjs",
    // "moduleResolution": "node16",

    // "module": "esnext",
    // "moduleResolution": "node",

    "lib": ["es2018"],
    "target": "es2018",
    "noEmit": true
  }
}
```

`test.ts`

```ts
import postcssCjs = require('postcss')
import postcssEsm, {Root} from 'postcss'
import RootCjs = require('postcss/lib/root')
import RootEsm from 'postcss/lib/root'

postcssCjs()
postcssEsm()
postcssCjs.default()
postcssEsm.default()
postcssCjs.default.default()
postcssEsm.default.default()

const root = new Root()
declare const thing: unknown

if(thing instanceof Root) {
 thing.clone()
}

const rootCjs = new RootCjs()
new RootCjs.default.default.default()
declare const thingCjs: unknown

if(thingCjs instanceof RootCjs) {
  thingCjs.clone()
}

const rootEsm = new RootEsm()
new RootEsm.default.default.default()
declare const thingEsm: unknown

if(thingEsm instanceof RootEsm) {
  thingEsm.clone()
}
```

Feel free to fiddle around with it.

As an additional step to verify the steps are right, you test the package on https://arethetypeswrong.github.io.

test/css-syntax-error.test.ts
```diff
@@ -1,12 +1,13 @@
...
 import { pathToFileURL } from 'url'
-import stripAnsi from 'strip-ansi'
+import stripAnsi = require('strip-ansi')
 import Concat from 'concat-with-sourcemaps'
 import { test } from 'uvu'
 import { is, equal, match, type } from 'uvu/assert'
 
-import postcss, {
+import postcss = require('../lib/postcss.js')
+import {
```
remcohaszing, line 10: Technically changes like this could be reverted now in tests. I think it’s best to keep it though, as it’s more important to verify `require('postcss')` is correct than to verify `require('postcss').default` is correct.

ai: Can we revert them to be sure that public API was not changed?

Or use only in a single additional test.

remcohaszing: Apparently it did change in a way I did not anticipate. I don't know how to fix this, so [I asked for help on Twitter](https://mobile.twitter.com/remcohaszing/status/1640292667857747968).

remcohaszing:
Ok, everything seems to work now. Now my main problem is, it works in CI, but not on my laptop:

```
✖ Check types

✖ node_modules/.pnpm/@types+node@18.11.18/node_modules/@types/node/globals.d.ts:67:13: Type error TS2502
  'AbortController' is referenced directly or indirectly in its own type annotation.

✖ node_modules/.pnpm/@types+node@18.11.18/node_modules/@types/node/globals.d.ts:74:13: Type error TS2502
  'AbortSignal' is referenced directly or indirectly in its own type annotation.
```

Anyway, good enough for me. I’ll perform some cleanups.

test/css-syntax-error.test.ts
remcohaszing: It should be good to go now.

ai:
I asked another users with bit TS codebase try this PR. The first feedback https://github.com/postcss/postcss/issues/1830

```
TS4058: Return type of exported function has or is using name 'CommentRaws' from external module "/projects/postcss-preset-env-mono-repo/node_modules/postcss/lib/comment" but cannot be named.
```

Any idea how to fix it?

jathak:
Hi! I'm from the Google Sass team and we use PostCSS via TypeScript internally.

From the limited tests I've been able to do (about to go on vacation, so probably won't have time to do anything more thorough any time soon), this looks like it should be a big improvement for us. Our infrastructure has some issues handling `export default`, so since PostCSS 8 we've had to manually patch the types to work around that, but with this PR it looks like they should work for us without any changes.

Looking forward to seeing this land!

remcohaszing:
Thanks for the feedback @jathak!

ai:
Thanks for the huge work.

I will release it tomorrow to have a day in case of any problem (now it is a night in my city).

ai:
Released in 8.4.22.

remcohaszing:
For a next major release I recommend to remove the default exports, as that has been the source of all problems. I’ll gladly help with that, so feel free to ping me then.

PR #1816, opened 2023-02-16
Ignore .tgz files

remcohaszing:
Running `npm pack` will generate this file. Running `npm pack` again would include the previously built package in the package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ coverage/
 
 docs/api/index.html
 docs/api/assets/
+*.tgz
--- a/.npmignore
+++ b/.npmignore
@@ -3,3 +3,4 @@ test/
 
 docs/
 tsconfig.json
+*.tgz


ai:
Why do you need to run `npm pack`?

We can add an extra line, but I need a rational reason for it (in a way like “it will improve this type of work around the project”).

remcohaszing:
It’s one of the output files from pnpm / yarn / pnpm. It’s relatively underused, but still part of basic package manager usage. Basically `node_modules/`, `*.log` and `*.tgz` should be the default files to ignore for JavaScript related projects.

`npm pack` will create the npm package that can be published to npm. Some reasons to use it are:

- The package can be introspected, e.g. useful for troubleshooting issues with `.npmignore` / the `files` field.
- The built package can be installed in a local project, e.g. to make sure all exports work as expected.
- The package can be built as a build artifact in CI, which can then be downloaded for further introspection.
- Postcss doesn’t have a build step, but for packages that do have a build step, separating the build step and the publish step is more secure, as the build step should not have access to the npm token.

ai:
We do not need to use `.tgz` for local projects, since for the projects without build step you can specify path to the project `"postcss": "../postcss"` in dependencies.

Also, without build step, you do not need a debugging.

Looks like all of these steps are for the project with build step, not for this project.

I have a rule to not add to `.gitignore` things needed only for specific developers.

If you think that all projects should have this line in ignore, you can add it to local git ignore.

remcohaszing:
> We do not need to use `.tgz` for local projects, since for the projects without build step you can specify path to the project `"postcss": "../postcss"` in dependencies.
>
> Also, without build step, you do not need a debugging.
>
> Looks like all of these steps are for the project with build step, not for this project.

This will create a symlink. This means files will be available that might not be available in the package that will be published to npm. Also this means `node_modules` will be resolved from the symlinked postcss module, instead of as if the package would be installed from npm.

Also packing can be considered a build step. Note that packing doesn’t just mean running `npm pack`. Publishing runs pack first (in memory), then publishes the tarball.

> I have a rule to not add to `.gitignore` things needed only for specific developers.
>
> If you think that all projects should have this line in ignore, you can add it to local git ignore.

I agree 100% that a project’s `.gitignore` should only contain project specific files, and that users should store patterns matching their own environment (`.vscode`, `.idea`, `*.swp`, `.DS_Store`, etc.) in their personal git ignore file.

The point on which our opinions differ, is how to categorize `.tgz` files. The way I see it, this project uses pnpm. pnpm generates `node_modules`, `.tgz`, and `.log` files in certain situations. Therefor these files are bound to this project, not the local user environment, so they belong in the project’s `.gitignore`.

---

Anyway, I just wanted to share these insights. Whether or not to add this, is not my decision to make. :)

PR #1824, opened 2023-03-27
Add JSDoc type for config

karlhorky:
First of all, thanks so much for PostCSS and the ecosystem, amazing tool! 🙌

Just wanted to surface the JSDoc types in the config recommendation to allow for autocompletion. (also type-checking if you configure `checkJs: true` in `tsconfig.json`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,12 +241,15 @@ module.exports = {
 Then create `postcss.config.js`:
 
 ```js
-module.exports = {
+/** @type {import('postcss-load-config').Config} */
+const config = {
   plugins: [
     require('autoprefixer'),
     require('postcss-nested')
   ]
 }
+
+module.exports = config
 ```
 
 [`astroturf`]: https://github.com/4Catalyzer/astroturf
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -241,12 +241,15 @@ module.exports = {
...
-module.exports = {
+/** @type {import('postcss-load-config').Config} */
+const config = {
   plugins: [
     require('autoprefixer'),
     require('postcss-nested')
   ]
 }
+
+module.exports = config
```
karlhorky, line 252: The reason that this is a named variable now and not just `module.exports =` directly after the JSDoc line is this bug in TypeScript:

-

ai:
That section is about the CSS-in-JS solution. Should we add separated TypeScript section?

karlhorky:
> That section is about the CSS-in-JS solution

Ah, didn't see that there are multiple examples with `postcss.config.js` - I've added the types to all of them in 8f237bf3d96718bef367135aff17e53c2234ab1d.

I would suggest this approach as opposed to a separate TypeScript section because it can be beneficial to users who do not use TypeScript as well.

karlhorky:
Thanks for the review and merge!

PR #1832, opened 2023-04-14
Update lockfile

remcohaszing:
A new version of pnpm has been released. This causes the postcss CI to fail. A solution is to update the lockfile.

All changes were generated by running `pnpm install`.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,88 +1,104 @@
-lockfileVersion: 5.4
-
-specifiers:
-  '@logux/eslint-config': ^49.0.0
-  '@size-limit/preset-small-lib': ^8.2.4
-  '@types/node': ^18.15.10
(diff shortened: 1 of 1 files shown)


ai:
I also fixed Node.js 14 support https://github.com/postcss/postcss/commit/a2c432eca1d1a2b4a73c7c4990fed9dbe84f9870

remcohaszing:
Huh? But doesn’t that mean this PR should be unnecessary? Also CI fails on main now.

ai:
> But doesn’t that mean this PR should be unnecessary?

We need to update lockfile since I locked on new `pnpm 8`

> Also CI fails on main now.

Yeap, I need another fix for `main` https://github.com/postcss/postcss/commit/c85028bb27b562861240f8bedc9d65d49e470754

PR #1837, opened 2023-05-15
Create SECURITY.md

OlyaDzh:


ai:
We are using Tidelyft for such thing. They have a current support versions (only latest) and provide [security communications](https://github.com/postcss/postcss#security-contact)

PR #1849, opened 2023-07-02
Improved documentation & types

vikaskaliramna07:
1. Improved documentation for lib/node.d.ts, lib/comment.d.ts and lib/declaration.d.ts
2. Changed return type for Node.clone() from `this` to `Node`, reason: `this` is the reference to the original node, and having type of original node instance on a duplicated node instance is not correct method. [click here to view](https://github.com/postcss/postcss/compare/main...vikaskaliramna07:postcss:main#diff-dd4f331e7d6fb4f40e6e8ca5d3940029a58cf47249304f93fd8492d71dfebe65R353)


lib/node.d.ts
```diff
@@ -253,51 +265,54 @@ declare abstract class Node_ {
...
-   * @return Created warning object.
+   * @return `Warning` instance is returned
    */
-  warn(result: Result, text: string, opts?: WarningOptions): Warning
+  warn(result: Result, message: string, options?: WarningOptions): Warning
 
   /**
-   * Removes the node from its parent and cleans the parent properties
-   * from the node and its children.
+   * The Node#remove method removes the node from its parent
```
ai, line 296: `Node#remove` will not look beautiful. Let’s avoid it.

lib/node.d.ts
```diff
@@ -253,51 +265,54 @@ declare abstract class Node_ {
    * }
    * ```
    *
-   * @param message Error description.
-   * @param opts    Options.
+   * @param message Description for the error instance.
+   * @param options Options for the error instance.
    *
-   * @return Error object to throw it.
+   * @return `CssSyntaxError` instance is returned.
```
ai, line 271: In TS docs we will get ugly `CssSyntaxError CssSyntaxError instance is returned`. Let’s rephrase.

lib/node.d.ts
vikaskaliramna07: should I re-write it as `Node.remove` or `{@linkcode Node.remove}`

ai: Just `It removes`

ai: And you need to change it in all other places too in your PR

vikaskaliramna07: sure

lib/comment.d.ts
```diff
@@ -30,18 +30,23 @@ declare namespace Comment {
   export { Comment_ as default }
 }
 
+
 /**
- * Represents a comment between declarations or statements (rule and at-rules).
+ * `Comment` represents a class that handles CSS comments.
```
ai, line 35: Remove the class name here as well. Just use `It`.

lib/node.d.ts
vikaskaliramna07: done

lib/node.d.ts
```diff
@@ -253,51 +265,54 @@ declare abstract class Node_ {
...
+   * @param options Options for the error instance.
    *
-   * @return Error object to throw it.
+   * @return Error instance is returned.
    */
   error(message: string, options?: Node.NodeErrorOptions): CssSyntaxError
 
   /**
-   * This method is provided as a convenience wrapper for `Result#warn`.
+   * The Node.warn method is a wrapper for Result#warn,
```
ai, line 276: `Node.warn` will not work too. Just refer it as `It`.

lib/declaration.d.ts
```diff
@@ -66,57 +70,74 @@ declare class Declaration_ extends Node {
   raws: Declaration.DeclarationRaws
 
   /**
-   * The declaration's property name.
+   * The `prop` property represents name of
```
ai, line 73: In API docs we will have `prop string The `prop` property`

Remove duplication.

lib/comment.d.ts
vikaskaliramna07: I am just trying to write documentation that someone who reads first time would feel comfortable.
Keep it or remove it?

ai: Check how it will work in real case:

https://postcss.org/api/#comment

```
Comment
The `Comment` represents
```

You will have unnecessary duplication of Comment on both lines.

lib/declaration.d.ts
vikaskaliramna07: sure, I didn't knew how API docs fetches it's data. I will remember that.

lib/comment.d.ts
vikaskaliramna07: okay, thanks for explaining. I will try not to repeat this mistake again

lib/comment.d.ts
```diff
@@ -30,18 +30,23 @@ declare namespace Comment {
...
  * Once (root, { Comment }) {
- *   let note = new Comment({ text: 'Note: …' })
+ *   const note = new Comment({ text: 'Note: …' })
  *   root.append(note)
  * }
  * ```
  *
- * Comments inside selectors, at-rule parameters, or declaration values
- * will be stored in the `raws` properties explained above.
+ * NOTE: CSS comments inside selectors, at-rule parameters, or
```
ai, line 47:
```suggestion
 * Note: CSS comments inside selectors, at-rule parameters, or
```

ai, line 47: Let’s avoid `NOTE:` prefix. Technically everything here is a note.

lib/declaration.d.ts
```diff
@@ -44,20 +44,24 @@ declare namespace Declaration {
 }
 
 /**
- * Represents a CSS declaration.
+ * It represents a class that handles CSS declarations.
+ *
+ * For more information about CSS declarations, please visit:
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/Syntax#css_declarations}
```
ai, line 50: Let’s put link on `CSS declarations` above

lib/declaration.d.ts
```diff
@@ -66,57 +70,72 @@ declare class Declaration_ extends Node {
...
    * ```
    */
   value: string
 
   /**
-   * `true` if the declaration has an `!important` annotation.
+   * The `important` property represents a boolean value. If true,
+   * the CSS declaration will have important specifier.
+   *
+   * For more information about CSS important specifier, please visit:
```
ai, line 107: Let’s avoid extra paragraph for links and use `important specifier` word to put a link

```
{@link [Foo.Bar](https://developer.mozilla.org/en-US/docs/Web/CSS/important) | important specifier}
```

https://typedoc.org/tags/link/

lib/declaration.d.ts
vikaskaliramna07: the line exceeds more than 80 characters :(
decreases readability.

Can I write it something like this

```md
For more information about CSS declarations,
click [here](https://developer.mozilla.org/en-US/docs/Web/CSS/Syntax#css_declarations)
```

ai: > Can I write it something like this

The problem is adding the whole paragraph just for a link.

> decreases readability

TSDoc is not to be read in a code. They are for https://postcss.org/api/ and VS Code TSDoc preview popup.

vikaskaliramna07: check out the updated version please

lib/declaration.d.ts
vikaskaliramna07: updated, please check

lib/comment.d.ts
```diff
@@ -30,18 +30,23 @@ declare namespace Comment {
   export { Comment_ as default }
 }
 
+
 /**
- * Represents a comment between declarations or statements (rule and at-rules).
+ * It represents a class that handles CSS comments.
+ *
+   * Learn more about CSS Specificity
+   * [here](https://developer.mozilla.org/en-US/docs/Web/CSS/Comments)
```
ai, line 38: Using `here` word is a bad practice for links. It is better to put it to the real word.

And we are again creating a whole paragraph just for a link.

Use `CSS comments` above instead.

vikaskaliramna07: sure

vikaskaliramna07:
I have pushed the changes, they appear on my fork but not appearing here... @ai

lib/comment.d.ts
```diff
@@ -30,18 +30,21 @@ declare namespace Comment {
   export { Comment_ as default }
 }
 
+
```
ai, line 33:
```suggestion
```

lib/declaration.d.ts
```diff
@@ -66,57 +68,68 @@ declare class Declaration_ extends Node {
...
+   * const decl = root.first?.first
+   *
+   * console.log(decl.value) //=> 'black'
    * ```
    */
   value: string
 
   /**
-   * `true` if the declaration has an `!important` annotation.
+   * The `important` property represents a boolean value. If true,
```
ai, line 102: Need to remove name duplication here as well

ai:
> I have pushed the changes, they appear on my fork but not appearing here... @ai

Strange.

What we can try to fix the problem:
1. Close this PR and open another from your branch (but GitHub will not suggest you a branch in yellow banner, you will need to click New PR and manually select branch). If not, I will get changes manually from your branch, don't worry (just don't delete branch in your repo).
2. Avoid do `merge main` commits. They are not very stable.
3. After your PR was accepted, I prefer to delete my fork (copy of PostCSS repo in your account) and create a fork again for the next PR.

idoros:
> Changed return type for Node.clone() from this to Node, reason: this is the reference to the original node, and having type of original node instance on a duplicated node instance is not correct method

@vikaskaliramna07 - can you expand on the reasoning behind this? When an instance of `postcss.Root` is cloned, the returned copy is also a Root node. This new return type require casting back into the original type. How can clone change the type of the returned instance?

PR #1881, opened 2023-09-16
improve sourcemap performance

romainmenke:
related change to PostCSS benchmarks : https://github.com/postcss/benchmark/pull/32

### DIFF
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -18,9 +18,9 @@ class MapGenerator {
     this.css = cssString
     this.usesFileUrls = !this.mapOpts.from && this.mapOpts.absolute
 
-    this.memoizedFileURLs = new Map();
-    this.memoizedPaths = new Map();
-    this.memoizedURLs = new Map();
+    this.memoizedFileURLs = new Map()
+    this.memoizedPaths = new Map()
+    this.memoizedURLs = new Map()
   }
 
   addAnnotation() {
@@ -258,7 +258,8 @@ class MapGenerator {
     }
 
     let path = relative(from, file)
-    this.memoizedPaths.set(file, path);
+    this.memoizedPaths.set(file, path)
+
     return path
   }
 
@@ -325,9 +326,9 @@ class MapGenerator {
   }
 
   toFileUrl(path) {
-    if (this.memoizedFileURLs.has(path)) {
-      return this.memoizedFileURLs.get(path);
-    }
+    let cached = this.memoizedFileURLs.get(path)
+    if (cached) return cached
+
     if (pathToFileURL) {
       let fileURL = pathToFileURL(path).toString()
       this.memoizedFileURLs.set(path, fileURL)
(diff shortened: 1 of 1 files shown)


lib/map-generator.js
```diff
@@ -241,18 +245,20 @@ class MapGenerator {
   }
 
   path(file) {
-    if (file.indexOf('<') === 0) return file
```
romainmenke, line 244: `indexOf` scans the entire string, this is slower than just checking the contents of the first code point.

lib/map-generator.js
```diff
@@ -17,6 +17,10 @@ class MapGenerator {
     this.opts = opts
     this.css = cssString
     this.usesFileUrls = !this.mapOpts.from && this.mapOpts.absolute
+
+    this.memoizedFileURLs = new Map();
+    this.memoizedPaths = new Map();
+    this.memoizedURLs = new Map();
```
romainmenke, line 23: Specific operations were done very frequently and were slow.
As far as I can tell these can be memoized.
This dramatically improves performance of sourcemap generation.

I don't have a strong opinion on naming or best practices here.
Please advice if there is a more idiomatic way to memoize these values within PostCSS.

ai:
Can you show the benchmark results (before/after) on you machine?

romainmenke:
On my machine (M2 air) : 

```
[22:24:27] Running suite Sourcemaps [/Users/romainmenke/projects/postcss-benchmark/sourcemaps.js]...
[22:24:33]    PostCSS x 8.41 ops/sec ±2.61% (45 runs sampled)
[22:24:39]    Next PostCSS x 17.42 ops/sec ±3.29% (85 runs sampled)
[22:24:39] Fastest test is Next PostCSS at 2.1x faster than PostCSS

Next PostCSS: 57 ms  (2.1 times faster)
PostCSS:      119 ms
```

lib/map-generator.js
```diff
@@ -241,18 +245,20 @@ class MapGenerator {
   }
 
   path(file) {
-    if (file.indexOf('<') === 0) return file
-    if (/^\w+:\/\//.test(file)) return file
     if (this.mapOpts.absolute) return file
+    if (file.charCodeAt(0) === 60 /* `<` */) return file
+    if (/^\w+:\/\//.test(file)) return file
+    if (this.memoizedPaths.has(file)) return this.memoizedPaths.get(file);
```
ai, line 251:
```suggestion
    let cached = this.memoizedPaths.get(file)
    if (cached) return cached
```

Could it be faster?

ai: Also, please remove `;`

lib/map-generator.js
```diff
@@ -241,18 +245,20 @@ class MapGenerator {
...
     let from = this.opts.to ? dirname(this.opts.to) : '.'
 
     if (typeof this.mapOpts.annotation === 'string') {
       from = dirname(resolve(from, this.mapOpts.annotation))
     }
 
-    file = relative(from, file)
-    return file
+    let path = relative(from, file)
+    this.memoizedPaths.set(file, path);
```
ai, line 260: `;`

lib/map-generator.js
romainmenke: I might be a little bit faster but not noticeable on my machine.
But I prefer this code style :)

ai:
Looks amazing. I will accept it tomorrow. Don’t want to accept and release before going to bed :D.

ai:
Thanks. Released in 8.4.30.

romainmenke:
Awesome! thank you 🙇

PR #1908, opened 2023-12-29
refactor: check all falsy conditions of plugins and opts before returning NoWorkResult in process()

### LINKED ISSUE #1869: `NoWorkResult` and `LazyResult` have very subtle differences and these can be accidentally encountered
https://github.com/postcss/postcss/blob/a0a82d3530ce7dfbfe1f12aa76c20f1b3f95eb7c/lib/processor.js#L44-L55

For a full repro see : https://github.com/romainmenke/postcss-no-work-result-sourcemaps

----

By setting `parser: null` as a processing option you can force a `LazyResult`.

Setting `null` instead of `undefined` is a common mistake, especially when not using TypeScript.

------

Is it intentional that `NoWorkResult` and `LazyResult` have different behaviors?

ahmdammarr:
resolves #1869

#### Fixes:
- refactor check of all falsy condition of plugins and opts before returing NoWorkResult in Processor.process().

> https://github.com/postcss/postcss/blob/a0a82d3530ce7dfbfe1f12aa76c20f1b3f95eb7c/lib/processor.js#L44-L55
> 
> For a full repro see : https://github.com/romainmenke/postcss-no-work-result-sourcemaps
> 
> By setting `parser: null` as a processing option you can force a `LazyResult`.
> 
> Setting `null` instead of `undefined` is a common mistake, especially when not using TypeScript.
> 
> Is it intentional that `NoWorkResult` and `LazyResult` have different CSS outputs?

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -43,10 +43,10 @@ class Processor {
 
   process(css, opts = {}) {
     if (
-      this.plugins.length === 0 &&
-      typeof opts.parser === 'undefined' &&
-      typeof opts.stringifier === 'undefined' &&
-      typeof opts.syntax === 'undefined'
+      !this.plugins.length &&
+      !opts.parser &&
+      !opts.stringifier &&
+      !opts.syntax
     ) {
       return new NoWorkResult(this, css, opts)
     } else {


ai:
It looks good to me. It is too late for release for me, I will try to release it tomorrow.

ai:
Thanks. Released in 8.4.33.

PR #1909, opened 2023-12-29
Ensure that `NoWorkResult` and `LazyResult` have the same outcome.

romainmenke:
see : https://github.com/postcss/postcss/issues/1869

### DIFF
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -16,6 +16,7 @@ class MapGenerator {
     this.root = root
     this.opts = opts
     this.css = cssString
+    this.originalCSS = cssString
     this.usesFileUrls = !this.mapOpts.from && this.mapOpts.absolute
 
     this.memoizedFileURLs = new Map()
@@ -74,7 +75,7 @@ class MapGenerator {
         }
       }
     } else if (this.css) {
-      this.css = this.css.replace(/(\n)?\/\*#[\S\s]*?\*\/$/gm, '')
+      this.css = this.css.replace(/\n*?\/\*#[\S\s]*?\*\/$/gm, '')
     }
   }
 
@@ -276,7 +277,7 @@ class MapGenerator {
           }
         })
       } else {
-        let input = new Input(this.css, this.opts)
+        let input = new Input(this.originalCSS, this.opts)
         if (input.map) this.previousMaps.push(input.map)
       }
     }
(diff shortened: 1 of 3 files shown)


lib/map-generator.js
```diff
@@ -16,6 +16,7 @@ class MapGenerator {
     this.root = root
     this.opts = opts
     this.css = cssString
+    this.originalCSS = cssString
```
romainmenke, line 19: There is a chicken vs. egg problem here that doesn't exist when fully parsing.

When the source is parsed there is already an `Input` instance created early. And this instance will have information about previous sourcemaps.

When the source is not parsed the `Input` instance is created in `previous()` but this is done **after** the annotation comments have already been removed.

Some mechanic is needed to get the previous sourcemaps from the original source and to store the modified css without the annotation comments.

I decided to add a second variable `this.originalCSS` which holds an unmodified copy.

I don't know if this is the best or even a good solution.
Please give feedback on this.

lib/map-generator.js
```diff
@@ -74,7 +75,7 @@ class MapGenerator {
         }
       }
     } else if (this.css) {
-      this.css = this.css.replace(/(\n)?\/\*#[\S\s]*?\*\/$/gm, '')
+      this.css = this.css.replace(/\n*?\/\*#[\S\s]*?\*\/$/gm, '')
```
romainmenke, line 78: When fully parsed all leading whitespace before an annotation comment will be removed, not only the first `\n`.

lib/no-work-result.js
```diff
@@ -37,6 +37,9 @@ class NoWorkResult {
       if (generatedMap) {
         this.result.map = generatedMap
       }
+    } else {
+      map.clearAnnotation();
+      this.result.css = map.css;
     }
```
romainmenke, line 43: If no sourcemap is requested the already present inline sourcemap must also be removed.

lib/map-generator.js
ahmdammarr: I don't understand how this part is related to the PR's main purpose (ensuring both `NoWorkResult` and `LazyResult` have the same outcome)

romainmenke: In a `LazyResult` the annotation is a parsed comment.
`Comment` nodes in PostCSS also contain all the leading whitespace.

In a `NoWorkResult` this RegExp is used and it only replaces a single preceding newline.

ahmdammarr (approved):

ai:
Thanks. Released in 8.4.33.

PR #1918, opened 2024-01-30
Fix code example for `name` in at-rule.d.ts

tim-we:
The order of lines in the example for AtRule#name got mixed up somehow. This PR corrects the order s.t. the example makes sense and would not produce an error.

https://postcss.org/api/#atrule-name


ai:
Thanks!

PR #1921, opened 2024-02-03
Improve AtRule#nodes type

### LINKED ISSUE #1920: AtRule#nodes can be undefined
Parsing something like
```css
@layer a, b, c;
```
with `postcss` will result in an `AtRule` object without a `nodes` property. The TS type and the docs suggest that its always present (type `postcss.ChildNode[]`). So my question is, should it be
1. always populated, in this case with an empty array
2. the types (and docs) are inaccurate?

In any case I'm happy to provide PR to fix this, just let me know what it should be.

https://postcss.org/api/#atrule-nodes

tim-we:
Parsing something like
```css
@layer a, b, c;
```
with `postcss` will result in an `AtRule` object without a `nodes` property. This however is not reflected by the types.

This PR changes the type of `AtRule#nodes` from `ChildNodes[]` to `ChildNodes[] | undefined` and updates the documentation.

I used `Container['nodes']` instead of `ChildNodes[]`, I can change that if you prefer it that way. I have also added a unit test to enforce this behavior (`AtRule#nodes` being `undefined` when the body is missing).

Closes #1920.

### DIFF
--- a/lib/at-rule.d.ts
+++ b/lib/at-rule.d.ts
@@ -87,6 +87,25 @@ declare class AtRule_ extends Container {
    */
   name: string
   /**
+   * An array containing the layer’s children.
+   *
+   * ```js
+   * const root = postcss.parse('@layer example { a { color: black } }')
+   * const layer = root.first
+   * layer.nodes.length           //=> 1
+   * layer.nodes[0].selector      //=> 'a'
+   * ```
+   *
+   * Can be `undefinded` if the at-rule has no body.
+   *
+   * ```js
+   * const root = postcss.parse('@layer a, b, c;')
+   * const layer = root.first
+   * layer.nodes //=> undefined
+   * ```
+   */
+  node: Container['nodes'] | undefined
+  /**
    * The at-rule’s parameters, the values that follow the at-rule’s name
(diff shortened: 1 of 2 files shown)


lib/at-rule.d.ts
```diff
@@ -86,6 +86,25 @@ declare class AtRule_ extends Container {
...
+   *
+   * Can be `undefinded` if the at-rule has no body.
+   *
+   * ```js
+   * const root = postcss.parse('@layer a, b, c;')
+   * const layer = root.first
+   * layer.nodes //=> undefined
+   * ```
+   */
+  node: Container['nodes'] | undefined
```
ai, line 107: Should it be `nodes:` instead of `node:`?

tim-we: Yes, sorry. I was wondering why that was working all of a sudden. Because with this change TS is not happy:
> Property 'nodes' in type 'AtRule_' is not assignable to the same property in base type 'Container_<ChildNode>'.

tim-we: I will open a new PR, I already have a branch ready :D

PR #1922, opened 2024-02-03
Improve AtRule#nodes type

### LINKED ISSUE #1920: AtRule#nodes can be undefined

tim-we:
Parsing something like
```css
@layer a, b, c;
```
with `postcss` will result in an `AtRule` object without a `nodes` property. This however is not reflected by the types.

This PR changes the type of `AtRule#nodes` from `ChildNodes[]` to `ChildNodes[] | undefined` and updates the documentation.

I had to change the types of `Container#nodes` (and undo that change for `Root#nodes`, `Rule#nodes` and `Document#nodes`), I did not find a better approach. Other things I have tried are  #1921 and changing this:
```js
declare class AtRule_ extends Container<ChildNode | undefined>
```
which sadly also does not work because `ChildNode | undefined` does not extend `Node`.

Closes #1920.

### DIFF
--- a/lib/at-rule.d.ts
+++ b/lib/at-rule.d.ts
@@ -95,9 +95,9 @@ declare class AtRule_ extends Container {
    * layer.nodes.length           //=> 1
    * layer.nodes[0].selector      //=> 'a'
    * ```
-   * 
+   *
    * Can be `undefinded` if the at-rule has no body.
-   * 
+   *
    * ```js
    * const root = postcss.parse('@layer a, b, c;')
    * const layer = root.first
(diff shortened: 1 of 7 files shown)


tim-we:
I'll have a look at those failing tests :weary:

tim-we:
Fixed tests.

PR #1932, opened 2024-03-20
Fix resulting `end` position when `endIndex` is `0` for `rangeBy`

romainmenke:
When `endIndex` is exactly `0` it is ignored and the `end` position will fallback to the end of the node.

Checking that it is a number instead of a simpler boolean check does seem to work as I would expect.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -241,7 +241,7 @@ class Node {
           column: opts.end.column,
           line: opts.end.line
         }
-      } else if (opts.endIndex) {
+      } else if (typeof opts.endIndex === 'number') {
         end = this.positionInside(opts.endIndex)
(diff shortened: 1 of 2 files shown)


ai:
Do we have a real use case? If yes, I can release it today (if not, I will wait another changes).

romainmenke:
Working on this for Stylelint: https://github.com/stylelint/stylelint/pull/7518
It isn't causing issues for end users. It would be nice to have for me, so that I can finish work there.

But please take your time, I always have plenty to do 😅

ai:
Thanks. Released in 8.4.38.

romainmenke:
Thank you 🙇

PR #1938, opened 2024-04-23
Fix useless warnings in console during source maps generation

pyatyispyatil:
It looks like [this fix](https://github.com/postcss/postcss/commit/1325896395a9a4693ba0d2d83e8ed99478ba6d36) didn't work completely. When using the postcss loader, quite a lot of warnings appear in the console, which do not really tell you anything. This appears to be due to the fact that ignoring an error in source-map-js directly calls `console.warn()`.

I think it's easier to just make an empty catch.

### DIFF
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -98,9 +98,10 @@ class MapGenerator {
     } else if (this.previous().length === 1) {
       let prev = this.previous()[0].consumer()
       prev.file = this.outputFile()
-      this.map = SourceMapGenerator.fromSourceMap(prev, {
-        ignoreInvalidMapping: true
-      })
+      try {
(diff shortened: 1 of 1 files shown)


xarety (approved):

dvun0g (approved):

ai:
Warnings by `console.warn()` don’t catch by `try {} catch {}`, so this PR will not fix the problem.

You have these warnings not because of PostCSS or `source-map-js`, but because your input CSS was broken. Hiding this warning is like hiding cough when you have pneumonia. You are fixing the symptom, not the source.

Send PR to the dependency with that CSS (with broken source map, which is loaded to PostCSS).

I will not hide this warning, since hiding the warning will make source map development without debugging tool.

PR #1940, opened 2024-04-25
Fix unreachable urls

Xvezda:
Updated the addresses in the document that were returning errors, inaccessible, or had changed to different locations


ai:
Thanks!

OKEAMAH (commented):
Fc

OKEAMAH (commented):
C

PR #1947, opened 2024-06-29
fix `CssSyntaxError` type declaration

romainmenke:
see : https://typescript-eslint.io/rules/only-throw-error/

When using TypeScript it is useful to only throw instances of `Error` (or subclasses thereof)

Currently we get an error for this lint rule with PostCSS on lines like these :

```ts
throw atRule.error('Rule must have valid params')
```

`CssSyntaxError` is a subclass of `Error`, but not declared as such in the `.d.ts` file.

### DIFF
--- a/lib/css-syntax-error.d.ts
+++ b/lib/css-syntax-error.d.ts
@@ -49,7 +49,7 @@ declare namespace CssSyntaxError {
  * }
  * ```
  */
-declare class CssSyntaxError_ {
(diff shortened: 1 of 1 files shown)


ai:
Thanks. Released at 8.4.39.

romainmenke:
Awesome, thank you 🙇

FreeMason9224:
Merge

PR #1950, opened 2024-07-24
Declare Node subclass attributes as getters

nex3:
This makes it easier to extend these classes and override these fields
with information that's derived from other, internal fields.

This helps enable https://github.com/sass/dart-sass/issues/88


ai:
Released in 8.4.40.

PR #1952, opened 2024-07-27
Make type declarations more accurate

nex3:
This also helps make it feasible to extend these types for
https://github.com/sass/dart-sass/issues/88.


lib/container.js
```diff
@@ -198,7 +198,7 @@ class Container extends Node {
         nodes.value = String(nodes.value)
       }
       nodes = [new Declaration(nodes)]
-    } else if (nodes.selector) {
+    } else if (nodes.selector || nodes.selectors) {
```
ai, line 201: It is not types. Why do we need this change?

nex3: It's just a bug I noticed while I was making these changes. The types here already allowed `{selectors: [".foo", ".bar"]}` as an argument for methods like `Container.prototype.append`, so this should handle it, right?

lib/at-rule.d.ts
```diff
@@ -130,9 +130,9 @@ declare class AtRule_ extends Container {
 
   constructor(defaults?: AtRule.AtRuleProps)
   assign(overrides: AtRule.AtRuleProps | object): this
-  clone(overrides?: Partial<AtRule.AtRuleProps>): AtRule
-  cloneAfter(overrides?: Partial<AtRule.AtRuleProps>): AtRule
-  cloneBefore(overrides?: Partial<AtRule.AtRuleProps>): AtRule
+  clone(overrides?: Partial<AtRule.AtRuleProps>): this
```
ai, line 133: Why it is `this` if we clone, create a new object?

lib/container.d.ts
```diff
@@ -24,7 +24,7 @@ declare namespace Container {
   }
 
   export interface ContainerProps extends NodeProps {
-    nodes?: (ChildNode | ChildProps)[]
+    nodes?: (Node | ChildProps)[]
```
ai, line 27: Why do we need this change?

lib/container.d.ts
```diff
@@ -161,8 +161,8 @@ declare abstract class Container_<Child extends Node = ChildNode> extends Node {
   insertAfter(
     oldNode: Child | number,
     newNode:
-      | Child
-      | Child[]
+      | Node
```
ai, line 164: It limits our types. We have `Child` for a reason and this change disable it.

lib/at-rule.d.ts
nex3: `this` doesn't mean _the current object_, it just means _the type of the current class_. This means that a subtype will automatically treat this as returning its own type, rather than returning the superclass `AtRule`. See [this documentation](https://www.typescriptlang.org/docs/handbook/2/classes.html#this-types).

lib/container.d.ts
nex3: It's only useful to limit your types to `Child` when dealing with referring to nodes that are _already_ in the parent. For nodes that are being _added to_ the parent, `Container.prototype.normalize` explicitly supports `Root`:

https://github.com/postcss/postcss/blob/165154b65a8e9af26f633b39e15d305321169413/lib/container.js#L187-L191

It's already totally valid for untyped JS code to pass in a `Root` here, this is just making the types match that.

nex3:
Merged main and added a declaration for `Input.toJSON()`.

lib/at-rule.d.ts
```diff
@@ -105,11 +105,11 @@ declare class AtRule_ extends Container {
   constructor(defaults?: AtRule.AtRuleProps)
   assign(overrides: AtRule.AtRuleProps | object): this
 
-  clone(overrides?: Partial<AtRule.AtRuleProps>): AtRule
+  clone(overrides?: Partial<AtRule.AtRuleProps>): this
```
ai, line 108: Does it have any practical change if `AtRule` has no type template?

ai:
Released in 8.4.41

PR #1953, opened 2024-07-29
Fix typo'd setter name in declaration.d.ts

querkmachine:
Hello, I just happened to notice this typo whilst reviewing the changes in the most recent release. (Thanks Dependabot!)

This was introduced in #1950.

I'm not familiar with PostCSS's internals or release process. As a release of PostCSS has already gone out with this setter name spelled incorrectly, I'm not sure if this warrants additional work to support the prior name and avoid a breaking change. 

Feel free to modify this PR as needed.

### DIFF
--- a/lib/declaration.d.ts
+++ b/lib/declaration.d.ts
@@ -138,7 +138,7 @@ declare class Declaration_ extends Node {
    * ```
    */
   get variable(): boolean
-  set varaible(value: string)
+  set variable(value: string)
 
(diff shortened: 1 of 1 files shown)


ai:
cc @nex3

ai:
In reality `Declaration#variable` has no setter, only getter https://github.com/postcss/postcss/commit/6127a44c27b1abbe532e52afcd0a243a472b1ed3

I think we could not release this fix since nobody can use broken property.

PR #1957, opened 2024-08-08
Add protected methods

nex3:
These are useful when writing a package like the Sass parser that inherits from PostCSS types.

### DIFF
--- a/lib/container.d.ts
+++ b/lib/container.d.ts
@@ -27,6 +27,19 @@ declare namespace Container {
     nodes?: (Node | ChildProps)[]
   }
 
+  /**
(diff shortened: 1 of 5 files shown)


ai:
Good additions

ai:
Can you use version from `github` during your refactoring to make one release later?

nex3:
> Can you use version from `github` during your refactoring to make one release later?

I'm not sure I understand what you mean by this.

ai:
Do you need npm release?

nex3:
Oh, I see. No, the changes that aren't yet released are things we can work around for now.

PR #1965, opened 2024-08-30
Pretty CssSyntaxError from onelined (minified) css

varpstar:
Add short snippet of code from onelined minified css file

### DIFF
--- a/lib/css-syntax-error.js
+++ b/lib/css-syntax-error.js
@@ -52,8 +52,14 @@ class CssSyntaxError extends Error {
 
     let css = this.source
     if (color == null) color = pico.isColorSupported
+
+    let highlightConditionally
     if (terminalHighlight) {
-      if (color) css = terminalHighlight(css)
+      if (color) {
+        highlightConditionally = text => terminalHighlight(text)
+      } else {
+        highlightConditionally = text => text
+      }
     }
 
     let lines = css.split(/\r?\n/)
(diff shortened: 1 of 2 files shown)


lib/css-syntax-error.js
```diff
@@ -71,6 +71,23 @@ class CssSyntaxError extends Error {
       mark = aside = str => str
     }
 
+    if (lines.length === 1 && this.source.length > 80) {
```
ai, line 74: Let's remove `lines.length` check since in many minified CSS could be comments with licence or `\n` at the end. 

But we will also change 80 to 160.

lib/css-syntax-error.js
```diff
@@ -71,6 +71,23 @@ class CssSyntaxError extends Error {
...
+      let gutter = '  1 | '
+      let spacing =
+        aside(' ' + gutter.replace(/\d/g, ' ')) +
+        subLine.slice(0, Math.min(this.column - 1, padding - 1)).replace(/[^\t]/g, ' ')
+
+      if (terminalHighlight) {
+        if (color) subLine = terminalHighlight(subLine)
+      }
+
+      return mark('>') + aside(gutter) + subLine + '\n' + spacing + mark('^')
```
ai, line 88: Will highlight works correctly if we break the line in the middle? Maybe break it by `{}`? It also could improve readability.

lib/css-syntax-error.js
```diff
@@ -71,6 +71,23 @@ class CssSyntaxError extends Error {
       mark = aside = str => str
     }
 
+    if (lines.length === 1 && this.source.length > 80) {
+      const padding = 20;
+      const subLineStart = Math.max(0, this.column - padding);
```
ai, line 76: We don't use `const`

ai:
LGTM. I will merge it this evening. Thanks.

PR #1968, opened 2024-09-06
Update remaining workflow jobs to use latest version of actions

Juneezee:
There are leftover actions from commit b952be76f0c7704a99dceb34294117d13e3470a4 that were not updated.

The workflow runs are annotated with the deprecation warnings. (Image from https://github.com/postcss/postcss/actions/runs/10699054789)

image


ai:
Thanks!

Issue #1979, opened 2024-11-05
`Node` methods that calculate source positions are inconsistent if there are AST mutations


romainmenke:
Context:

- `rangeBy` was introduced in https://github.com/postcss/postcss/pull/1669, specifically to aid Stylelint.
- `positionBy` was introduced in https://github.com/postcss/postcss/commit/31f3e86c65ebd00d98652990df1d93e5d0ef8389, specifically to aid Stylelint (related issue: https://github.com/postcss/postcss/issues/348)
- `positionAt` was introduced in https://github.com/postcss/postcss/commit/8d5d1b8d47040c76735a14dd1fb5ecad00beb1f6, specifically to aid Stylelint (related issue: https://github.com/postcss/postcss/issues/348)
- `positionAt` was renamed to the current `positionInside`

-------

Each method that calculates these positions (`positionInside`, `positionBy`, `rangeBy`, ...) starts by taking the `source` start position:

https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L191

This is the position of the current node as found in the original source, not the current position of serializing the entire AST.

However each function uses `this.toString()` to scan for new lines or to find the index of a given `word`: 

https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L190

This leads to inconsistencies and bugs when making mutations to the AST.

-------

A minimal reproduction:

```css
a {
	one: 1;
	two: 2;
}
```

- `a.positionInside(15)` -> `{ column: 3, line: 3 }`
- `two.positionInside(1)` -> `{ column: 3, line: 3 }`

If we remove `one: 1;` we alter the string representation of `a { ... }`, but not that of `two: 2;`.

If we then calculate the same positions we get different and inconsistent outcomes:

- `a.positionInside(15)` -> `{ column: 12, line: 2 }`
- `two.positionInside(1)` -> `{ column: 3, line: 3 }`

`a.positionInside(15)` reflects the position after mutation.
While `two.positionInside(1)` reflects the position in the original source.

I found that any positions calculated without making doing mutations are reliable and usable. But once you start changing or removing AST nodes these functions become unpredictable.

```ts
test('positionInside() returns original source position even after remove the previous node', () => {
  let css = parse('a {\n\tone: 1;\n\ttwo: 2;}')
  let a = css.first as Rule
  let one = a.first as Declaration
  let two = one.next() as Declaration

[...]

PR #1980, opened 2024-11-05
Fix inconsistent position calculations

romainmenke:
fixes: https://github.com/postcss/postcss/issues/1979

----

Benchmarks:

```
[16:51:26] Running suite Preprocessors [/Users/romainmenke/projects/postcss-benchmark/preprocessors.js]...

PostCSS sync:      29 ms  (1.0 times faster)
Next PostCSS sync: 29 ms  (1.0 times faster)
PostCSS:           30 ms
Next PostCSS:      31 ms  (1.0 times slower)

...

[16:52:19] Running suite Parsers [/Users/romainmenke/projects/postcss-benchmark/parsers.js]...

PostCSS:      13 ms
Next PostCSS: 13 ms (1.0 times slower)

...

[16:54:02] Running suite Linters [/Users/romainmenke/projects/postcss-benchmark/linters.js]...

Next PostCSS: 207 ms (1.1 times faster)
PostCSS:      219 ms

...

[16:54:14] Running suite Sourcemaps [/Users/romainmenke/projects/postcss-benchmark/sourcemaps.js]...

Next PostCSS: 67 ms (1.0 times faster)
PostCSS:      67 ms
```

Stylelint even seems to run a bit faster after this update.

I assume that calling `String.prototype.slice` is noticeably faster than recursively calling `Node.prototype.toString()` and serializing all nodes.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -177,22 +177,23 @@ class Node {
   positionBy(opts, stringRepresentation) {
     let pos = this.source.start
     if (opts.index) {
-      pos = this.positionInside(opts.index, stringRepresentation)
+      pos = this.positionInside(opts.index)
     } else if (opts.word) {
-      stringRepresentation = this.toString()
+      stringRepresentation = this.source.input.css.slice(this.source.start.offset, this.source.end.offset)
       let index = stringRepresentation.indexOf(opts.word)
-      if (index !== -1) pos = this.positionInside(index, stringRepresentation)
+      if (index !== -1) pos = this.positionInside(index)
     }
     return pos
   }
 
-  positionInside(index, stringRepresentation) {
-    let string = stringRepresentation || this.toString()
+  positionInside(index) {
     let column = this.source.start.column
     let line = this.source.start.line
+    let offset = this.source.start.offset
+    let end = offset + index
 
-    for (let i = 0; i < index; i++) {
-      if (string[i] === '\n') {
+    for (let i = offset; i < end; i++) {
+      if (this.source.input.css[i] === '\n') {
         column = 1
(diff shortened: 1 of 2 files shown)


lib/node.js
```diff
@@ -177,22 +177,23 @@ class Node {
...
       let index = stringRepresentation.indexOf(opts.word)
-      if (index !== -1) pos = this.positionInside(index, stringRepresentation)
+      if (index !== -1) pos = this.positionInside(index)
     }
     return pos
   }
 
-  positionInside(index, stringRepresentation) {
-    let string = stringRepresentation || this.toString()
+  positionInside(index) {
```
romainmenke, line 189: `stringRepresentation` argument wasn't part of the public API and it isn't needed anymore.

test/node.test.ts
```diff
@@ -412,11 +412,43 @@ test('positionInside() returns position when node contains newlines', () => {
...
+  equal(two.positionInside(1), { column: 3, line: 3 })
+})
+
 test('positionBy() returns position for word', () => {
   let css = parse('a {  one: X  }')
   let a = css.first as Rule
   let one = a.first as Declaration
   equal(one.positionBy({ word: 'one' }), { column: 6, line: 1 })
+  equal(one.positionBy({ word: 'X' }), { column: 11, line: 1 })
+  equal(a.positionBy({ word: '}' }), { column: 14, line: 1 })
```
romainmenke, line 436: I wanted some extra test coverage for characters at the end of nodes.

romainmenke:
Take your time ☺️, there is no pressure on our end.

Issue #1981, opened 2024-11-08
BUG - `require function is used in a way in which dependencies cannot be statically extracted`


rifont:
**What?**
Modern bundlers (Next.js build tooling) perform deep static code analysis to determine when `require` is used in a way that prevents static dependency extraction. This surfaced via:
`@novu/framework` -> `sanitize-html` -> `postcss` -> `picocolors`

`picocolors` have addressed the root cause in https://github.com/alexeyraspopov/picocolors/pull/87.

A simple dependency bump will resolve the issue.

rifont:
Thanks for getting onto this so quickly!

The latest version is picocolor@[1.1.1](https://www.npmjs.com/package/picocolors/v/1.1.1)

PR #1982, opened 2024-11-08
chore(deps): update picocolors to version `1.1.1`

rifont:
Closes the static dependency analysis bug in https://github.com/postcss/postcss/issues/1981


ai:
`^` means any version >=1.1 0 and <2.0.0.

So we don't need to change this line to use 1.1 1.

Just updated nested dependencies (there are many ways I use `pnpm update`).

PR #1994, opened 2024-12-19
Fix the documentation for `RuleRaws.ownSemicolon`

nex3:
This didn't match the actual type or behavior of the field.


ai:
Thanks!

Issue #1995, opened 2024-12-24
Syntaxes like `postcss-html` adjust node source locations relative to enclosing non-css documents


romainmenke:
See: https://github.com/ota-meshi/postcss-html/blob/168b49076ad66fe0e8766cb6ee899fa1534d3680/lib/html/parse-styles.js#L55-L76

For example:

```html
<p>Hello</p>

<style>
p {
  color: green;
}
</style>
```

The `p {}` rule when parsed with `postcss-html` has source locations that are offset relative to the entire html doc.

So `p {}` starts on line 4, not on line 1.

This is very useful when reporting errors (i.e. Stylelint) to developers in a code editors.
As the error will be reported on the line and column in the entire document.

---

However this unfortunately conflicts with the changes I introduced in https://github.com/postcss/postcss/pull/1980

Since those changes, we use `node.source.input.css` to infer positions.
But we assume the start of `node.source.input.css` corresponds with:
- `index == 0`
- `line == 1` and `column == 1`

Because `postcss-html` offsets the positions relative to the enclosing document this assumptions proves to be incorrect, making any position calculations incorrect.

-----

The simplest way forward I could think of was to add an extra field on `Input` to keep track of both the source of the enclosing document and of the CSS block independently.

https://github.com/postcss/postcss/blob/99da2f28aeded88c5e0fa3b4dd6fb44230710f57/lib/input.js#L17-L34

```js
class Input {
  constructor(css, opts = {}) {
    if (
      css === null ||
      typeof css === 'undefined' ||
      (typeof css === 'object' && !css.toString)
    ) {
      throw new Error(`PostCSS received ${css} instead of CSS string`)
    }

    this.css = css.toString()

    if (this.css[0] === '\uFEFF' || this.css[0] === '\uFFFE') {
      this.hasBOM = true
      this.css = this.css.slice(1)
    } else {
      this.hasBOM = false
    }

    this.document = this.css
    if (opts.document) this.document = opts.document.toString()

    // ...
```

For almost all usage of PostCSS `input.document` would correspond to `input.css` as the document is the CSS stylesheet.

But for `CSS-in-X` syntaxes (like `postcss-html`) the syntax authors could set `document` to the enclosing source (e.g. the html document).

----

When inferring positions we would use `node.source.input.document` instead of `node.source.input.css` in PostCSS itself.

```js
  positionBy(opts) {
    let pos = this.source.start
    if (opts.index) {
      pos = this.positionInside(opts.index)
[...]

PR #1996, opened 2024-12-25
Add `Input#document`

romainmenke:
fixes: https://github.com/postcss/postcss/issues/1995

### DIFF
--- a/lib/input.d.ts
+++ b/lib/input.d.ts
@@ -63,6 +63,16 @@ declare class Input_ {
   css: string
 
   /**
+   * Input source with support for non-CSS documents.
+   *
+   * ```js
+   * const input = postcss.parse('a{}', { from: file }).input
+   * input.document //=> "a{}"
+   * ```
+   */
+  document: string
+
+  /**
    * The absolute path to the CSS source file defined
    * with the `from` option.
    *
--- a/lib/input.js
+++ b/lib/input.js
@@ -33,6 +33,9 @@ class Input {
       this.hasBOM = false
     }
 
+    this.document = this.css
+    if (opts.document) this.document = opts.document.toString()
+
     if (opts.from) {
       if (
         !pathAvailable ||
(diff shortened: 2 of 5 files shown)


lib/input.js
```diff
@@ -33,6 +33,9 @@ class Input {
       this.hasBOM = false
     }
 
+    this.document = this.css
+    if (opts.document) this.document = opts.document.toString()
```
romainmenke, line 37: I chose to use `opts.document` and add `document` to `ProcessOptions`.

Alternatively it could also be less exposed and only exists as a field on `Input`.

```js
let root = parse('a {} b {}')
root.source.input.document = '<style>a {} b {}</style>';
```

We would still keep `this.document = this.css` to make sure that there is always a value.

lib/input.d.ts
```diff
@@ -62,6 +62,16 @@ declare class Input_ {
    */
   css: string
 
+  /**
+   * Input source with support for non-CSS documents.
+   *
+   * ```js
+   * const input = postcss.parse('a{}', { from: file }).input
```
ai, line 69: Let’s use HTML in example to show why we need `document`

lib/postcss.d.ts
```diff
@@ -315,6 +315,11 @@ declare namespace postcss {
   }
 
   export interface ProcessOptions<RootNode = Document | Root> {
+    /**
+     * The enclosing document for one or more CSS blocks.
```
ai, line 319:
```suggestion
     * Input file if it is not simple CSS file, but HTML with <style> or JS with CSS-in-JS blocks.
```

ai:
Looks amazing (only a few small docs changes suggested), but I prefer to merge and release it after holidays to avoid rush fixes in the middle of holiday :D

romainmenke:
Thank you for the review 🙇 
Yes, let's wait a bit before merging and releasing :)

Scutua (approved):

ai:
Thanks! The new property was released in 8.5.0.

romainmenke:
Thank you @ai for the reviews and release 🙇

PR #2020, opened 2025-02-23
Tạo ra sự an toàn.md

LD009:


ai:
We use Tidelift for that information.

But for non-Tidelify users our policy is simple:
1. We support the latest version
2. We have `Security` section in `README.md` with contacts to report an issue

We can re-open issue if you will provide extra details and arguments.

PR #2023, opened 2025-03-03
The return type of the method rawValue in Stringifier should be `any`…

Mallikarjun-0:
The return type of the method rawValue in Stringifier should be `any` instead of just a `string`
Here:
- `let rule = new AtRule({ name: 'page', nodes: [], params: 1 });`
   The type of return value of the method rawValue in Stringifier is a number (This is taken from one of the unit tests)
- `let rule = new AtRule({ name: 'page', nodes: [], params: true });`
   The type of return value of the method rawValue in Stringifier is a boolean

### DIFF
--- a/lib/stringifier.d.ts
+++ b/lib/stringifier.d.ts
@@ -35,7 +35,7 @@ declare class Stringifier_ {
   rawEmptyBody(root: Root): string | undefined
   rawIndent(root: Root): string | undefined
   rawSemicolon(root: Root): boolean | undefined
-  rawValue(node: AnyNode, prop: string): string
+  rawValue(node: AnyNode, prop: string): any
   root(node: Root): void
   rule(node: Rule): void
   stringify(node: AnyNode, semicolon?: boolean): void


lib/stringifier.d.ts
```diff
@@ -35,7 +35,7 @@ declare class Stringifier_ {
   rawEmptyBody(root: Root): string | undefined
   rawIndent(root: Root): string | undefined
   rawSemicolon(root: Root): boolean | undefined
-  rawValue(node: AnyNode, prop: string): string
+  rawValue(node: AnyNode, prop: string): any
```
ai, line 38:
```suggestion
  rawValue(node: AnyNode, prop: string): unknown
```

`any` is bad practice. If we don’t sure about type we must use `unknown`, so it will force developer to detect type.

ai:
`params: true` is not a valid value. `params` can be only `string` or `number`.

Do we have any other `props` with non-string type? If not, we should use `number | string` instead of `unknown`.

PR #2025, opened 2025-03-04
Change in type definition of property `css` of `result`

Mallikarjun-0:
In the `constructor` of class `Result` property `css` is initialized to `undefined` and the same is returned by the method `toString()`.

### DIFF
--- a/lib/result.d.ts
+++ b/lib/result.d.ts
@@ -67,7 +67,7 @@ declare class Result_<RootNode = Document | Root> {
    * postcss.parse('a{}').toResult().css //=> "a{}"
    * ```
    */
-  css: string
+  css?: string
 
   /**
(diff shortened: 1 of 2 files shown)


ai:
I think it will be better to fix it. Having `undefined` could break some types.

Do we have a result code by calling `postcss()` where `result.css` will be `undefined`?

Mallikarjun-0:
You're right, having `result.css` as `undefined` could break type safety. Currently, `result.css` is always updated to a string after creating a `Result` instance. It would be an issue if the instance of `Result` is created directly. 
I suggest to initialize `result.css` to an empty string.

Should I update my PR accordingly?

ai:
Good idea. Yes, please update

PR #2026, opened 2025-03-11
add postcss-auto-var-fallback fallback plugin to plugins.md

Ch-Valentine:
[`postcss-auto-var-fallback`] adds var fallbacks by provided files of variables


ai:
You forgot URL and `.`

Issue #2028, opened 2025-03-22
Calling `Node#rangeBy()`/`Node#positionBy()` with no arguments throws an error


mizdra:
## Summary

According to the type definition, the user can omit the `opts` argument of `Node#rangeBy()`.

https://github.com/postcss/postcss/blob/25443a59303678426e521fdb434bf95f4925ca14/lib/node.d.ts#L427

However, if it is omitted, `Node#rangeBy()`/`Node#positionBy()` will throw an error.

## Reproduction

https://stackblitz.com/edit/stackblitz-starters-moxcrard?file=index.ts&view=editor

`index.ts`:
```ts
import { parse, Rule } from 'postcss';

const css = parse('a {  one: X  }');
const a = css.first as Rule;
console.log(a.rangeBy()); // or `console.log(a.positionBy());`
```

When `a.rangeBy()` is called, the result is as follows:

```console
$ ts-node index.ts

/home/projects/stackblitz-starters-moxcrard/node_modules/postcss/lib/node.js:271
    if (opts.word) {
             ^
TypeError: Cannot read properties of undefined (reading 'word')
    at Rule.rangeBy (/home/projects/stackblitz-starters-moxcrard/node_modules/postcss/lib/node.js:271:14)
    at Object.eval (/home/projects/stackblitz-starters-moxcrard/index.ts:6:15)
    at Object._0x2b1965 (https://stackblitzstartersmoxcrard-tnaw-0o4eivqm.w-credentialless-staticblitz.com/blitz.5a421e5b.js:40:815191)
    at Module._compile (node:internal/modules/cjs/loader:144:14246)
    at m._compile (/home/projects/stackblitz-starters-moxcrard/node_modules/ts-node/dist/index.js:857:29)
    at Module._extensions..js (node:internal/modules/cjs/loader:144:14855)
    at require.extensions.<computed> (/home/projects/stackblitz-starters-moxcrard/node_modules/ts-node/dist/index.js:859:16)
    at Module.load (node:internal/modules/cjs/loader:144:12820)
    at Module._load (node:internal/modules/cjs/loader:144:10273)
    at Module.executeUserEntryPoint (node:internal/modules/run_main:165:1641)
```

When `a.positionBy()` is called, the result is as follows:

```console
$ ts-node index.ts

/home/projects/stackblitz-starters-moxcrard/node_modules/postcss/lib/node.js:213
    if (opts.index) {
             ^
TypeError: Cannot read properties of undefined (reading 'index')
[...]

Issue #2029, opened 2025-03-22
Missing `offset` property of `Node.Position` in some cases


mizdra:
## Summary

In https://github.com/postcss/postcss/commit/f09e2961b57b753a4b058d7ad97c2a9480326d65, `offset` property is added to `Node.Position`. But, the property is missing in some cases.

## Reproduction

- https://stackblitz.com/edit/stackblitz-starters-hvajrwqh?file=index.ts&view=editor

```ts
import { parse, Rule } from 'postcss';

const css = parse('a {  one: X  }');
const a = css.first as Rule;
const { start: a_invalid_1, end: a_invalid_2 } = a.rangeBy({});
const a_invalid_3 = a.positionBy({ index: 1 });
const a_invalid_4 = a.positionInside(0);
console.log(a_invalid_1);
console.log(a_invalid_2);
console.log(a_invalid_3);
console.log(a_invalid_4);

const a_valid_1 = a.source?.start;
const a_valid_2 = a.source?.end;
const a_valid_3 = a.positionBy({});
console.log(a_valid_1);
console.log(a_valid_2);
console.log(a_valid_3);
```

```console
$ npm start

> start
> ts-node index.ts

{ column: 1, line: 1 }
{ column: 15, line: 1 }
{ column: 2, line: 1 }
{ column: 1, line: 1 }
{ column: 1, line: 1, offset: 0 }
{ column: 14, line: 1, offset: 14 }
{ column: 1, line: 1, offset: 0 }
```

## Expected Behavior

As the `Node.Position` type definition indicates, the `offset` property is always included.

## Additional Information
- I discovered this behavior while working on https://github.com/postcss/postcss/issues/2022.
- If it is a bug, I can send a pull request to fix it.

PR #2031, opened 2025-03-22
Fix `opts` argument of `Node#rangeBy()` and `Node#positionBy()` cannot be omitted

mizdra:
fix: #2028

### DIFF
--- a/test/node.test.ts
+++ b/test/node.test.ts
@@ -453,6 +453,29 @@ test('positionInside() supports multi-root documents', () => {
   equal(a.positionInside(1), { column: 9, line: 1 })
 })
 
+test('positionBy() returns position', () => {
+  let css = parse('a {  one: X  }')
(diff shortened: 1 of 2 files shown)


test/node.test.ts
```diff
@@ -527,6 +550,59 @@ test('positionBy() supports multi-root documents', () => {
   equal(a.positionBy({ word: 'a' }), { column: 8, line: 1 })
 })
 
+test('rangeBy() returns range', () => {
+  let css = parse('a {  one: X  }')
+  let a = css.first as Rule
+  let one = a.first as Declaration
+  equal(one.rangeBy({}), {
+    end: { column: 12, line: 1 },
+    start: { column: 6, line: 1 }
```
mizdra, line 559: The `offset` is missing due to #2029.

ai:
Thanks!

Issue #2030, opened 2025-03-22
`node.source.end.offset` is wrong value in some cases


mizdra:
## Summary

In https://github.com/postcss/postcss/commit/f09e2961b57b753a4b058d7ad97c2a9480326d65, offset property is added to Node.Position. But, the property is wrong value in some cases. The `offset` is 1 larger than expected.

## Reproduction

- https://stackblitz.com/edit/stackblitz-starters-6ycx24v4?file=index.ts&view=editor

```ts
import { Declaration, parse, Rule } from 'postcss';

const css = parse('a { one: X }');
const a = css.first as Rule;
let one = a.first as Declaration;

console.log(css.source?.end); // ok
console.log(a.source?.end); // NG
console.log(one.source?.end); // NG
```

```console
$ npm start

> start
> ts-node index.ts

{ column: 13, line: 1, offset: 12 }
{ column: 12, line: 1, offset: 12 }
{ column: 10, line: 1, offset: 10 }
```

## Expected Behavior

The following will be output:
```console
$ npm start

> start
> ts-node index.ts

{ column: 13, line: 1, offset: 12 }
{ column: 12, line: 1, offset: 11 }
{ column: 10, line: 1, offset: 9 }
```

## Additional Information
- I discovered this behavior while working on https://github.com/postcss/postcss/issues/2022.
- I can send a pull request to fix it.

mizdra:
Off topic, but why don't `css.end` and `a.end` match? I think `css.end` should be `{ column: 12, line: 1, offset: 11 }`. That's the end of the file.

But the actual value is `{ column: 13, line: 1, offset: 12 }`. Is this a bug?

mizdra:
Uh, are the two issues (https://github.com/postcss/postcss/issues/2030#issue-2940170631 and https://github.com/postcss/postcss/issues/2030#issuecomment-2745135869) I noted both expected behavior?

According to #1879:

> The problem stemmed from direct use of the token location as the offset when resolving line and column with `getPosition()`, which worked well for `start.offset`, but should point to the index immediately after the node for the `end.offset` to ensure that `source.slice(start.offset, end.offset)` provides the correct node source.
> 
> The solution is straightforward: I've reviewed the parser's instances where the end offset is set and made a simple adjustment off adding 1 (except the final root location that works as expected).

ai:
cc @idoros

Also, I think we should call @romainmenke to not be sure that we are not breaking any of his Stylelint fixes for position.

[...]

PR #2032, opened 2025-03-22
Clarify documentation for `node.source.end.offset`

mizdra:
ref: #2030

The behavior of `node.source.end.offset` is confusing to users. I want to reduce this confusion. So I have tried to describe the behavior in the documentation.

### DIFF
--- a/lib/node.d.ts
+++ b/lib/node.d.ts
@@ -67,10 +67,21 @@ declare namespace Node {
      * The inclusive ending position for the source
      * code of a node.
      *
-     * However, only `end.offset` may be an exclusive position.
-     * The offset of `Root` node is the inclusive position,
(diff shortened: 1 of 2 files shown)


lib/node.d.ts
```diff
@@ -66,6 +66,11 @@ declare namespace Node {
     /**
      * The inclusive ending position for the source
      * code of a node.
+     *
+     * However, only `end.offset` may be an exclusive position.
```
ai, line 70: The wording is not very clear for everyone. I think it is better replace `exclusive` and `inclusive` words and show just an example.

mizdra: How about https://github.com/postcss/postcss/pull/2032/commits/c9b73ec046b547389ecfbfa862a140b022b2d7c6?

Issue #2022, opened 2025-03-01
Add `CssSyntaxError.offset` and `CssSyntaxError.endOffset`


mizdra:
## Problem

I want to display `CssSyntaxError` formatted to the user. For that I want to get the offset of the code where the error occurred. But `CssSyntaxError` does not have that property.

## Proposal

Add the `offset`/`endOffset` properties to `CssSyntaxError`.

erhab-jack-daw9292:
AI tried to solve the issue:

Added offset and endOffset properties to CssSyntaxError class to expose error position information. Updated TypeScript types and implementation to support the new properties.

You can review changes in this commit: https://github.com/erhab-jack-daw9292/postcss-postcss-2022/commit/dd68fbb2c65aa17acffbe817f913b34f2f6b5301.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

ai:
Hi. I like the idea.

Can I ask you to send PR?

mizdra:
Thank you. I will send PR.

mizdra:
I am currently working on a patch :)

PR #2034, opened 2025-03-22
Add `Input#fromLineAndColumn`, `CssSyntaxError.input.offset` and `CssSyntaxError.input.endOffset`

mizdra:
fix: #2022 

I initially tried to add `CssSyntaxError.offset` and `CssSyntaxError.endOffset`. But in the case of CSS with source map, it is always undefined. The sourcemap has the original line and column, but not the offset.

Instead, I added `CssSyntaxError.input.offset` and `CssSyntaxError.input.endOffset`. `CssSyntaxError.input` is a property with a location that ignores the sourcemap. I thought it appropriate to add `offset` and `endOffset` here.

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -9,7 +9,7 @@ let CssSyntaxError = require('./css-syntax-error')
 let PreviousMap = require('./previous-map')
 let terminalHighlight = require('./terminal-highlight')
 
-let fromOffsetCache = Symbol('fromOffsetCache')
+let lineToIndexCache = Symbol('lineToIndexCache')
 
 let sourceMapAvailable = Boolean(SourceMapConsumer && SourceMapGenerator)
 let pathAvailable = Boolean(resolve && isAbsolute)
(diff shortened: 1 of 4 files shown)


lib/input.js
```diff
@@ -67,32 +67,54 @@ class Input {
     if (this.map) this.map.file = this.from
   }
 
+  _getLineToIndex() {
```
ai, line 70: We don’t use `_`. Just don’t add it to TSDoc to make it as private.

ai: Or it is better to use functions instead of methods for private things (they are better in compression)

mizdra: I converted it to function: https://github.com/postcss/postcss/pull/2034/commits/d66f1d726b192bbceceb1c62b640356c4c291145

test/css-syntax-error.test.ts
```diff
@@ -285,14 +295,19 @@ test('uses source map', () => {
 
   is(error.file, join(__dirname, 'b.css'))
   is(error.line, 2)
+  is(error.column, 0) // Is this correct?
```
mizdra, line 298: It appears that the columns restored from the sourcemap are smaller by 1. This may be a bug.

I will investigate the details and create an issue.

ai: OK. Let’s merge it and think about it in separated issues/PRs.

ai:
Amazing. Thanks!

PR #2046, opened 2025-05-22
fixed error at line 401

git-sumitchaudhary:
when using postcss using parcel 
an error was occuring due to  no check mentioned for value at line 401 , which caused error while reading value when it was undefined hence i added a check if its value is undefined continue and dont move forward


git-sumitchaudhary:
minor error fixed at line 401 , when value is == null we should not  move forward and continue to avoid error, which i faced while using postcss with parcel (while using tailwind)

PR #2052, opened 2025-06-10
refactor: `import` should be listed before `require`

JounQin:
related https://github.com/un-ts/eslint-plugin-import-x/issues/334#issuecomment-2958489115

See also https://nodejs.org/api/packages.html#conditional-exports

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   },
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


ai:
Released in [8.5.5](https://github.com/postcss/postcss/releases/tag/8.5.5)

PR #2057, opened 2025-08-20
docs: add postcss-inline-extract plugin to plugins list

knowledgecode:
Add postcss-inline-extract plugin that extracts inline styles from HTML and converts them to CSS rules. Added to Language Extensions section in alphabetical order.


ai:
Thanks!

PR #2077, opened 2026-04-16
perf: optimize parser position tracking — 2ms faster parse, no pipeline regression

alanzabihi:
`Parser.getPosition()` calls `Input.fromOffset()` on every token, allocating a fresh `{col, line}` object each time. This patch inlines the binary search, caches the line-to-index array on the parser instance, and tracks the last resolved line so sequential tokens (same line or nearby) skip the search entirely.

| Single-file parse+stringify (44K lines) | 59 ms | 57 ms | -2 ms |
| 250-file plugin pipeline (4 plugins) | 273 ms | 273 ms | no change |

All tests pass. Output is byte-identical to the original.

Files changed: `lib/input.js`, `lib/parser.js`

---

This is 1 of 4 independent performance patches. The full set brings parse+stringify from 59 ms to 53 ms (10% faster) and the 250-file plugin pipeline from 273 ms to 228 ms (16% faster). They don't depend on each other and can be reviewed or merged in any order.

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -195,6 +195,10 @@ class Input {
     }
   }
 
+  getLineToIndex() {
+    return getLineToIndex(this)
+  }
+
   mapResolve(file) {
     if (/^\w+:\/\//.test(file)) {
       return file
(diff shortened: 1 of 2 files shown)


ai:
Sorry, the difference (3%) is too small for such increase of complexity

PR #2078, opened 2026-04-16
perf: inline getEvents() in walkSync — 35ms faster plugin pipeline

alanzabihi:
`walkSync()` calls `getEvents()` on every AST node, allocating a new array of event names each time. This patch inlines the dispatch logic, pre-computes whether any key-specific listeners exist (so it can skip `toLowerCase()` + string concat when there are none), and caches the proxy reference to avoid redundant `toProxy()` calls within a single node's event sequence.

| Single-file parse+stringify (44K lines) | 57 ms | 57 ms | no change |
| 250-file plugin pipeline (4 plugins) | 273 ms | 239 ms | -34 ms (13% faster) |

The speedup is in the plugin pipeline where the visitor dispatch hot path runs across 250 files with 4 plugins. Parse+stringify is unaffected since it doesn't use plugins.

All tests pass. Output is byte-identical.

Files changed: `lib/lazy-result.js`

---

This is 1 of 4 independent performance patches. Combined, they bring parse+stringify from 59 ms to 53 ms (10% faster) and the 250-file plugin pipeline from 273 ms to 228 ms (16% faster). No dependencies between them.

### DIFF
--- a/lib/lazy-result.js
+++ b/lib/lazy-result.js
@@ -281,6 +281,13 @@ class LazyResult {
       }
     }
     this.hasListener = Object.keys(this.listeners).length > 0
+    this.hasKeyListeners = false
+    for (let key in this.listeners) {
+      if (key.includes('-')) {
+        this.hasKeyListeners = true
+        break
+      }
+    }
   }
 
   async runAsync() {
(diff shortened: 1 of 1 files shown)


ai:
13% is very interesting.

1. What plugins was in second benchmark?
2. What are results on Preprocessors benchmark https://github.com/postcss/benchmark

lib/lazy-result.js
```diff
@@ -517,19 +524,51 @@ class LazyResult {
 
   walkSync(node) {
     node[isClean] = true
-    let events = getEvents(node)
```
ai, line 520: Should we delete `getEvents()` if we don't use it?

lib/lazy-result.js
```diff
@@ -517,19 +524,51 @@ class LazyResult {
 
   walkSync(node) {
```
ai, line 525: Should we also inline listeners login in `runAsync`?

alanzabihi:
The second benchmark processes 250 generated CSS files through 4 plugins:

1. `bench-bulk-transform` — Once handler, adds `-webkit-` prefixes and strips comments
2. `bench-decl-inspect` — Declaration visitor (read-only), counts declarations per file via `prepare()`
3. `bench-structure-inspect` — Rule + AtRule visitor (read-only), counts rules and selector length via `prepare()`
4. `bench-namespace` — Once handler, namespaces class selectors (`.foo` → `.ns-foo`)

Two use `prepare()` for per-file state, two use `Once` handlers — so both the visitor dispatch path and the single-pass walk path get exercised.

I ran the [preprocessors benchmark](https://github.com/postcss/benchmark) on the same server:

|  | Baseline | With patches |
|--|----------|-------------|
| PostCSS sync | 43 ms | 42 ms |
| PostCSS | 43 ms | 43 ms |

No meaningful difference there. That benchmark processes a single file with postcss-nested, postcss-simple-vars, and postcss-mixins. The walkSync improvement from this PR only really shows up across many files, the per-node dispatch overhead is small per file but adds up over hundreds of them.

lib/lazy-result.js
alanzabihi: Still used by `toStack()` on line 86 for the async path. Only the sync path inlines it here.

lib/lazy-result.js
alanzabihi: Yeah, could be worth doing. The async path uses a stack machine (`visitTick` + `toStack`), so an inline there would touch `visitTick`, `toStack`, and everything that reads stack frames. Bigger than the sync case here, but the same idea should apply.

ai:
Artificial plugins are not the best benchmark strategy. You should use the most popular plugins (Tailwind?) to see actual changes.

alanzabihi:
You were right to push back on synthetic plugins.

Tested against the tailwindcss.com site (pinned, 416 source files) through Tailwind v4 + postcss-preset-env + autoprefixer + cssnano. 10 interleaved A/B samples, plus a sequential pass to rule out CPU contention.

| Pipeline | Baseline | Optimized | Delta |
|----------|----------|-----------|-------|
| Full (Tailwind + 3) | 19,537 ms | 19,673 ms | +136 ms |
| Post-Tailwind only (3 plugins on Tailwind's 637KB output) | 19,593 ms | 19,629 ms | +36 ms |

Both within noise (stdev ~200-280 ms). Output byte-identical across all 20 optimized runs.

The 13% I measured earlier came from plugins that did almost no work per node, so dispatch overhead was most of the cost. Real plugins like cssnano do enough per-node work that the dispatch saving disappears.

Feel free to close. The change doesn't pay off in practice.

PR #2079, opened 2026-04-16
perf: skip MapGenerator when no source map — 9ms faster plugin pipeline

alanzabihi:
When `opts.map` is undefined and the root has no previous source map, `LazyResult.stringify()` still constructs a full `MapGenerator` — 3 `Map` allocations, a `clearAnnotation` AST walk, and a `previous()` AST walk — then discovers there's nothing to do. This patch checks first and stringifies directly when no map is needed.

| Single-file parse+stringify (44K lines) | 57 ms | 56 ms | -1 ms |
| 250-file plugin pipeline (4 plugins) | 239 ms | 230 ms | -9 ms (4% faster) |

The saving compounds across files. In the 250-file pipeline, each file was paying the `MapGenerator` construction cost for nothing.

All tests pass. Output unchanged.

Files changed: `lib/lazy-result.js`

---

This is 1 of 4 independent performance patches. The full set brings parse+stringify from 59 ms to 53 ms (10% faster) and the 250-file plugin pipeline from 273 ms to 228 ms (16% faster). Each stands on its own.

### DIFF
--- a/lib/lazy-result.js
+++ b/lib/lazy-result.js
@@ -378,6 +378,16 @@ class LazyResult {
     if (opts.stringifier) str = opts.stringifier
     if (str.stringify) str = str.stringify
 
+    let rootSource = this.result.root.source
+    if (opts.map === undefined && !(rootSource && rootSource.input && rootSource.input.map)) {
+      let result = ''
+      str(this.result.root, i => {
+        result += i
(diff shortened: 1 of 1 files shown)


ai:
Thanks. Nice idea.

PR #2080, opened 2026-04-16
perf: cache node.raws in Stringifier hot methods — 3ms faster parse+stringify

alanzabihi:
The Stringifier's `atrule()`, `block()`, `body()`, `comment()`, and `decl()` methods all access `node.raws` multiple times per call. This patch reads it once into a local variable. It also inlines the common `rawValue()` path in `decl()` when `node.raws.value` already exists, caches `root.rawCache` locally in `raw()`, and short-circuits `raw(child, 'before')` in `body()` when `child.raws.before` is already set.

| Single-file parse+stringify (44K lines) | 56 ms | 53 ms | -3 ms (5% faster) |
| 250-file plugin pipeline (4 plugins) | 230 ms | 228 ms | -2 ms |

All tests pass. Output is identical.

Files changed: `lib/stringifier.js`

---

This is 1 of 4 independent performance patches. Together they bring parse+stringify from 59 ms to 53 ms (10% faster) and the 250-file plugin pipeline from 273 ms to 228 ms (16% faster). No dependencies between them — they touch different files.

### DIFF
--- a/lib/stringifier.js
+++ b/lib/stringifier.js
@@ -37,11 +37,12 @@ class Stringifier {
   }
 
   atrule(node, semicolon) {
+    let raws = node.raws
     let name = '@' + node.name
     let params = node.params ? this.rawValue(node, 'params') : ''
 
-    if (typeof node.raws.afterName !== 'undefined') {
-      name += node.raws.afterName
+    if (typeof raws.afterName !== 'undefined') {
+      name += raws.afterName
     } else if (params) {
       name += ' '
     }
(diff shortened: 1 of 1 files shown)


ai:
5% is small for increasing the complexity, but we can try to merge it

43081j:
@ai this actually breaks custom syntaxes unfortunately.

specifically because it no longer calls `this.raw` thanks to changes like this one:

```ts
    let between = typeof raws.between !== 'undefined'
      ? raws.between
      : this.raw(node, 'between', 'beforeOpen')
```

this means a custom syntax that overrides `raw` doesn't get called anymore.

e.g. in postcss-lit, we override the stringifier to be able to correct indentation when stringifying raws. we can't do that anymore with this change, and not sure there's any non-trivial alternative

ai:
Let’s revert! Can you send PR reverting only related things and I will release it quickly.

43081j:
No worries I can get a pr together later today 👍

The rest of this PR looks good so it's only the short circuit stuff I think.

PR #2084, opened 2026-05-04
fix: always call `raw` to retrieve raw values

43081j:
Without this, it becomes impossible to override the stringifier with
your own raw computation, etc.


ai:
Thanks. Fixed in 8.5.14.

PR #2094, opened 2026-06-10
Fix explicit comment spacing on root insertion

### LINKED ISSUE #2017: `.before()` modifies `raws.before` after insertion.
Given a loud comment with new lines:
```scss
@mixin x {
    margin: 10px;
}

/*
* loud
* comment
*
*/

.a {
    @include x;
}
```
If I wanted to insert a new comment, e.g.: 

```js
const node = postcss.comment({
    text: 'New Comment',
    raws: {'inline': true, 'right': '', 'left': ' '},
});
console.log(node);
```
The log has the correct raws:
```json
{"raws":{"inline":true,"right":"","left":" ","before":""},"text":"New Comment","type":"comment","inputs":[]}
```

But after insertion, the `raws.before` property is cloned from the original comment:
```js
comment.before(node);
const previousComment = comment.prev() as postcss.Comment;
console.log(previousComment)
```

So the result is the following:
```js
{
"raws": {
    "inline": true,
    "right": "",
    "left": " ",
    "before": "\n\n" // Should be ""
},
"text": "New comment",
"type": "comment",
"inputs": []
}
```

sjh9714:
## Summary

Fixes #2017.

Root-level insertion was overwriting a newly supplied node's explicit whitespace-only `raws.before` with spacing from the following sample node. This preserves intentionally supplied spacing for new nodes while leaving parsed, moved, root, string, and non-whitespace raw values normalized as before.

## Motivation

The linked issue does not name a public plugin. It shows direct PostCSS API usage that a plugin or transform can use: create a new comment node, set explicit raw spacing, and insert it before an existing root-level loud comment.

Input SCSS:

```scss
@mixin x {
    margin: 10px;
}

/*
* loud
* comment
*
*/

.a {
    @include x;
}
```

Insertion code from the issue:

```js
const node = postcss.comment({
  text: 'New Comment',
  raws: { inline: true, right: '', left: ' ', before: '' },
})

comment.before(node)
```

Before insertion, the created node has the intended `raws.before: ''`. After `comment.before(node)`, the current behavior overwrites that explicit value with the existing root comment's spacing:

```js
{
  raws: {
    inline: true,
    right: '',
    left: ' ',
    before: '\n\n', // expected ''
  },
  text: 'New Comment',
  type: 'comment',
}
```

That means the inserted comment inherits the blank-line spacing from the following loud comment. In CSS terms, the current output shape adds the old root-comment spacing before the new generated comment:

```scss
@mixin x {
    margin: 10px;
}

// New Comment

/*
* loud
* comment
*
*/
```

The expected output shape keeps the caller's explicit `before: ''`, so the inserted comment is placed without inheriting the two-newline prefix from the existing comment:

```scss
@mixin x {
    margin: 10px;
}

// New Comment
/*
* loud
* comment
*
*/
```

Setting `raws.before` in user code is not enough today because the insertion normalization overwrites it after the node is inserted.

The ecosystem safety comes from the scoped condition in this change: only newly supplied nodes without a parent and with whitespace-only explicit `raws.before` keep that value. Existing or moved nodes still get normalized from the target sample, and non-whitespace `raws.before` values such as embedded `</style>` content still go through the old normalization path so stringifier escaping behavior is preserved.

## Changes

- Preserve whitespace-only explicit `raws.before` for newly supplied root nodes during root normalization.
- Keep existing normalization for moved nodes, strings, roots, and non-whitespace `raws.before` values.
- Add regression coverage for new comments/rules preserving explicit spacing and moved comments updating to target spacing.

## Testing

- `pnpm exec uvu -r ts-node/register/transpile-only test "stringifier\\.test\\.js$"`
- `pnpm exec uvu -r ts-node/register/transpile-only test "container\\.test\\.ts$"`
- `pnpm exec uvu -r ts-node/register/transpile-only test "(container|root|stringifier|visitor)\\.test\\.(ts|js)$"`
- `pnpm test:lint && pnpm test:types`
- `git diff --check`
- `pnpm unit` (all non-color-related tests passed; two existing `css-syntax-error` color-output assertions fail in this environment because `FORCE_COLOR` is set while `NO_COLOR=1`.)

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/lib/root.js
+++ b/lib/root.js
@@ -13,20 +13,20 @@ function hasExplicitBefore(node) {
   )
 }
 
-function isNewCommentWithExplicitBefore(node) {
+function isNewNodeWithExplicitBefore(node) {
   return (
     node &&
     typeof node === 'object' &&
     !node.parent &&
-    (node.type === 'comment' || node.text) &&
     hasExplicitBefore(node)
   )
 }
 
 function explicitBeforeValues(child) {
   if (!child || typeof child === 'string') return []
-  if (Array.isArray(child) || child.type === 'root') return []
-  return [isNewCommentWithExplicitBefore(child)]
+  if (Array.isArray(child)) return child.map(isNewNodeWithExplicitBefore)
+  if (child.type === 'root') return []
+  return [isNewNodeWithExplicitBefore(child)]
 }
 
 class Root extends Container {
--- a/lib/root.js
+++ b/lib/root.js
@@ -4,12 +4,13 @@ let Container = require('./container')
 
 let LazyResult, Processor
 
-function hasExplicitBefore(node) {
+function hasExplicitWhitespaceBefore(node) {
   return (
     node &&
     typeof node === 'object' &&
     node.raws &&
-    typeof node.raws.before !== 'undefined'
+    typeof node.raws.before !== 'undefined' &&
+    /^\s*$/.test(node.raws.before)
   )
 }
 
@@ -18,15 +19,19 @@ function isNewNodeWithExplicitBefore(node) {
     node &&
     typeof node === 'object' &&
     !node.parent &&
-    hasExplicitBefore(node)
+    hasExplicitWhitespaceBefore(node)
   )
 }
 
-function explicitBeforeValues(child) {
-  if (!child || typeof child === 'string') return []
-  if (Array.isArray(child)) return child.map(isNewNodeWithExplicitBefore)
-  if (child.type === 'root') return []
-  return [isNewNodeWithExplicitBefore(child)]
+function collectNewNodesWithExplicitBefore(child, values = new Set()) {
+  if (!child || typeof child === 'string') return values
+  if (Array.isArray(child)) {
+    for (let node of child) collectNewNodesWithExplicitBefore(node, values)
+    return values
+  }
+  if (child.type === 'root') return values
+  if (isNewNodeWithExplicitBefore(child)) values.add(child)
+  return values
 }
 
 class Root extends Container {
@@ -37,7 +42,7 @@ class Root extends Container {
   }
 
   normalize(child, sample, type) {
-    let explicitBefore = explicitBeforeValues(child)
+    let explicitBefore = collectNewNodesWithExplicitBefore(child)
     let nodes = super.normalize(child)
 
     if (sample) {
@@ -48,8 +53,8 @@ class Root extends Container {
           delete sample.raws.before
         }
       } else if (this.first !== sample) {
-        for (let [index, node] of nodes.entries()) {
-          if (explicitBefore[index]) continue
+        for (let node of nodes) {
+          if (explicitBefore.has(node)) continue
           node.raws.before = sample.raws.before
         }
       }
--- a/lib/root.js
+++ b/lib/root.js
@@ -13,10 +13,20 @@ function hasExplicitBefore(node) {
   )
 }
 
+function isNewCommentWithExplicitBefore(node) {
+  return (
+    node &&
+    typeof node === 'object' &&
+    !node.parent &&
+    (node.type === 'comment' || node.text) &&
+    hasExplicitBefore(node)
+  )
+}
+
 function explicitBeforeValues(child) {
   if (!child || typeof child === 'string') return []
   if (Array.isArray(child) || child.type === 'root') return []
-  return [(child.type === 'comment' || child.text) && hasExplicitBefore(child)]
+  return [isNewCommentWithExplicitBefore(child)]
 }
 
 class Root extends Container {
--- a/lib/root.js
+++ b/lib/root.js
@@ -3,6 +3,7 @@
 let Container = require('./container')
 
 let LazyResult, Processor
+let preserveBefore = Symbol('preserveBefore')
 
 function hasExplicitWhitespaceBefore(node) {
   return (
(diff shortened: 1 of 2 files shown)


ai:
I need motivation section for this type of PR.

What is your user case? Why do you need this change? Why it will not break ecosystem?

sjh9714:
Added a Motivation section to the PR description.

The user case is plugins that construct comment nodes and intentionally set `raws.before` to control exact spacing when inserting at the root. The change is scoped to inserted comment nodes with an explicit `raws.before`, so existing normalization remains unchanged for parsed nodes, string children, root children, and nodes without explicit spacing.

ai:
Sorry, you have not explained any extra details in motivation section, just told that you already have.

What plugin? Why you can’t set `raws.before`? What is your code? Show some real input, current output, expected output CSS.

> This should not break the ecosystem because the change is limited to inserted comment nodes that explicitly provide raws.before. Parsed nodes, string children, root children, and nodes without an explicit raws.before continue to use the existing normalization behavior.

It sounds like excuse, not a real research.

Why it will not break expected behavior for other people?

sjh9714:
I updated the Motivation section with the concrete reproduction from #2017.

There is no named public plugin in the linked issue; it shows direct PostCSS API usage that plugin code can use. The important part is that the caller already sets `raws.before: ""` on the new `postcss.comment(...)` node before insertion, but `comment.before(node)` overwrites that value with the following root comment's `"\n\n"` spacing. So the workaround cannot simply be “set `raws.before`” because that is exactly what gets replaced during insertion normalization.

I also added the input SCSS, the insertion code, and current vs expected output shape. The safety scope is narrow: only newly inserted `comment` nodes with their own explicit `raws.before` keep that value. Parsed nodes, string children, root insertions that are not comments, comments without `raws.before`, and other normalized nodes continue through the existing behavior.

ai:
We have explicit code to clean `raws.before` for a reason. Can you check when it was added to see the reason in the commit?

sjh9714:
I checked the history around this root spacing code.

The closest commits I found are:

- `cffb99ba` (`Do not copy before from first element`, 2014-12-27): added the regression that `insertAfter()` must not copy spacing from the first root child. In `a{} b{}`, the node inserted after the first child keeps `raws.before` undefined, while the node inserted after the later child gets the sample spacing.
- `981173d7` (`Clean up Root before hacks`, 2014-12-27): cleaned up the root spacing workaround so root-level insertions copy spacing from the sample node, and `prepend` adjusts/removes the first node spacing.
- `5744739b` (`Fix keeping before on moving roots`, 2017-06-12): preserved root spacing when moving nodes between roots by avoiding the normal first-child spacing transfer during that internal remove.

So I read the reason as: keep root-level formatting consistent when adding/moving nodes, and especially avoid copying the first root child’s leading spacing into inserted nodes.

This PR should not remove that behavior. It only skips the sample-spacing overwrite for a newly inserted `comment` node that already has its own explicit `raws.before`. Nodes without explicit `raws.before` still go through the existing normalization path.

I re-ran the focused root/container tests to check the old behavior is still covered:

- `pnpm exec uvu -r ts-node/register/transpile-only test "(container|root)\\.test\\.(ts|js)$"`

That passed 100/100, including `insertAfter() does not use before of first rule`, root moving, prepend, and the new explicit-comment-spacing regression.

ai:
> Fix keeping before on moving roots

But when we move node from `a{}` into `a{b{}}` the moved node should have `raws.before`? And it will not be updated?

sjh9714:
I reproduced this concern. A moved comment from another root already has `raws.before`, and the previous version of this PR kept that old spacing instead of applying the target root sample spacing.

I narrowed the exception to comments that do not have a parent yet, so it only preserves explicit spacing for newly created comment nodes. Existing/moved comments now keep the old root normalization behavior and get their spacing updated from the target sample.

Added a regression for the moved-comment case and re-ran:

- `pnpm exec uvu -r ts-node/register/transpile-only test "container\\.test\\.ts$"`
- `pnpm exec uvu -r ts-node/register/transpile-only test "(container|root)\\.test\\.(ts|js)$"`
- `pnpm test:lint && pnpm test:types`
- `git diff --check`

ai:
I don't like special logic only for comments. It makes behavior more complex and less predictable.

I ask you to not do a quick fix (just for your case) and stop yourself, think more about the whole PostCSS ecosystem, think about other users and other use case and came with a plan.

sjh9714:
I reworked this to remove the comment-only special case.

The rule is now type-agnostic:

- newly supplied nodes without a parent keep an explicit `raws.before`
- moved/existing nodes keep the existing root normalization behavior and can have `raws.before` updated from the target sample
- strings and roots stay out of this path

I updated the regression to cover both a newly created comment and a newly created rule with explicit `raws.before`, and kept the moved-comment regression to prevent the issue you pointed out.

Re-ran:

- `pnpm exec uvu -r ts-node/register/transpile-only test "container\\.test\\.ts$"`
- `pnpm exec uvu -r ts-node/register/transpile-only test "(container|root)\\.test\\.(ts|js)$"`
- `pnpm test:lint && pnpm test:types`
- `git diff --check`

lib/root.js
```diff
@@ -12,6 +37,7 @@ class Root extends Container {
   }
 
   normalize(child, sample, type) {
+    let explicitBefore = explicitBeforeValues(child)
```
ai: Why do we return an array here and not calling a helper in `for (let [index, node] of nodes.entries())`?

sjh9714: Reworked this after checking the CI failure.

I removed the index-aligned array and now collect the original newly supplied nodes in a `Set` before `super.normalize()` mutates their `parent`. The loop then checks `explicitBefore.has(node)` directly.

I also narrowed the preservation rule to whitespace-only `raws.before`. That keeps the `before: ''` use case from #2017, but lets non-whitespace raw values keep the existing normalization path. This fixed the failing `stringifier.test.js` `</style>` escaping case from CI.

Re-ran:

- `pnpm exec uvu -r ts-node/register/transpile-only test "stringifier\\.test\\.js$"`
- `pnpm exec uvu -r ts-node/register/transpile-only test "container\\.test\\.ts$"`
- `pnpm exec uvu -r ts-node/register/transpile-only test "(container|root|stringifier|visitor)\\.test\\.(ts|js)$"`
- `pnpm test:lint && pnpm test:types`
- `git diff --check`

ai: It is not an answer. Why you run it before the task and not on per-node basis?

The current code is complicated and use more resources because you return an array.

---

Do not add `Re-ran` section please.

sjh9714: Changed this to a per-node check in the loop.

`super.normalize()` mutates `parent`, so the only pre-normalize work left is a temporary marker on nodes that were parentless and had whitespace-only `raws.before`. The loop now calls `shouldPreserveExplicitBefore(node)` for each normalized node and removes that marker immediately. There is no index-aligned array or `Set` anymore.

ai: The current code is even more complicated.

sjh9714:
Closing this PR. The current approach adds too much complexity to root normalization, and this behavior needs a simpler design before changing it. I’ll step back rather than keep iterating on this branch.

PR #2098, opened 2026-06-25
Add postcss-nth-nested to plugins list

georgeadamson:
Adds postcss-nth-nested to the Language Extensions section of the PostCSS plugins list.

Hiiii! I believe this is the process to add a plugin to the list! Thank you for PostCSS BTW, it's pretty amazing


ai:
Thanks :) Nice plugin

PR #2033, opened 2025-03-22
Fix missing offset in `Node#positionInside`, `Node#positionBy` and `Node#rangeBy`

mizdra:
fix: #2029 
base on: #2031

PR #2099, opened 2026-06-26
fix(node): always compute offset in positionBy() like rangeBy()

mahirhir:
Mirrors #2033 to the one sibling it left out.

#2033 (issue #2029) normalized `rangeBy()` and `positionInside()` to always compute `offset` via `sourceOffset()`, so they return a complete `Position`. `positionBy()`'s default branch was left returning `this.source.start` as-is:

`lib/node.js` (default branch of `positionBy`)
```js
position = this.source.start
```

For custom syntaxes whose `source.start` carries no `offset` (e.g. postcss-html), this returns a position missing `offset`, which violates the `Position` type that declares `offset: number` as required. `rangeBy()` two lines away already handles exactly this via `sourceOffset()`; `positionBy()` is the unmirrored twin.

Fix mirrors `rangeBy()` precisely (compute `offset` through `sourceOffset()` on the default branch).

Verification: added a regression test that fails on baseline (1/90 fail) and passes with the fix (90/90); the full existing suite is 89/89 on both baseline and patched, so zero regression; `tsc --noEmit` is clean. Branch is based on the current `main` tip.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -206,14 +206,18 @@ class Node {
   }
 
   positionBy(opts = {}) {
-    let pos = this.source.start
+    let inputString =
+      'document' in this.source.input
+        ? this.source.input.document
+        : this.source.input.css
+    let pos = {
+      column: this.source.start.column,
+      line: this.source.start.line,
+      offset: sourceOffset(inputString, this.source.start)
+    }
     if (opts.index) {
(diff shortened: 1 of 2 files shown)


ai:
Thanks

ai:
Released in 8.5.16.

PR #2100, opened 2026-06-28
Preserve node raws when rehydrating a JSON AST

sarathfrancis90:
`fromJSON(root.toJSON())` isn't lossless when a stylesheet has uneven spacing between its top-level nodes:

```js
let css = 'a {}\nb {}\n\nc {}\n'
let root = postcss.parse(css)
postcss.fromJSON(JSON.parse(JSON.stringify(root.toJSON()))).toString()
// => 'a {}\nb {}\nc {}\n'   the blank line before `c` is gone
```

`toJSON()` serializes every `raws.before` correctly — the loss happens on the way back. `fromJSON` builds the child nodes and then passes them to the container constructor, which appends them; for root children that re-runs the insertion spacing logic in `Root.normalize`, overwriting each node's own `raws.before` with the previous sibling's. So every top-level node after the first inherits the first one's spacing.

It matters because round-tripping through JSON is meant to restore an AST exactly (e.g. caching a parsed AST), so a rehydrated root should stringify to the same CSS as the original.

I rehydrate the children and attach them directly instead of routing them through `append`, leaving their raws untouched. Runtime `append`/insertion behavior is unchanged. Found it round-tripping the postcss-parser-tests fixtures; added a regression test and `pnpm test` is green.

### DIFF
--- a/lib/fromJSON.js
+++ b/lib/fromJSON.js
@@ -25,8 +25,13 @@ function fromJSON(json, inputs) {
       inputs.push(inputHydrated)
     }
   }
+  // Rehydrate children separately and attach them after construction.
+  // Passing them through the container constructor would re-run insertion
+  // spacing normalization and overwrite each child's own `raws.before`.
+  let nodes
   if (defaults.nodes) {
-    defaults.nodes = json.nodes.map(n => fromJSON(n, inputs))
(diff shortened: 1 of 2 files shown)


ai:
Thanks!

ai:
Released in 8.5.16.

PR #2036, opened 2025-03-24
Fix `Input#origin()` returning incorrect position

mizdra:
fix: #2035

PR #2106, opened 2026-06-30
Fix `Input#origin()` mixing null and undefined for unmapped end position

chatman-media:
When the end position passed to `origin()` isn't covered by the source map, `originalPositionFor()` comes back with `line`/`column` set to `null` instead of just not having a mapping. `endLine` ends up `null` in that case, but `endColumn` was being built as `to && to.column + 1`, which because of precedence is actually `to && (to.column + 1)`, so `null` silently becomes `1`. You can end up with something like `{ endColumn: 1, endLine: null }`, which is inconsistent with how the rest of the function treats an unresolved position (both stay `undefined`).

Looks like #2036 fixed the off-by-one for the case where the end position does resolve, but didn't account for it not resolving at all.

Guarded `to` the same way `from` already is a few lines above, and added a test that reproduces it with a source map that doesn't cover the requested end line.

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -211,10 +211,16 @@ class Input {
 
     let to
     if (typeof endLine === 'number') {
-      to = consumer.originalPositionFor({
+      let toPosition = consumer.originalPositionFor({
         column: endColumn - 1,
         line: endLine
       })
+      // The source map may not have a mapping that covers the end position
+      // (`originalPositionFor()` then returns `null` for `line`/`column`
+      // instead of omitting them). Treat that the same as not requesting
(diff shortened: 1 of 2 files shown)


ai:
Thanks!

Do you think we should release it right now or the user case is pretty rare (so we will wait more features to release together)?

chatman-media:
Up to you — it's not a long-standing bug, more a fresh regression from #2036/8.5.16, so I'd lean toward not sitting on it too long rather than batching it. Whenever a reported range's end falls outside sourcemap coverage, anything reading `endColumn` as a number gets a bogus 1 with no matching `endLine` — not sure how rare that is in the wild, but it's not a contrived case.

PR #2110, opened 2026-07-08
Fix non-closed `<div align="center">` in README

ybiquitous:
Fixes the problem that all elements after `<div align="center">` are centered in `README.md`.

Ref: <https://github.com/postcss/postcss/blob/1bf9076867e70f5a01776aebebee5aacbf9771ce/README.md>

Screenshot:

image


ai:
Thanks!

PR #2111, opened 2026-07-13
Keep explicitly set raws.before when inserting nodes into root

### LINKED ISSUE #2017: `.before()` modifies `raws.before` after insertion.

MahinAnowar:
Fixes #2017

`Root.normalize` unconditionally copies `sample.raws.before` onto every inserted node, so a node created with an explicit `raws.before` (like the loud comment in the issue, built with `postcss.comment({ text, raws: { before: '' } })`) loses it on insertion. `Container.normalize` already treats `raws.before` as user-owned and only fills it when missing — this brings the root-level path in line with that, which is also what you suggested in the issue thread.

The rule I went with is type-agnostic: if an incoming node is detached (no parent) and has `raws.before` set, that value is kept. Nodes that are attached somewhere else at insertion time are still renormalized to the target root's spacing, so moving nodes between roots behaves exactly as before — there's a guard test for that, and the existing "keeps spaces on moving root" test still passes.

One existing test needed its expected string updated: the `</style>` escape test sets `rule.raws.before = '\n</style>'` on a detached rule, but the old overwrite silently discarded that value before the stringifier ever saw it, so the escape of `before` was never actually exercised. With the fix the explicit `before` survives and renders escaped (`\3c /style>`), which as far as I can tell is what that test meant to check all along, since it sets both `before` and `after`.

Full suite passes (670/670), plus eslint and check-dts.

### DIFF
--- a/lib/root.js
+++ b/lib/root.js
@@ -12,6 +12,19 @@ class Root extends Container {
   }
 
   normalize(child, sample, type) {
+    let keepBefore = new Set()
+    for (let node of Array.isArray(child) ? child : [child]) {
+      if (
+        node &&
+        typeof node === 'object' &&
+        !node.parent &&
+        node.raws &&
+        typeof node.raws.before !== 'undefined'
+      ) {
+        keepBefore.add(node.raws)
+      }
+    }
+
     let nodes = super.normalize(child)
 
     if (sample) {
@@ -23,7 +36,9 @@ class Root extends Container {
         }
       } else if (this.first !== sample) {
         for (let node of nodes) {
-          node.raws.before = sample.raws.before
+          if (!keepBefore.has(node.raws)) {
+            node.raws.before = sample.raws.before
+          }
         }
       }
     }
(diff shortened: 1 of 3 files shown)


ai:
Thanks. Released in 8.5.19.

PR #2113, opened 2026-07-19
Fix missing space when AtRule#params is set after parsing

sarathfrancis90:
Setting `AtRule#params` on an at-rule that was parsed without params glues the params onto the name:

```js
let root = postcss.parse('@layer{a{color:black}}')
root.first.params = 'utilities'
root.toString() //=> '@layerutilities{a{color:black}}'
```

`@layerutilities` is a single at-word, so the output no longer parses back to the at-rule that produced it. Same for `@media{}`, `@scope{}`, `@font-face{}` — anything written without params.

`atruleStart()` already has a fallback that inserts the space when `raws.afterName` is missing, but the parser sets `raws.afterName = ''` explicitly for a params-less at-rule, so the fallback never gets a chance to run.

I still prefer the raw when it's there, and only fall back to a space when the params would otherwise merge into the at-rule name token. `@media(min-width:0)` and `@import"a.css"` stay byte-identical: `(` and `"` end the at-word on their own, and those delimiters are the only way the parser can produce an empty `afterName` together with non-empty params.

Found this fuzzing mutation round-trips over the postcss-parser-tests fixtures — across ~10k parse → set params → re-parse cases this was the only invariant that broke (100 distinct failures before, 0 after). Added tests for both directions; `pnpm test` is green.

### DIFF
--- a/lib/stringifier.js
+++ b/lib/stringifier.js
@@ -6,6 +6,10 @@
 const STYLE_TAG = /(<)(\/?style\b)/gi
 const COMMENT_OPEN = /(<)(!--)/g
 
+// Characters that end an at-rule name, mirroring RE_AT_END in the tokenizer.
+// Params starting with anything else need a space to stay separate tokens.
+const AT_NAME_END = /[\t\n\f\r "#'()/;[\\\]{}]/
+
 function escapeHTMLInCSS(str) {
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Released in 8.5.20.

PR #2114, opened 2026-07-21
Fix mismatched JSDoc comments on Position

isker:
Also, clarify that `column` is 1-based, like `line`.

I noticed this while looking at the [API docs](https://postcss.org/api/#position).


ai:
Thanks!

PR #2115, opened 2026-07-21
Fix childless at-rule losing its semicolon before a comment

sarathfrancis90:
While building a tree I put a childless at-rule (for example `@import`) just before a comment node using `append()`/`insertBefore()`, and the comment vanished after a stringify + re-parse round-trip.

The stringifier treats the last non-comment child as the one whose semicolon is optional, ignoring any trailing comments. That is fine for a declaration, but a childless at-rule reads everything up to the next `;`/`{` as its prelude, so without the semicolon the following comments get folded into the at-rule and are dropped on re-parse.

```js
let root = postcss.parse('a {}\n/* keep me */')
root.insertBefore(root.last, new AtRule({ name: 'import', params: '"x.css"' }))
root.toString()
// before: 'a {}\n@import "x.css"\n/* keep me */'  -> comment lost on re-parse
// after:  'a {}\n@import "x.css";\n/* keep me */'
```

The fix emits the semicolon when a childless at-rule still has following siblings, so the output round-trips. A genuinely last at-rule with nothing after it is unchanged.

Tested with two new unit tests (root and nested cases) and the full `pnpm test` suite.

### DIFF
--- a/lib/stringifier.js
+++ b/lib/stringifier.js
@@ -60,10 +60,23 @@ function pushBody(str, stack, node) {
   let semicolon = str.raw(node, 'semicolon')
   let isDocument = node.type === 'document'
   for (let i = nodes.length - 1; i >= 0; i--) {
+    let child = nodes[i]
+    let childSemicolon = last !== i || semicolon
+    // A childless at-rule that still has following siblings must be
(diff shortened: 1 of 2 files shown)


ai:
Thanks. Released in 8.5.21.

PR #2118, opened 2026-07-22
docs: show ESM and TypeScript plugin declaration

### LINKED ISSUE #1771: Document how to declare a PostCSS plugin in an ES6/TypeScript module

Hashim1999164:
## Summary

- Adds ESM and TypeScript examples for declaring a PostCSS plugin with `plugin.postcss = true` and `export default plugin`
- Points the plugin guidelines at the same pattern
- Fixes #1771

## Test plan

- [ ] Confirm the ESM example matches the `PluginCreator` / `postcss = true` API in `lib/processor.js`
- [ ] Skim docs site rendering for the new snippets in Writing a Plugin and Plugin Guidelines

### DIFF
--- a/docs/guidelines/plugin.md
+++ b/docs/guidelines/plugin.md
@@ -119,6 +119,10 @@ module.exports = opts => {
 module.exports.postcss = true
 ```
 
+For ESM / TypeScript modules, use the same `postcss = true` flag on the
+creator function and `export default plugin` (see
+[Writing a PostCSS Plugin](../writing-a-plugin.md)).
+
 ## 2. Processing
 
 ### 2.1. Plugin must be tested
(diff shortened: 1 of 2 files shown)


ai:
Thanks!

Can I ask you to replace CommonJS in examples to ESM. Let’s not keep ESM as an option since it is a new default in JS.

Hashim1999164:
Updated the examples so ESM is the default rather than an additional option: the JavaScript examples now default-export the plugin creator, while the TypeScript example remains for typing guidance.

PR #2121, opened 2026-07-28
fix: return empty array for empty string in list.split

amir-rezaei:
## Description
This PR updates `list.split` in `lib/list.js`.

## Details
- Returns an empty array `[]` early when `string` is empty or falsy to prevent spurious empty element insertion when parsing empty CSS value strings.


ai:
Thanks!

PR #2123, opened 2026-08-02
Fix tokenizer handling of escaped identifiers

### LINKED ISSUE #1349: Handling escape symbols
From https://acss.io/#pseudo-classes:

Each line should parse as one word (i.e. one identifier):

```css
C\(\#0280ae\)
C\(brandColor\)
C\(\#fff\)\:h:hover
```

This can be verified with https://rawgit.com/tabatkins/parse-css/master/example.html

This issue continues from https://github.com/shellscape/postcss-values-parser/issues/93

With thanks and credit to @nex3 and @ai for identifying this issue.

deepakganesh78:
Fixes #1349

## Reproduction

Before this change, tokenizing the issue examples split one escaped identifier into many `word` tokens:

```js
const tokenizer = require('./lib/tokenize')
const { Input } = require('./lib/postcss')
let input = String.raw`C\(\#0280ae\)`
```

Current `main` produced `C`, `\(`, `\#`, `0280ae`, `\)` as separate words instead of one identifier token.

## Root cause

The word scanner stopped at every backslash because `RE_WORD_END` treats `\` as a word boundary. The dedicated backslash branch consumed only the escape sequence itself, so escaped punctuation inside identifiers could never stay attached to the surrounding identifier text.

## Fix

Reuse the existing escape-consumption rules while scanning word endings. When the scanner reaches a valid CSS escape, it skips over that escape and continues the same word token; invalid escapes keep the previous token boundaries. This also keeps consecutive hexadecimal escapes in the same identifier token.

## Compatibility

No new dependencies. Parsing/stringification behavior is unchanged for normal CSS; the tokenizer now matches CSS identifier semantics for valid escaped punctuation.

## Validation

- `node -r ts-node/register/transpile-only test/tokenize.test.js`: 34/34 passed.
- Regression test with `lib/tokenize.js` reverted: failed as expected (`keeps escaped punctuation in identifiers`).
- `pnpm run test:coverage`: 682/682 unit tests passed.
- `pnpm run test:lint`: passed with 0 errors and one existing warning in `test/visitor.test.ts`.
- `pnpm run test:types`: passed.
- `pnpm run test:version`: passed.
- `pnpm run test:integration`: 33 real-world CSS fixtures passed.
- `pnpm run test:size`: passed, 16.37 kB / 16.5 kB.

Note: on Windows, raw `pnpm test` fails before running because the package script uses POSIX `FORCE_COLOR=1`; I ran the equivalent `test:*` scripts individually with `$env:FORCE_COLOR=1`.

### DIFF
--- a/lib/tokenize.js
+++ b/lib/tokenize.js
@@ -50,6 +50,54 @@ module.exports = function tokenizer(input, options = {}) {
     return returned.length === 0 && pos >= length
   }
 
+  function escapedEnd(start) {
+    next = start
+    escape = true
+    while (css.charCodeAt(next + 1) === BACKSLASH) {
+      next += 1
+      escape = !escape
+    }
+    code = css.charCodeAt(next + 1)
+    if (
+      escape &&
+      code !== SLASH &&
+      code !== SPACE &&
+      code !== NEWLINE &&
+      code !== TAB &&
+      code !== CR &&
+      code !== FEED
+    ) {
+      next += 1
+      if (RE_HEX_ESCAPE.test(css.charAt(next))) {
+        while (RE_HEX_ESCAPE.test(css.charAt(next + 1))) {
+          next += 1
+        }
+        if (css.charCodeAt(next + 1) === SPACE) {
+          next += 1
+        }
+      }
+    }
+    return next
+  }
+
+  function wordEnd(start) {
+    next = start
+    do {
+      RE_WORD_END.lastIndex = next + 1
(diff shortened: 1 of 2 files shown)


ai:
1. Such changes could be breaking changes. We can move it to PostCSS 9.0.
2. Do you know any use case of this new AST? How parsers will use it later?

De-ASI-INTERFACE (approved):

PR #2124, opened 2026-08-04
Upgrade nanoid to fix infinite loop on zero size

sinchang:
image

https://github.com/ai/nanoid/releases/tag/3.3.17


ai:
There is no need for this change. `^` is already containing `3.3.17`.

PR #2129, opened 2026-08-06
Fix Rule#selectors losing the empty selector

### LINKED ISSUE #2128: Issue with layer generation in 8.5.25
I'm facing an issue with latest postcss 8.5.25 (works correctly in 8.5.24) that one of my layers is missing `.dark` generated class name completely. I've created minimal reproduction, with a bit more info in the readme, so you can try for yourself: https://github.com/VojtechVidra/postcss-issue

I'm not using postcss directly, I'm using pandacss which itself is using postcss to generate final stylesheet. But the issue is only with the latest postcss version, which makes me believe it's an issue with postcss itself.

I'm happy to provide any more info if needed.

lazerg:
8.5.25 made `list.split()` return `[]` for an empty string (#2121), so `list.comma('')` went from `['']` to `[]`, and with it `Rule#selectors` on a rule whose selector is empty.

Plugins read that getter and assume at least one entry. `postcss-nested`'s `mergeSelectors()` loops over `parent.selectors` and joins the result back into the child, so an empty parent selector now wipes the child's:

```js
postcss([nested]).process('{ color: red; .dark & { color: blue } }', { from: undefined }).css
// 8.5.24: '{ color: red; } .dark  { color: blue }'
// 8.5.25: '{ color: red; }  { color: blue }'
```

That is what the Panda CSS reproduction in #2128 hits: the selector is wiped, the block is merged away, and `.dark` is gone from the tokens layer.

`''` is a string, and its `last: true` split has always been `['']`, same as the trailing item in `list.comma('a, b,')`. So this narrows the guard to non-strings, which keeps what #2121 was after (`list.split(undefined)` no longer throws) without changing the empty-string result.

Fixes #2128

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -11,7 +11,7 @@ let list = {
   },
 
   split(string, separators, last) {
-    if (!string) return []
+    if (typeof string !== 'string') return []
     let array = []
     let current = ''
     let split = false
--- a/test/list.test.ts
+++ b/test/list.test.ts
@@ -36,6 +36,15 @@ test('comma() adds last empty', () => {
   equal(list.comma('a, b,'), ['a', 'b', ''])
 })
 
+test('comma() keeps empty value', () => {
+  equal(list.comma(''), [''])
+})
+
+test('comma() ignores non-string values', () => {
+  // @ts-expect-error Testing invalid API
+  equal(list.comma(undefined), [])
+})
+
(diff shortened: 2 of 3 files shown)


ai:
cc @amir-rezaei

ai:
Thanks. Released in 8.5.26.

PR #2130, opened 2026-08-07
Upgrade nanoid to 5.1.6

max-culley-AJBELL:
Bump nanoid dependency from ^3.3.17 to ^5.1.6 in package.json, pnpm-lock.yaml, and pnpm-workspace.yaml.

nanoid v5+ ships as ESM-only (no CommonJS build), which would break 
equire('nanoid/non-secure') in lib/input.js on Node versions without require(esm) interop (< 20.19 / < 22.12). To avoid a breaking change across postcss's supported engines range, wrap the require in a try/catch that falls back to nanoid's own non-secure ID generation algorithm inlined locally when the ESM require fails.

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -1,6 +1,21 @@
 'use strict'
 
-let { nanoid } = require('nanoid/non-secure')
+let nanoid
+try {
+  ;({ nanoid } = require('nanoid/non-secure'))
+} catch {
+  // nanoid 5+ ships ESM-only, so `require()` throws on Node versions
+  // without ESM interop (< 20.19 / < 22.12). Fall back to the same
+  // non-secure algorithm nanoid uses internally.
+  let urlAlphabet =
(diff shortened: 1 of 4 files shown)


ai:
It will break node.js <22 users whom we still support

max-culley-AJBELL:
> It will break node.js <22 users whom we still support

https://security.snyk.io/package/npm/nanoid/3.3.16

Vuln fix needed

ai:
No, all security fixes were backported to 3.x branch and 3.3.16 is secure. snyk just has outdated data.

max-culley-AJBELL:
> No, all security fixes were backported to 3.x branch and 3.3.16 is secure. snyk just has outdated data.

What would you recommend using which shows this version is not vulnerable? Also this project uses 3.3.17, not 3.3.16

ai:
Use `npm audit` or `pnpm audit`

PR #2136, opened 2026-08-14
Do not emit multiple UTF-8 BOM markers

### LINKED ISSUE #2133: Multiple BOM symbols inside generated css in v8.5.24
Hi, I've got following issue after update postcss to latest version.
It's not reproduced on v8.5.23 and lower.

I have sass files in react app, all of them are encoded as UTF-8 without bom (correct encoding is guaranteed by stylelint stylistics lint rule)

Image

old versions:

Image

v8.5.24:

Image

arronKler:
## Why

v8.5.24 started restoring a leading BOM on stringify (#2119). That is correct for a single file, but `parse`/`stringify` could still produce more than one U+FEFF:

- input with several leading BOMs kept extras after the first
- concatenated CSS (`a{}\uFEFFb{}`) re-emitted the mid-file marker as part of the next selector
- a `Document` of several BOM-tagged roots wrote a BOM before every root

Those extra markers show up in the middle of generated CSS and break the first rule after each one.

## Solution

- Strip every leading U+FEFF / U+FFFE in `Input`, not only the first
- Skip stray BOM markers in the tokenizer so they are not part of the AST
- Emit at most one BOM from `Stringifier#root`

A single leading BOM is still preserved. `parse('\uFEFFa{}').toString()` remains `\uFEFFa{}`.

Fixes #2133

### DIFF
--- a/lib/tokenize.js
+++ b/lib/tokenize.js
@@ -46,12 +46,23 @@ module.exports = function tokenizer(input, options = {}) {
     throw input.error('Unclosed ' + what, pos)
   }
 
+  // Extra U+FEFF markers appear when BOM-preserving outputs are concatenated.
+  function skipBom() {
+    while (pos < length) {
+      code = css.charCodeAt(pos)
+      if (code !== 0xfeff && code !== 0xfffe) break
+      pos += 1
+    }
+  }
+
   function endOfFile() {
+    if (returned.length === 0) skipBom()
     return returned.length === 0 && pos >= length
   }
 
   function nextToken(opts) {
     if (returned.length) return returned.pop()
+    skipBom()
     if (pos >= length) return
 
     let ignoreUnclosed = opts ? opts.ignoreUnclosed : false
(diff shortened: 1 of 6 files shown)


lib/tokenize.js
```diff
@@ -46,12 +46,23 @@ module.exports = function tokenizer(input, options = {}) {
...
+  }
+
   function endOfFile() {
+    if (returned.length === 0) skipBom()
     return returned.length === 0 && pos >= length
   }
 
   function nextToken(opts) {
     if (returned.length) return returned.pop()
+    skipBom()
```
ai, line 65: Why do we need it if we already have `this.css.replace(/^[\uFEFF\uFFFE]+/, '')`?

De-ASI-INTERFACE (approved):

ai:
@romainmenke found an issue with current approach

> This removal isn't correct as U+FEFF is only a BOM if found at the start and has another (deprecated) meaning if found elsewhere, as a zero width no-break space.

But we still can think of removing `U+FEFF` in the context where `U+FEFF` is illegal anyway

PR #2109, opened 2026-07-05
Keep only major.minor in Processor#version

MahinAnowar:
Fixes #2101

`Processor#version` now holds only `major.minor` (`'8.5'`), so patch releases no longer require the manual `lib/processor.js` sync (the repo is currently on the 8.5.16 tag while the file said `8.5.15` — exactly the chore this removes).

Details:

- The only functional consumer, the plugin-compatibility warning in `lib/lazy-result.js`, already compares just `split('.')[0]`/`[1]`, so it works unchanged with a two-segment version.
- `test/version.js` still guards against drift: it now compares `Processor#version` against `package.json`'s `major.minor`, so a minor/major bump without updating `lib/processor.js` still fails CI, while patch releases no longer trip it.
- Updated the two format assertions (`test/processor.test.ts`, `test/postcss.test.ts`) to `/^\d+\.\d+$/` — they fail on the previous three-segment value and pass now — and the `Processor#version` JSDoc.

All checks pass locally: 652/652 unit tests, lint, types, version guard.

## Summary by CodeRabbit

* **Changes**
  * Updated the reported processor version to use `major.minor` format.
  * Updated version validation and related checks to match the shortened format.
  * Clarified version documentation to describe the new format.

PR #2137, opened 2026-08-14
Put only major.minor in Processor.version

### LINKED ISSUE #2101: Put opnly `major.minor` in `Processor.version`
It will prevent doing a lot job of keeping it synced with `package.json`

theRizwan:
Fixes #2101.

`Processor.version` was a full `major.minor.patch` string hardcoded in `lib/processor.js`, which meant it had to be edited on every patch release to stay in step with `package.json`. This drops the patch component so it only needs touching on a minor bump.

## Why this is safe

The patch component was never read. The only runtime consumer is the plugin-version mismatch warning in `lib/lazy-result.js`:

```js
let a = pluginVer.split('.')
let b = runtimeVer.split('.')

if (a[0] !== b[0] || parseInt(a[1]) > parseInt(b[1])) {
```

It compares `[0]` and `[1]` only, so `'8.5'` behaves identically to `'8.5.26'` here. `postcssVersion`, which `lib/postcss.js` sets from `new Processor().version`, flows into the same comparison and is unaffected. The documented example in `processor.d.ts` uses `version.split('.')[0]`, which also still works.

## Tests

`test/version.js` previously asserted exact equality with `package.json`, which would now always fail. It compares the major and minor components instead, so the guard against the two drifting is kept rather than removed:

```js
let expected = pkg.version.split('.').slice(0, 2).join('.')
```

Two assertions matched `/\d+.\d+.\d+/`. Both are now `/^\d+\.\d+$/` — anchored, and with the dots escaped, which the originals did not do.

I have also noted the format in the `processor.d.ts` doc comment, since it is a visible change for anyone reading `version` directly.

## Checks

- `pnpm unit` — 696 passing, 0 failing
- `node ./test/version.js` — passes
- `pnpm test:types` — clean
- `pnpm test:lint` — one pre-existing warning in `test/visitor.test.ts` (`perfectionist/sort-objects`), present on an unmodified checkout and untouched here

### DIFF
--- a/lib/processor.d.ts
+++ b/lib/processor.d.ts
@@ -37,7 +37,7 @@ declare class Processor_ {
   plugins: (Plugin | TransformCallback | Transformer)[]
 
   /**
-   * Current PostCSS version.
+   * Current PostCSS version as `major.minor`, without the patch component.
    *
    * ```js
    * if (result.processor.version.split('.')[0] !== '6') {
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -7,7 +7,7 @@ let Root = require('./root')
 
 class Processor {
   constructor(plugins = []) {
-    this.version = '8.5.26'
+    this.version = '8.5'
     this.plugins = this.normalize(plugins)
   }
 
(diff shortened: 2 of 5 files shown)


ai:
Duplicate https://github.com/postcss/postcss/pull/2109

And honestly, it looks like LLM slop (description is irrelevant to the changes

theRizwan:
You're right I missed #2109, which does the same thing. Apologies for the noise.

PR #2138, opened 2026-08-15
Fix `'source' does not exist in type 'DeclarationProps'` error

romainmenke:
Screenshot 2026-08-15 at 17 37 48

I think it was simply an oversight that this type does not extend `NodeProps`

---

This isn't blocking for me, I've added `@ts-expect-error` comments on my end :)


ai:
Thanks. Good catch.

PR #2139, opened 2026-08-18
Replace picocolors with node:util styleText

### LINKED ISSUE #2116: Remove colors lib
Move to Node.js native

mrpmohiburrahman:
Fixes #2116

`picocolors` was only there to color terminal output, and `node:util` does the
same job, so this drops it from `dependencies`. New `lib/colors.js` exports a
`createColor(...formats)` factory plus an `isColorSupported` flag, modelled on
the `colors.js` in size-limit; the three call sites keep their `text => string`
shape, so `HIGHLIGHT_THEME` and `showSourceCode()` are unchanged in structure.

Rather than eyeball the output, I diffed it against picocolors byte for byte —
old and new `showSourceCode()` and `showSourceCode(true)` are identical across
9 environments (piped, `FORCE_COLOR`, `NO_COLOR`, `CI`, `TERM=dumb`, `--color`,
and the empty-string variants of each). The empty-string cases are why the
detector uses truthiness rather than presence: `NO_COLOR=""` must not disable
color, per no-color.org and picocolors.

Two details worth flagging, both of which changed the code:

- `styleText` accepts an **array** of formats only since **20.13**, not 20.12.
  On 20.12.x, `createColor('bold', 'red')` throws `ERR_INVALID_ARG_VALUE` —
  while formatting a syntax error, which is the worst possible moment. So the
  guard is a feature detect, not a version check.
- `./lib/terminal-highlight` is a published entry in `exports`, and its old
  `pico.yellow` values were no-ops when color was unsupported. The theme is
  gated on `isColorSupported` so direct importers keep that contract.

`size-limit` drops 16.38 kB → 16.18 kB, so I lowered the budget to 16.25 kB.
New `test/colors.test.ts` covers the detection signals and the degrade path;
I checked it is not vacuous by breaking the code four ways and confirming a
specific test goes red each time.

### One thing I need you to decide

This is the conservative version: on Node < 20.13 the module degrades to plain
text, the color assertions degrade with it, every CI leg stays green, and
`engines` is untouched. The cost is a real regression — **error output loses
its color on Node 14, 16, 18 and 20.0–20.12**, where picocolors colored fine.
It degrades silently rather than throwing.

If 9.0 is bumping the Node floor anyway, the other version is smaller and
cleaner: `engines` → `>=20.13`, prune the `old` matrix to 20, and delete the
fallback branch entirely. I did not want to make that call on your behalf.
Say which you prefer and I will rewrite this.

### DIFF
--- a/lib/colors.js
+++ b/lib/colors.js
@@ -1,0 +1,38 @@
+'use strict'
+
+let { styleText } = require('util')
+
+// util.styleText() was added in Node.js 20.12, but it accepts an array
+// of formats only since 20.13. Detect what we actually use.
+let hasStyleText = typeof styleText === 'function'
+if (hasStyleText) {
+  try {
+    styleText(['bold', 'red'], '', { validateStream: false })
+    /* c8 ignore next 3 */
+  } catch {
+    hasStyleText = false
+  }
+}
+
+// node:util has no isColorSupported, so repeat picocolors' detection here
+let isColorSupported =
+  hasStyleText &&
+  !(process.env.NO_COLOR || process.argv.includes('--no-color')) &&
+  Boolean(
+    process.env.FORCE_COLOR ||
+      process.argv.includes('--color') ||
+      process.platform === 'win32' ||
+      ((process.stdout || {}).isTTY && process.env.TERM !== 'dumb') ||
+      process.env.CI
+  )
+
+// Node.js re-opens the parent style after the nested one only since 22.19
+// and 24.5. To support older versions, styles should be combined
+// in a single call like createColor('bold', 'red') instead of bold(red()).
+function createColor(...formats) {
+  /* c8 ignore next */
+  if (!hasStyleText) return text => String(text)
+  return text => styleText(formats, String(text), { validateStream: false })
+}
+
+module.exports = { createColor, isColorSupported }
--- a/lib/colors.js
+++ b/lib/colors.js
@@ -2,21 +2,8 @@
 
 let { styleText } = require('util')
 
-// util.styleText() was added in Node.js 20.12, but it accepts an array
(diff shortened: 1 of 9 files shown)


lib/colors.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
+
+let { styleText } = require('util')
+
+// util.styleText() was added in Node.js 20.12, but it accepts an array
+// of formats only since 20.13. Detect what we actually use.
+let hasStyleText = typeof styleText === 'function'
```
ai: No, we do not need it. We will work with >=22 in PostCSS 9.

mrpmohiburrahman: Done — the detect and the plain-text fallback are gone, and `engines` is now `>=22`.

Heads-up on CI: the `old` job now runs entirely below that floor. Node 20 still passes (setup-node resolves it to 20.19, which has `styleText` and the array form), but 10, 12, 14, 16 and 18 have no `styleText` at all, so those five legs fail by design. I left `.github/workflows/` alone since that job is yours to decide on — happy to drop it here, along with `test/old-node.js` and the `old` script, if you want that in this PR.

One detail worth knowing: `eslint-plugin-n` still flags `util.styleText` because it treats it as experimental until 23.5, so I kept the `node-builtins` ignore entry. `>=22.13` doesn't clear it either — I checked.

lib/colors.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+let { styleText } = require('util')
+
+// node:util has no isColorSupported, so repeat picocolors' detection here
```
ai: Why do you think that we should copy `isColorSupported` from `picocolors`?

Should it be equal to `styleText`'s one?

Please, do not change code and give me your thought when you write it.

PR #2142, opened 2026-08-19
Add typed plugin creator helper

### LINKED ISSUE #2102: Add a way to create plugins in TypeScript
Right now plugin need to be a function with `fn.postcss = 1` marker.

But in TS there is no way to create function and sen key in the same manner. So create plugin is no easy.

evgenyponomarev:
## Summary

- add a one-argument `postcss.plugin(creator)` helper that marks and returns the same plugin creator function
- preserve the deprecated two-argument `postcss.plugin(name, initializer)` behavior and add declarations for both overloads
- preserve the creator’s concrete plugin/processor return type in TypeScript
- document the helper and cover its runtime, named-export, and declaration behavior

Fixes #2102

## Validation

- `env -u NO_COLOR pnpm test` — 700/700 unit tests, integration, lint, declarations, version, coverage, and size checks pass
- ESM named-export smoke test using `lib/postcss.mjs`
- bundled size remains below the 16.5 kB limit at 16.44 kB

## AI disclosure

This contribution was made with OpenAI Codex assistance. Codex inspected PostCSS’s current plugin normalization, legacy creator API, TypeScript declarations, ESM exports, documentation, and 9.0 milestone; implemented the backward-compatible overload; and ran the validation above. The API reasoning was to reuse the already-exported `plugin` entry point with a one-argument modern form rather than introduce another public helper name.

### DIFF
--- a/docs/writing-a-plugin.md
+++ b/docs/writing-a-plugin.md
@@ -103,24 +103,23 @@ plugin.postcss = true
 export default plugin
 ```
 
-In TypeScript, type the creator as `PluginCreator` so `postcss = true` is valid:
+In TypeScript, use the `plugin` helper to mark and type the creator:
 
 ```ts
-import type { PluginCreator } from 'postcss'
+import { plugin } from 'postcss'
 
 export interface PluginNameOptions {
   // Types for plugin options
 }
 
-const plugin: PluginCreator<PluginNameOptions> = (opts = {}) => {
+const pluginName = plugin<PluginNameOptions>((opts = {}) => {
   return {
     postcssPlugin: 'PLUGIN NAME'
     // Plugin listeners
   }
-}
-plugin.postcss = true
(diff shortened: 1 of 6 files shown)


ai:
But we are asking plugins to not have import of `postcss` because we can't guarantee that all plugins will use the same version of PostCSS and it could lead to broken AST with nodes from different PostCSS versions. 

It is exactly why we deprecated this method

evgenyponomarev:
You’re right — the runtime helper would require plugin packages to load their own `postcss` copy, which violates the project’s no-runtime-import guidance and can mix AST nodes across versions. The type-only `PluginCreator` pattern added in #2118 is erased at compile time and already provides the safe TypeScript path, so this PR is redundant rather than something that should be revised into another runtime helper. Closing it. Thanks for the clarification.

PR #2143, opened 2026-08-19
Preserve comments in values and at-rule params

### LINKED ISSUE #1706: Remove comment cleaning from declaration value and at-rule parameters in PostCSS 9.0

evgenyponomarev:
## Summary

- keep comment tokens directly in `Declaration#value` and `AtRule#params` so changing either property no longer discards embedded comments
- retain the existing selector comment-cleaning behavior
- adapt the shared PostCSS 8 parser fixture expectations to the planned 9.0 semantics
- update the AST documentation and add mutation regressions for declarations, at-rules, and selectors

Fixes #1706

## Validation

- `env -u NO_COLOR pnpm test`
- 702/702 unit tests pass with 100% line coverage
- integration, TypeScript declarations, lint, version, and size checks pass
- bundled size remains below the 16.5 kB limit at 16.37 kB

## AI disclosure

This contribution was made with OpenAI Codex assistance. Codex reproduced comment loss on current `main`, traced it to `Parser#raw`, limited the behavior change to declaration values and at-rule params, added PostCSS 9 fixture adaptation and regression coverage, and ran the complete validation above. The selector path was explicitly kept unchanged after reviewing the shared parser boundary.

### DIFF
--- a/docs/writing-a-plugin.md
+++ b/docs/writing-a-plugin.md
@@ -141,8 +141,8 @@ PostCSS parses CSS to the tree of nodes (we call it AST). This tree may content:
   or `@media (screen) {}`.
 - [`Rule`]: selector with declaration inside. For instance `input, button {}`.
 - [`Declaration`]: key-value pair like `color: black`;
-- [`Comment`]: stand-alone comment. Comments inside selectors, at-rule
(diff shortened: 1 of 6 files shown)


ai:
We will be able to do it only in upcoming PostCSS 9.0 release

PR #2144, opened 2026-08-26
Fix chinese print not work

Jesse205:
The language code for Chinese is `zh`, while `cn` is an ISO country code (referred to as territory in POSIX). According to the [POSIX specification](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html#tag_08_02), the format of the LANG environment variable is `language[_territory][.codeset][@modifier]`. Since LANG starts with language, we should check whether it is `zh`.
image


ai:
Thanks!

PR #2150, opened 2026-09-03
Keep non-annotation comments when the processor has no plugins

dylanpulver:
With no plugins, `process()` deletes every comment beginning with `/*#` — including the `/*#region */` / `/*#endregion */` folding markers editors use.

```js
let css = "/*#region layout */\na{color:red}\n/*#endregion */\n"
postcss([]).process(css, { from: undefined }).css
// "\na{color:red}\n"
postcss([() => {}]).process(css, { from: undefined }).css
// unchanged
```

This happens with `map` unset too: `NoWorkResult` always calls `clearAnnotation()`.

`clearAnnotation()` has two implementations. The AST one removes only comments whose text starts with `# sourceMappingURL=`, and `PreviousMap#loadAnnotation` finds annotations by that marker. The string one — the only path `NoWorkResult` takes — searched for `/*#` alone.

Searching for the whole `/*# sourceMappingURL=` marker keeps the plain string scan, so 497daef7's move off RegExp stands. The old RegExp had the same gap, anchored to end of line.

Two tests added. Reverting `lib/map-generator.js` fails both. A narrower fix keeping the `/*#` needle and stopping at the first non-annotation still fails the second, where the stray comment follows the annotation.

`pnpm test` green: 701/701 unit (699 before), types, size, integration, lint (one pre-existing warning in `visitor.test.ts`).

Unchanged and disclosed: the string path still leaves `/* # sourceMappingURL=… */` (space after `/*`) and appends a second annotation, where the AST path clears it.

Drafted with Claude Opus 5 (`claude-opus-5`); every claim was executed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source-map annotation cleanup so regular comments are preserved.
  * Ensured comments following source-map annotations remain intact when no source path is provided.

* **Tests**
  * Added coverage for preserving non-annotation comments and correctly removing source-map annotations.

### DIFF
--- a/lib/map-generator.js
+++ b/lib/map-generator.js
@@ -75,9 +75,10 @@ class MapGenerator {
         }
       }
     } else if (this.css) {
+      let annotation = '/*# sourceMappingURL='
       let startIndex
-      while ((startIndex = this.css.lastIndexOf('/*#')) !== -1) {
-        let endIndex = this.css.indexOf('*/', startIndex + 3)
+      while ((startIndex = this.css.lastIndexOf(annotation)) !== -1) {
+        let endIndex = this.css.indexOf('*/', startIndex + annotation.length)
         if (endIndex === -1) break
         while (startIndex > 0 && this.css[startIndex - 1] === '\n') {
           startIndex--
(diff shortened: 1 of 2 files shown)


ai:
Thanks!

ai:
Released in 8.5.27

PR #2155, opened 2026-09-13
Return the processed CSS from NoWorkResult#toString()

giaBaoJS:
When `Processor#process()` gets no plugins it returns a `NoWorkResult`. Its
constructor runs the map generator and rewrites `this.result.css`: it appends a
generated `sourceMappingURL` annotation, or strips a stale one left in the
input. `toString()` skipped all of that and returned the raw input string, so
`String(result)` and `result.css` disagreed whenever the constructor had
changed anything.

`lib/no-work-result.d.ts` declares `NoWorkResult_ implements LazyResult<Root>`,
and `LazyResult#toString()` is documented as "Alias for the `LazyResult#css`
property. `lazy + '' === lazy.css`". `LazyResult#toString()` returns
`this.css`; `NoWorkResult` had `css` and `content` reading `this.result.css`
while `toString()` read the untouched `this._css`.

Reproduction:

```js

// 1. a generated inline map is dropped
let a = postcss([]).process('.foo { color: red }\n', {
  from: 'foo.css',
  map: true
})
a.css      // '.foo { color: red }\n\n/*# sourceMappingURL=data:application/json;base64,... */'
String(a)  // '.foo { color: red }\n'

// 2. a stale annotation is leaked
let b = postcss([]).process(
  '.foo { color: red }\n\n/*# sourceMappingURL=old.css.map */\n',
  { from: 'foo.css', map: false }
)
b.css      // '.foo { color: red }\n'
String(b)  // the input, annotation and all
```

With one plugin registered the same calls take the `LazyResult` path and both
values agree.

The fix makes `toString()` return `this.result.css`, matching `css`, `content`
and `LazyResult#toString()`. This is residue from #1909, which converged the
`NoWorkResult` constructor with `LazyResult` but left `toString()` reading the
pre-map string.

The three new tests sit next to the `no work result matches lazy result...`
tests that #1909 added; those compare `.css`, these compare `toString()`. They
cover the two branches of the constructor separately: a generated inline map, a
generated external map (which also sets `result.map`), and a stale inline map
being stripped. The existing `stringifies css` test already asserted
`` `${result}` === result.css ``, but only on input the constructor never
touches, so it passed throughout.

Reverting the one-line change turns all three red; a patch that fixes only the
map-generating branch, only the annotation-stripping branch, or that trims the
result is caught too. Full `pnpm test` is green, 704 tests (701 before), c8
line coverage still at 100%.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stringifying a no-work result now consistently returns the processed CSS, including source map transformations, instead of the original source.

* **Tests**
  * Added coverage for string conversion with enabled maps, non-inlined maps, and inline source maps.


ai:
Good idea. But I prefer to keep to 9.0 release soon since it could create some breaking changes.

giaBaoJS:
Understood, and 9.0 is the right place for it. Returning the processed CSS changes what a lazy result prints once map options are involved, so anyone relying on the current output would see a difference.

The branch still merges cleanly on `main` and CI is green, so it is ready whenever you pick up 9.0. I did not see a 9.0 branch to retarget at, so I will leave it open against `main` unless you would rather I moved it.

PR #2158, opened 2026-09-15
Terminate a custom property after a free semicolon before a comment

wahidrizka:
#2126 made the stringifier decide whether a declaration is a custom property from `raws.before`: it counts as one unless `before` ends with a non-space character, which excludes the `*`/`_` hack prefix. But the parser also puts free semicolons into `before` (`Parser#freeSemicolon`), so a custom property after a double semicolon is no longer terminated before a following comment, and the comment is lost when the output is parsed again. This worked in 8.5.26:

```js
let root = postcss.parse('a{color:red;;--x:blue}')
root.first.append(new postcss.Comment({ text: 'note' }))
root.toString()
// 8.5.28: 'a{color:red;;--x:blue/* note */}'   re-parses as decl, decl
// 8.5.26: 'a{color:red;;--x:blue;/* note */}'  re-parses as decl, decl, comment
```

`a{color:red;;--x:blue}` is valid CSS: CSS Syntax 3 discards a lone `;` inside a block ([consume a block's contents](https://drafts.csswg.org/css-syntax-3/#consume-block-contents)), and Chrome 152 reads the rule as `color: red; --x: blue`. So `--x` is a real custom property here, which is also what the parser decides from its first token.

A `--` property now counts as a custom property when `before` ends with a space or a semicolon. The hack-prefix test from #2126 still passes, and the new test fails without the change. The unit tests (702) and ESLint pass, and line coverage stays at 100%.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of custom properties when preceded by semicolons, whitespace, parser-generated prefixes, or stray tokens.
  - Preserved correct formatting and parsing when comments follow custom property declarations.

- **Tests**
  - Added coverage for custom properties following free semicolons and preceding comments.


test/stringifier.test.js
```diff
@@ -240,6 +240,32 @@ test('terminates indented custom property followed by a comment', () => {
...
+  is(
+    parse(css.toString())
+      .first.nodes.map(i => i.type)
+      .join(','),
+    'decl,decl,comment'
+  )
+})
+
+test('keeps property after a stray token before a comment unchanged', () => {
+  for (let css of ['a{: --x:red/*c*/}', 'a{] --x:red/*c*/}']) {
```
ai: I need some real proofs that in browser or any other good reference it will be `red` value and not `--x:red/*c*/`

wahidrizka: You are right to ask. A browser does not read that input as a declaration at all. CSS Syntax 3 only [consumes a declaration](https://drafts.csswg.org/css-syntax-3/#consume-declaration) when it starts with an ident; otherwise it is dropped as a bad declaration. In Chrome 152 (Edge 153 gives the same), the rules `.t1{: --x:red/*c*/}` and `.t2{] --x:red/*c*/}` end up with no declarations (`cssText` is `.t1 { }`), while `.t3{--x:red/*c*/}` gives `--x: red`.

So there is no reference for how that input should print. I removed this test and narrowed the change: only a semicolon at the end of `before` is now treated like a space, so input like this prints as it does on main.

What is left is the double-semicolon case, which is valid CSS: Chrome reads `.t4{color:red;;--x:blue}` as `color: red; --x: blue`, and 8.5.26 kept a comment appended after that `--x`. I updated the title and description to match.

ai:
Thanks

PR #2159, opened 2026-09-19
Split CSS lists on carriage returns and form feeds

jakezwang:
`list.space('1px\r2px')` and `list.space('1px\f2px')` currently return one value. Include carriage returns and form feeds in the separators, matching the whitespace already recognized by the tokenizer.

Add regressions for both characters and controls for function contents and escaped values.

Validation: both regressions fail before the fix. Full `pnpm test` passes on Node 26, including all 713 unit tests, coverage, types, lint, integration, version, and size checks. All 713 unit tests also pass on Node 22 and 24.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved whitespace handling so carriage returns and form feeds correctly separate values.
  - Preserved whitespace within functions such as `calc()` and within escaped values.

- **Tests**
  - Added coverage for additional whitespace separators and whitespace-only input behavior.


ai:
Why do you need it?

jakezwang:
The case is a declaration that PostCSS parses successfully, but `list.space()` cannot split:

```js
const value = postcss.parse('a { margin: 1px\r2px }').first.first.value
postcss.list.space(value) // currently ['1px\r2px'], expected ['1px', '2px']
```

A plugin using `list.space()` to inspect or expand that shorthand sees one component instead of two. Form feeds have the same result. Both characters are CSS whitespace, and the parser preserves them in `decl.value`.

This came from checking the parser and list helper for consistency; I don't have a downstream production bug report.

ai:
Do you really have a case with `\f` separator? Or it is just LLM's PR to get contribution?

jakezwang:
I don't have a production case for `\f`. This was found during an AI-assisted consistency check, and the reproduction is synthetic. I should have made that explicit when opening the PR.

ai:
Sorry, I am not interested in increasing code complexity by fixing theoretical issues which was never happen in 5+ years history of PostCSS.

You should add more personal input to LLM-assisted checks. I could run Fable with `find all issues` too.

